PR #1, opened 2020-08-26
Add Github Actions to build / test / lint in a linux environment and format all source-code files

xTVaser:
I need to follow up with some documentation, but to run clang-format enmasse:
```
python ./third-party/run-clang-format/run-clang-format.py -r common decompiler game goalc test -i
```

Also removed some prints in the test code, mainly the ellipsis in the while loops, this was causing tens of thousands of lines in the test output.  I think there are other instances of `prints` in the test code.  Perhaps gtest provides a logger of some kind to use so the output can be truncated, as I'm running it in brief mode (only shows a summary and failed tests)

This is my first time working with github actions so...not sure how this merge will go...i think it just has to be merged?  There are some issues with forked pull requests and github actions but I think that doesn't apply in this case.  You can see it in action here - https://github.com/xTVaser/jak-project/runs/1029924771?check_suite_focus=true

Once we get Windows support merged, I'll add AppVeyor support so we know if theres ever a regression


test/test_listener_deci2.cpp
```diff
@@ -58,16 +58,16 @@ TEST(Listener, DeciThenListener) {
     EXPECT_FALSE(s.check_for_listener());
     EXPECT_TRUE(l.connect_to_target());
     // kind of a hack.
-    while(!s.check_for_listener()) {
-      printf("...\n");
+    while (!s.check_for_listener()) {
+      // printf("...\n");
```
water111, line 62: This isn't an issue for me (it prints only zero time or once), but I imagine the networking stack on the github actions machines might be a lot busier than my desktop.  The test on github actions takes 1.5 seconds, compared to 1 second on my computer, so I don't think there's a serious delay here.  Just getting rid of the silly print seems like a good plan.

water111:
You were correct, after merging github actions magically started working!
https://github.com/water111/jak-project/runs/1034620122

PR #6, opened 2020-09-05
Add Windows CI

xTVaser:


water111:
Thanks, this is awesome, and it looks like it works perfectly!
https://github.com/water111/jak-project/actions/runs/240910447

PR #21, opened 2020-09-09
Add a function to get file paths

blahpy:
I think this should resolve https://github.com/water111/jak-project/issues/7 but please let me know if it's not doing what you expected.

CMake also updated to add new files.

Please let me know if I'm doing this wrong! :)


common/util/FileUtil.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+std::string FileUtil::GetCurrentWorkingDir() {
+  char buff[FILENAME_MAX];
+  GetCurrentDir(buff, FILENAME_MAX);
+  std::string current_working_dir(buff);
+  return current_working_dir;
+}
+
+std::string FileUtil::get_file_path(std::vector<std::string> input) {
+  std::string currentPath = FileUtil::GetCurrentWorkingDir();  // std::filesystem::current_path();
```
water111, line 21: Vaser pointed out that we could use `std::filesystem` if we upgraded to C++17, so I did.  Maybe we can use `std::filesystem` here instead, if that's easier.

common/util/FileUtil.cpp
```diff
@@ -0,0 +1,36 @@
...
+#ifdef _WIN32
+  #include <direct.h>
+  #define GetCurrentDir _getcwd
+#else
+  #include <unistd.h>
+  #define GetCurrentDir getcwd
+#endif
+
+std::string FileUtil::GetCurrentWorkingDir() {
+  char buff[FILENAME_MAX];
```
water111, line 14: I think this may need to be modified slightly - if you run the program from a different folder, `getcwd` will give you the name of the folder you ran it from, not the name of the folder the program is actually in.  I'm not sure yet what the best thing is to do here, but we can discuss it more in the group chat.

common/util/FileUtil.cpp
```diff
@@ -0,0 +1,36 @@
...
+  char dirSeparator;
+
+  #ifdef _WIN32
+    dirSeparator = '\\';
+  #else
+    dirSeparator = '/';
+  #endif
+
+  std::string filePath = currentPath;
+    for (int i = 0; i < input.size(); i++) {
```
water111, line 31: It looks like the formatting here is slightly off - we use a tool called `clang-format` to automatically format the code and there's a test that checks for this.  There's probably a way to run `clang-format` to format your code within visual studio.

common/util/FileUtil.cpp
```diff
@@ -0,0 +1,36 @@
...
+#endif
+
+std::string FileUtil::GetCurrentWorkingDir() {
+  char buff[FILENAME_MAX];
+  GetCurrentDir(buff, FILENAME_MAX);
+  std::string current_working_dir(buff);
+  return current_working_dir;
+}
+
+std::string FileUtil::get_file_path(std::vector<std::string> input) {
```
water111, line 20: This is a super minor detail, but if you replace `std::vector<std::string> input` with `const std::vector<std::string>& input` it will make this code slightly faster.  The combination of `const` and `&` tell the C++ compiler that this function isn't going to modify `input`, so instead of making a separate copy of `input` to be used in this function, it will just take a reference to `input`.  In this case it's not a big enough difference to matter (the vector of strings is tiny and fast to copy), but it's a good pattern to get used to using.

water111 (changes requested):
Thanks - this looks great!  I think setting up a `common/util` library like this will be very helpful.

There are a few things to change before we can merge this, but they are all minor:
- I think you will need to run `clang-format` to format this code
- The `GetCurrentWorkingDir` function needs to do something a little bit different and we should discuss this more
- You'll have to fix the merge conflicts in order for tests to run. Let me know if you need help doing this in the group chat - it shouldn't be very hard.

common/util/FileUtil.cpp
blahpy: Thanks, I've got `clang-format` working and have run it on all files. Very helpful because VS kept reformatting things strangely on its own!

common/util/FileUtil.cpp
blahpy: Thank you for this!

xTVaser:
I need to put this in the documentation, but if you want to run clang-format on the entire project you can do so in one line (assuming you have python installed):
`python ./third-party/run-clang-format/run-clang-format.py -r common decompiler game goalc test -i`

common/util/FileUtil.cpp
blahpy: As far as I can tell this would have had the same result as getcwd, so I went ahead and fixed it up the other way that we discussed in the Discord group.

common/util/FileUtil.cpp
blahpy: All right, I have (hopefully) updated it to find and use the filepath that you suggested.

water111 (approved):
Looks good, thanks!

PR #36, opened 2020-09-13
Ignore third-party folder in PRs, add syntax highlighting for GOAL files

xTVaser:


water111:
Thanks! At some point we'll probably need to look into formatting all the GOAL code too - I wonder if there's a clang-format for lisp.

PR #37, opened 2020-09-13
Add basic support for floats, add first game GOAL function

water111:
Bringing things up to date before adding spdlog.

Adds the first GOAL function: `identity` inside of `gcommon.gc`!


water111:
Failing test is coveralls - it seems to fail on PRs I make with only one commit.

PR #40, opened 2020-09-14
Resolve a good chunk of windows compiler warnings

xTVaser:
Mostly revolved around the new MSVC check for functions that don't cover all paths (ie. a switch statement without a default case).  It appears to not see an assert as a valid default case.

I switched assert(false) to exceptions in these cases.  I believe this should also abort the program, but will also provide a hopefully useful message?  Hopefully this is an improvement.

### DIFF
--- a/game/kernel/kprint.cpp
+++ b/game/kernel/kprint.cpp
@@ -884,6 +884,8 @@ s32 format_impl(uint64_t* args) {
                 call_method_of_type(in, type, GOAL_PRINT_METHOD);
               }
             } else {
+              // TODO - if we can't throw exceptions, what is the option?
+              // log, break and continue?
               throw std::runtime_error("failed to find symbol in format!");
             }
           }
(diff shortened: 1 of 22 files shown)


game/kernel/kprint.cpp
```diff
@@ -884,6 +884,8 @@ s32 format_impl(uint64_t* args) {
                 call_method_of_type(in, type, GOAL_PRINT_METHOD);
               }
             } else {
+              // TODO - if we can't throw exceptions, what is the option?
+              // log, break and continue?
               throw std::runtime_error("failed to find symbol in format!");
```
xTVaser, line 889: wasn't sure about how to fix these

water111: I think it's fine for now. I'm planning to add some better handling in the runtime for GOAL crashes that will include an `abort` which can safely be called in C code called from GOAL.

water111:
Thanks for doing this!  For the compiler, some exceptions will be caught by the compiler error system and show up like compiler errors.  They might be a little bit confusing, but it's better than the current "crash the compiler when there's a bug".

PR #45, opened 2020-09-15
Fix a Listener Bug

water111:
We've had a few random test failures related to the listener connection on github, and recently a test for `format` that sent a lot of data and failed always on github actions.  It seems like the listener can send data faster than the runtime can read it, causing an error that I wasn't checking for. After adding the error check, the `format` test seems to pass.

If all tests pass on this PR, I think this is the fix.

### DIFF
--- a/common/util/FileUtil.cpp
+++ b/common/util/FileUtil.cpp
@@ -16,10 +16,8 @@ std::string file_util::get_project_path() {
 #ifdef _WIN32
   char buffer[FILENAME_MAX];
   GetModuleFileNameA(NULL, buffer, FILENAME_MAX);
-  printf("using path %s\n", buffer);
   std::string::size_type pos =
       std::string(buffer).rfind("jak-project");  // Strip file path down to \jak-project\ directory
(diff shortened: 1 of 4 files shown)


water111:
Lots of failures!

- Linux failure is due to Coveralls, the tests passed. 
- Windows failed once because of listener reasons
- Windows failed a second time because format is totally broken there. I suspect I forgot to update some assembly stuff for windows.

water111:
BEEF appears:
```
[DECI2] Got message:
 3886 0 0xe042 H -> E
[DECI2] Got message:
 10572 47870 0xbeef � -> �
[DECI2] Got message:
 24 0 0xe042 H -> E
```
The `beef` is a placeholder in instructions, so I think the Listener is sending code twice (unlikely?) or the message receive is aborting too early and receiving part way through a message.

water111:
Looks like there are two separate bugs. This fixes a stupid mistake in the listener, the other bug is here: https://github.com/water111/jak-project/issues/46 and I'll fix it tomorrow.

PR #60, opened 2020-10-04
Add CFG to IR decompiler pass

water111:
This pass combines the control flow graph with the basic ops to build a single IR for an entire function. Most CFG forms are properly cleaned up, except for the short circuiting operators `and` and `or`.  Several special-case sequences for abs, ash, getting the runtime type of a boxed object, and comparisons which return a boolean explicitly are detected here.

There is still some work to do before this pass is finished:
- [x] Investigate several strange uses of likely branches. These are probably similar cases to abs/ash but occur very rarely.
- [x] Invert conditions as necessary during cleanup - currently they are the condition that causes the branch to be taken, but this is sometimes the opposite of the condition that would be written in source code.
- [x] Clean up the `and` and `or` forms.  This may be somewhat tricky if the cfg generating pass gets nesting wrong. If this does happen, it's not clear if this should be fixed in the cfg pass or in the pass, like I did with cond.
- [ ] Handle return-from-function
- [ ] Confirm the pass works on all functions with valid cfgs
- [x] Write some comments
- [ ] Consider merging some `IR` classes so there are fewer of them.


water111:
I think everything here works now!  We're still skipping functions that have `break` or early `return`, so only about 90% of functions make it through.  These will require work on the cfgvertex pass first.  There may also be some special cases we should detect during this pass, but I will deal with these as they come up later on.

PR #66, opened 2020-10-07
Make decompiler more successful

water111:
Before moving on to the higher level parts of the decompiler, I want to make the first passes 100% successful. Previously only around 90% of the functions made it through, and the other 10% failed for a variety of reasons.  Also only 96% of instructions are successfully converted to IR.  Ideally we would flag all functions that we expect to fail (due to inline asm), and all the other functions pass.

- [x] Handle `break` in the CfgVertex pass
- [ ] Handle the stupid case of two consecutive returns, with the second one unreachable, in the CfgVertex pass (happens 5x!)
- [x] Handle break/return-from-function in CFG IR Pass
- [x] Handle real but rare instructions in Basic Op pass
- [x] "asm pass through" for VU0 ops / MMI ops
- [x] 100% pass on cfg vtx
- [x] 100% pass on basic op
- [x] 100% pass on cfg ir
- [ ] review functions flagged as asm
  - [ ] config option to save all these to a file for easy inspection
- [ ] come up with a better way to name unnamed functions that will be robust to the order of decompilation


water111:
Current Status:
```
- Analyzing Functions...
Found 5939 functions (3050 with no control flow)
Named 4751/5939 functions (80.00%)
Excluding 224 asm functions
Found 43432 basic blocks in 1663.798 ms
 5715/5715 functions passed cfg analysis stage (100.00%)
 331551/331551 basic ops converted successfully (100.00%)
 5273/5715 cfgs converted to ir (92.27%)
```

water111:
```
- Analyzing Functions...
Found 5939 functions (3050 with no control flow)
Named 4751/5939 functions (80.00%)
Excluding 246 asm functions
Found 43432 basic blocks in 1682.907 ms
 5693/5693 functions passed cfg analysis stage (100.00%)
 328874/328874 basic ops converted successfully (100.00%)
 5693/5693 cfgs converted to ir (100.00%)
```

PR #67, opened 2020-10-08
Create more general Test Framework and Organize the bulk of current tests

xTVaser:
This PR doesn't yet add new torture tests, I figure that can come later.  Just trying to replicate the current tests with the new framework and work-out the kinks

Of which there are still a few.


test/goalc/test_arithmetic.cpp
```diff
@@ -0,0 +1,232 @@
...
+    if (hex) {
+      return std::string("#x") + std::string(std::to_string(val));
+    }
+    return std::to_string(val);
+  }
+
+	// This is used by the test runner code to know what the expected value is
+	// For a simple example like this, a single eval is all that's required, but for
+	// more complex tests, this may not be the case.
+  std::string eval() {
```
water111, line 60: This is great and will probably be super useful for more complicated randomly generated expressions.

test/goalc/test_arithmetic.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+// ValuesIn, is not the only way to use a parameterized test, but the most applicable for this example
+// You can actually get googletest to compute the permutations for you, which may be useful.  Consult their docs.
+// - https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#value-parameterized-tests
+INSTANTIATE_TEST_SUITE_P(EvalIntegers,
+                         ArithmeticTests,
+                         testing::ValuesIn(genIntegerTests(4,
+                                                           true,
+                                                           true,
+                                                           {IntegerParam(-2147483648),
```
water111, line 177: At some point we should add in more edge case integers.  See 'IR_LoadConstant64::do_codegen` which picks one of 5 different instructions for doing the load depending on the size/sign of the value.

test/goalc/test_arithmetic.cpp
```diff
@@ -0,0 +1,232 @@
...
+    IntegerParam test = additionalTests.at(i);
+    test.index = i + numTests - 1;
+    tests.push_back(test);
+  }
+
+  return tests;
+}
+
+// In the interest of speed, we want to share the same thread/compiler across
+// all the tests in this suite, so we have to over-ride this.
```
water111, line 114: :+1: without this the tests can take a while

test/goalc/test_arithmetic.cpp
```diff
@@ -0,0 +1,232 @@
...
+  void SetUp() {
+    GoalTest::createDirIfAbsent(GoalTest::getTemplateDir(testCategory));
+    GoalTest::createDirIfAbsent(GoalTest::getGeneratedDir(testCategory));
+  }
+
+  // You can define per-test tear-down logic as usual.
+  void TearDown() {}
+
+  // Common Resources Across all Tests in the Suite
+  static std::thread runtime_thread;
```
water111, line 141: This seems kinda weird, but maybe this is what you're supposed to do for gtests?

test/goalc/test_with_game.cpp
```diff
@@ -0,0 +1,116 @@
...
+  inja::Environment env{GoalTest::getTemplateDir(testCategory),
+                        GoalTest::getGeneratedDir(testCategory)};
+};
+
+std::thread WithGameTests::runtime_thread;
+Compiler WithGameTests::compiler;
+GoalTest::CompilerTestRunner WithGameTests::runner;
+
+// TODO - havn't done anything with these really yet
+TEST_F(WithGameTests, All) {
```
water111, line 60: This is probably for another PR, but I think a lot of these tests just call a single function with a few arguments and check the result. I imagine we could have a Call Function test that generates these without needing all these separate test `.gc` files.

water111 (commented):
This is awesome, I took a look and really liked this approach.  It will be a huge help for making a a reliable compiler!

test/goalc/test_arithmetic.cpp
xTVaser: Well, using gtest's Fixture support was the best way I could think of to share the thread/compiler/runner across all tests.

However I don't really like it for several reasons:
- Since the Fixture has to extend with the parameter, you can't easily have 2 different types of parameters to generate tests, unless you make a new fixture, which means a new thread,etc.
- Lots of boilerplate

So this is an area for improvement.  I think it would be nicer to have a "global" runner with/without the kernel, and have some sort of mechanism to "refresh/wipe" the state so we can ensure tests are hermetic.  I ran into some difficulty trying to do the former though.

There might also be a better way to do the parameterized tests - https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#creating-value-parameterized-abstract-tests so we can have a common library of some sorts.

water111: I looked at this a little bit more and I think what you have now is good. The `static` stuff in the fixture class seemed weird to me at first because I didn't realize each test would get its own fixture.

I agree we need a better way to reset the state of the compiler/runtime between tests. I don't think it would be too hard to add a `reset_state` to the compiler that clears out all the variable/constant/function/type definitions.  Doing a fast reset of the runtime is going to be much harder though.  There's some hacky stuff we could do like saving the entire runtime GOAL memory before a test and restoring it, but this won't catch everything.

test/goalc/framework/test_runner.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+  return result;
+}
+
+void CompilerTestRunner::run_static_test(inja::Environment& env,
+                                         std::string& testCategory,
+                                         const std::string& test_file,
+                                         const std::vector<std::string>& expected,
+                                         MatchParam<int> truncate) {
```
water111, line 40: Don't worry about changing this now, but C++17 now has `std::optional` which is better than the MatchParam thing.  Eventually I'll switch everything to std::optional

.github/workflows/workflow.yaml
```diff
@@ -5,7 +5,7 @@ jobs:
   common:
     # TODO - Start using clang-tidy - requires a bit more setup though - http://clang.llvm.org/docs/HowToSetupToolingForLLVM.html
     name: Linting & Formatting
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
water111, line 8: good idea

water111 (approved):
:+1:

PR #68, opened 2020-10-08
More logging improvements

doctashay:
spdlog now logs most output to `logs/game.log`. Also cleanup old prints and replace them with `spdlog` equivalents


doctashay:
spdlog is currently configured to build as header-only, which will slow down build times. this was necessary to work around a nasm multiple input error, and can be reverted once the nasm issue is addressed. 

i git blame myself for this PR being such a mess, need to start opening draft PRs. Should still be able to serve its' purpose

game/main.cpp
```diff
@@ -11,8 +11,12 @@
 int main(int argc, char** argv) {
   while (true) {
     spdlog::set_level(spdlog::level::debug);
+    auto game_logger = spdlog::basic_logger_mt("GOAL Runtime", "logs/runtime.log");
+    spdlog::set_default_logger(game_logger);
+    spdlog::flush_on(spdlog::level::info);
+
     // run the runtime in a loop so we can reset the game and have it restart cleanly
-    spdlog::info("gk {}.{} OK!\n", versions::GOAL_VERSION_MAJOR, versions::GOAL_VERSION_MINOR);
+    spdlog::info("gk {}.{} OK!", versions::GOAL_VERSION_MAJOR, versions::GOAL_VERSION_MINOR);
```
water111, line 19: Let's change this print message to say `OpenGOAL Runtime {}.{}`, now that the project has a name.

decompiler/util/LispPrint.cpp
```diff
@@ -0,0 +1,527 @@
+#include "LispPrint.h"
```
water111, line 1: I think this was added in by mistake?

CMakeLists.txt
```diff
@@ -55,11 +55,11 @@ include_directories(./)
 
 # build spdlog as a shared library to improve compile times
 # adding this as a SYSTEM include suppresses all the terrible warnings in spdlog
+IF (WIN32)
+include_directories(third-party/spdlog/include)
+ELSE()
 include_directories(SYSTEM third-party/spdlog/include)
-# this makes spdlog generate a shared library that we can link against
```
water111, line 59: I will take another look at this today, hopefully we can figure out what is going wrong here.

game/main.cpp
```diff
@@ -11,8 +11,12 @@
 int main(int argc, char** argv) {
   while (true) {
     spdlog::set_level(spdlog::level::debug);
+    auto game_logger = spdlog::basic_logger_mt("GOAL Runtime", "logs/runtime.log");
+    spdlog::set_default_logger(game_logger);
+    spdlog::flush_on(spdlog::level::info);
```
water111, line 16: This is great, we can change this if we ever want to see more or less logs for debugging.

water111 (commented):
Thanks, this is much cleaner! I think some other changes were accidentally included in this PR, like including all of the files for spdlog and removing the git submodule. 

I also left some comments on some prints that I think should be warnings - if these ever get printed something will probably crash or go wrong.

CMakeLists.txt
doctashay: https://github.com/water111/jak-project/pull/69 Is relevant

game/main.cpp
doctashay: Excellent! We should be able to have individual log files for the compiler, decompiler, and runtime, or at least individual logger objects.

PR #71, opened 2020-10-10
Clean up some warnings and add some extra integer tests

water111:
I think this should avoid all the warnings from inja.

I also added some more integer test cases because I'm paranoid about integer constants for some reason.


xTVaser (approved):
No errors on linux, looks good to me 
image

PR #77, opened 2020-10-12
rearrange for faster build time

water111:
On my computer, this decreased the single-thread build from 5 minutes to 3 minutes and the multi-threaded build from 27 to 16 seconds.


test/goalc/goalc_test_list.cpp
```diff
@@ -0,0 +1,18 @@
+// this list should contain all of the goalc test files.
```
water111, line 1: @xTVaser I combined all of your test files into a single giant file.  This seems to make a pretty big different because it ends up parsing the inja header only once.  Let me know if think this could have some negative consequence.

water111:
on github actions, the build time went from 2:49 to 1:38. But windows didn't work.... I'll look into this more later.

water111:
This turned out to be too hard to fix on windows, I'll revisit this later.

PR #84, opened 2020-10-15
Add auto-generated `inspect` methods to `deftypes`

xTVaser:
Partially addresses https://github.com/water111/jak-project/issues/72

image


goalc/compiler/compilation/Type.cpp
```diff
@@ -85,6 +85,75 @@ RegVal* Compiler::compile_get_method_of_object(RegVal* object,
   return deref->to_reg(env);
 }
 
+Val* Compiler::compile_format_string(const goos::Object& form,
+                                     Env* env,
+                                     std::string& fmt_template,
+                                     std::vector<RegVal*> args,
+                                     const std::string& out_stream) {
+  // Add first two format args
+  args.insert(args.begin(), compile_string(fmt_template, env, DEBUG_SEGMENT)->to_reg(env));
```
water111, line 94: Instead of putting `DEBUG_SEGMENT` here, you could do `get_parent_env_of_type<FunctionEnv>(env)->segment` which gets the segment that the enclosing function is in. In your case this will always be the `DEBUG_SEGMENT` because you pass in the `method_env`, but it's nice to make this more general.

The convention for most `compile` functions that generate static data is that they either explicitly take a segment or default to the enclosing function's segment.

goalc/compiler/compilation/Type.cpp
```diff
@@ -85,6 +85,75 @@ RegVal* Compiler::compile_get_method_of_object(RegVal* object,
...
+
+  std::string str_template = "";
+  std::vector<RegVal*> formatArgs = {};
+
+  // Check if there are no fields
+  if (structured_type->fields().empty()) {
+    str_template = "Type has no fields!~%";
+  } else {
+    for (Field f : structured_type->fields()) {
+      str_template += "~A~%";
```
water111, line 133: I'm imagining that this could work a little bit differently.  You could put the field's name directly in the `str_template`, then print out the actual values of the fields of the object that's passed to the inspect.

So for a field named `name`, the `str_template` would be something like `name: ~A`, and the you would add the value of `name` field to the `formatArgs`.  I'll leave more details in another comment.

water111 (commented):
Looks great so far! I realize that I didn't really give you enough information to know what to print out for each field.

Right now I don't know the exact rules for printing all the fields, so let's skip fields that are either dynamic or an array.  You can check these with `field.is_dynamic()` or `field.is_array()`. If either is true, just skip that field.

For a field named `my-field` which is a `basic`, `binteger`, or `pair` the format string should be `~Tmy-field: ~A`.  The three types `basic`, `binteger`, and `pair` can all be printed with the `~A` format code.

For a field which is an `integer`, do the same thing, but replace `~A` with `~D`.  This prints as a decimal.

For a field which is a `float`, do the same thing, but replace `~A` with `~f`. This prints a float.

For a field which is a `pointer`, do the same thing, but replace `~A` with `#x~X`.

For fields which don't satisfy any of these, for now don't print anything.

To check if a field is a certain type, do
```
m_ts.typecheck(m_ts.make_typespec("integer"), field.type(), "", false, false)
```

To get the value of the field to put in `formatArg`, you can do
```
get_field_of_structure(structure_type, input, field.name(), method_env).to_gpr(method_env);
```

So for your example, the output would be:
```
gc> (inspect test-object) 
    type: test-type
    param: 0
```
(by default all fields are zero initialized when the object is allocated with `new 'global`)

water111:
If the field is an array, you can do,
`~Tmy-field[<size>] @ #x~X~%`
where you replace `<size>` with the size of the array.

If the field is dynamic, do the same thing as array but make size `0`. (weird, but this is what GOAL does).

If the field is doesn't match any of the others, but is a `structure`, do
`~Tmy-field: #<type-of-field @ #x~X>~%`
where you replace `type-of-field` with the result of `field.type().print()`

PR #90, opened 2020-10-17
Partially Implement Decompiler Type Analysis

water111:
This is the beginning of the decompiler type analysis pass.  This is probably the most important/riskiest decompiler pass, for a few reasons:
- It will dramatically improve the readability of the code.
  - Once we give the decompiler type information, it will be able to actually decompile accessing fields of types.
- It is required for the next 2 big decompiler passes, which should make the code easier to read as well
- It must be done in order to know how many arguments each function call takes, which is very important to know.
  - There may be ways around this but they will be ugly - like always calling the function with 8 arguments and having "junk" in the unused arguments.  This will make the code very hard to read and not much better than just "translating" the MIPS to x86.

Currently the PR sets up the basic framework in the decompiler and adds features to type system to reverse field/array access.  The most common IR operations support type propagation, so the pass can be tested on simple functions.

Unfortunately I made a few mistakes when designing this pass so there are some issues that I need to resolve before continuing.  Neither are massive, but will take some time to fix.

**Manually Specify Types**
There are some rare cases where the compiler can't figure out types from looking at code.  I knew this would happen and figured that we would just manually specify types in these cases, but completely failed to take this into account when implementing things.  As a result, there is currently no way to manually specify types.

Probably the most common example is a GOAL pair, which holds two `object`s. In the case where the GOAL code calls a method of or accesses a field of one of these objects, the decompiler should know the type of the `object`.  There's two ways I can think of do this:

1. The pair being accessed is static (stored in the CGO file) and we can go look at the pair and see what the type is.
2. The pair being accessed is not static (or we can't prove that it's static automatically) so we will manually have to specify the type.

I think pairs are used pretty infrequently, so we could easily get away with 2.  But right now there's no easy way to do this.  It would be really nice to have a way to specify that a certain register holds a certain type at a certain instruction, but I don't have a way to easily refer to a single instruction. 

Maybe there could be a UI for viewing the disassembly? You could click on an MIPS instruction and edit some options, and it would save this into a JSON file.  I think object name + function name + instruction index will be a stable way to refer to individual instructions.

An easier solution is to just print out the index of each instruction and then you type it in when editing a config file by hand.

**Dynamic Array Access**
If there is GOAL code like
```
(-> my-object my-array 12)  ; like my_object->my_array[12] in C
```
then my current approach will work. It will analyze the GOAL code and figure out that it is reading from an array and get the appropriate type.

However, if the array index is a variable, it doesn't work.
```
(-> my-object my-array x) ; like my_object->my_array[x] in C
```
Here's an example of the assembly, with Basic OP IR next to it:
```
    dsll v1, a1, 2            ;; (set! v1 (shl a1 2)) compute offset into the array
    daddu v1, v1, a0          ;; (set! v1 (+.i v1 a0)) add array offset to object
    lwu v1, 16(v1)            ;; (set! v1 (l.wu (+.i v1 16))) load, adding the offset of the array into the object
```
Where `a0` is the `my-array`, `a1` is the index, `v1` is the result, and the size of each element in the array is 4.

This pointer to load from is the sum of three things:
- The object
- The offset of the array into the offset
- The offset into the array

Annoyingly GOAL does things in this order
- The object
- The offset into the array
- The offset of the array into the offset (constant)

and the type system completely fails when it sees (object + offset into array).

There are a few solutions here that could work:

1. Recognize this pattern of `dsll`, `daddu`, load and turn it into a single Basic Op for "load array" in the Basic Op pass.
2. Do this type analysis at a higher level, once we know that the load occurs at `(+ v1 (* a1 4) 16)`, and we can analyze the entire expression to figure out it's an array access. This would be after expression building (currently unimplemented), which I previously assumed would happen after type analysis. So this would require some redesigning.
3. Some way to manually specify these three instructions as an array access. 

I'm leaning toward 1 because it seems like it could "just work" without too much more work. I suspect there could be false positives, but that might be fine - an IR clean pass that runs after type analysis could expand it back to the old version.

Are there going to be other cases like this, where we need higher-level information to determine types?  For example, getting the address of an element in an array.  If we keep hitting these cases, we may need to do the type pass a little bit later, which is going to make it much more complicated.

Example output of this pass (shown in the comments of the disassembly for now)
```
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
; .function basic-type?
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
  ;stack: total 0x00, fp? 0 ra? 0 ep? 1
L285:
    lwu v1, -4(a0)            ;; (set! v1 (l.wu (+.i a0 -4))) tm: v1: type, a0: basic, a1: type
    lw a0, object(s7)         ;; (set! a0 object)            tm: v1: type, a0: type, a1: type
L286:
    bne v1, a1, L287          ;; (b! (!= v1 a1) L287 (set! a2 '#f)) tm: v1: type, a0: type, a1: type, a2: basic
    or a2, s7, r0

    daddiu v1, s7, #t         ;; (set! v1 '#t)               tm: v1: symbol, a0: type, a1: type, a2: basic
    or v0, v1, r0             ;; (set! v0 v1)                tm: v0: symbol, v1: symbol, a0: type, a1: type, a2: basic
    beq r0, r0, L288          ;; (b! '#t L288 (nop))         tm: v0: symbol, v1: symbol, a0: type, a1: type, a2: basic
    sll r0, r0, 0

    or v1, r0, r0             ;; (set! v1 r0)                tm: 
L287:
    lwu v1, 4(v1)             ;; (set! v1 (l.wu (+.i v1 4))) tm: v1: type, a0: type, a1: type, a2: basic
    bne v1, a0, L286          ;; (b! (!= v1 a0) L286 (nop))  tm: v1: type, a0: type, a1: type, a2: basic
    sll r0, r0, 0

    or v0, s7, r0             ;; (set! v0 '#f)               tm: v0: symbol, v1: type, a0: type, a1: type, a2: basic
L288:
    jr ra
    daddu sp, sp, r0

    sll r0, r0, 0
    sll r0, r0, 0

(seq b0 (until b4 (cond (b1 (return-from-function b2 b3)))) b5 b6)
;; ir
(begin
  (set! v1 (l.wu (+.i a0 -4)))
  (set! a0 object)
  (until
   (begin (set! v1 (l.wu (+.i v1 4))) (!= v1 a0))
   (if (= v1 a1) (return ((begin (set! v1 '#t) (set! v0 v1))) ((set! v1 r0))))
   )
  (set! v0 '#f)
  )
```

### DIFF
--- a/common/type_system/TypeSystem.cpp
+++ b/common/type_system/TypeSystem.cpp
@@ -107,7 +107,7 @@ std::string TypeSystem::get_runtime_type(const TypeSpec& ts) {
 /*!
  * Get information about what happens if you dereference an object of given type
  */
-DerefInfo TypeSystem::get_deref_info(const TypeSpec& ts) {
+DerefInfo TypeSystem::get_deref_info(const TypeSpec& ts) const {
   DerefInfo info;
 
   if (!ts.has_single_arg()) {
@@ -193,7 +193,7 @@ bool TypeSystem::partially_defined_type_exists(const std::string& name) const {
  * return type.
  */
 TypeSpec TypeSystem::make_function_typespec(const std::vector<std::string>& arg_types,
-                                            const std::string& return_type) {
+                                            const std::string& return_type) const {
   auto result = make_typespec("function");
   for (auto& x : arg_types) {
     result.add_arg(make_typespec(x));
@@ -205,28 +205,28 @@ TypeSpec TypeSystem::make_function_typespec(const std::vector<std::string>& arg_
 /*!
  * Create a TypeSpec for a pointer to a type.
  */
-TypeSpec TypeSystem::make_pointer_typespec(const std::string& type) {
+TypeSpec TypeSystem::make_pointer_typespec(const std::string& type) const {
   return make_pointer_typespec(make_typespec(type));
 }
 
 /*!
  * Create a TypeSpec for a pointer to a type.
  */
-TypeSpec TypeSystem::make_pointer_typespec(const TypeSpec& type) {
+TypeSpec TypeSystem::make_pointer_typespec(const TypeSpec& type) const {
   return TypeSpec("pointer", {type});
 }
 
 /*!
  * Create a TypeSpec for an inline-array of type
  */
-TypeSpec TypeSystem::make_inline_array_typespec(const std::string& type) {
+TypeSpec TypeSystem::make_inline_array_typespec(const std::string& type) const {
   return make_inline_array_typespec(make_typespec(type));
 }
 
 /*!
  * Create a TypeSpec for an inline-array of type
  */
-TypeSpec TypeSystem::make_inline_array_typespec(const TypeSpec& type) {
+TypeSpec TypeSystem::make_inline_array_typespec(const TypeSpec& type) const {
   return TypeSpec("inline-array", {type});
 }
 
@@ -466,7 +466,7 @@ void TypeSystem::assert_method_id(const std::string& type_name,
  * and how to access it.
  */
 FieldLookupInfo TypeSystem::lookup_field_info(const std::string& type_name,
-                                              const std::string& field_name) {
+                                              const std::string& field_name) const {
   FieldLookupInfo info;
   info.field = lookup_field(type_name, field_name);
 
@@ -764,7 +764,7 @@ int TypeSystem::get_next_method_id(Type* type) {
 /*!
  * Lookup a field of a type by name
  */
-Field TypeSystem::lookup_field(const std::string& type_name, const std::string& field_name) {
+Field TypeSystem::lookup_field(const std::string& type_name, const std::string& field_name) const {
   auto type = get_type_of_type<StructureType>(type_name);
   Field field;
   if (!type->lookup_field(field_name, &field)) {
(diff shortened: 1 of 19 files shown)


water111:
After looking some more, the 
```
    dsll v1, a1, 2            ;; (set! v1 (shl a1 2))
    daddu v1, v1, a0          ;; (set! v1 (+.i v1 a0))
    lwu v1, 16(v1)            ;; (set! v1 (l.wu (+.i v1 16)))
```
example may only occur when grabbing a method by index from a type, and not general dynamic array access.

The general array access with a variable index looks like
```
    addiu a0, r0, 12          ;; (set! a0 12)
    mult3 a0, a0, v1          ;; (set! a0 (*.si a0 v1))
    daddiu a0, a0, 100        ;; (set! a0 (+.i a0 100))
    daddu a0, a0, v0          ;; (set! a0 (+.i a0 v0))
    lw a1, *null-process*(s7) ;; (set! a1 *null-process*)
    sw a1, 0(a0)              ;; (s.w (+.i a0 0) a1)
```
for doing a store at least.

Sadly the same problem exists.

water111:
And a final example, proving that the instructions to access an array don't even have to come in order:
```
    addiu v1, r0, 12          ;; (set! v1 12)
->  daddiu a0, s5, -1         ;; (set! a0 (+.i s5 -1))
    mult3 v1, v1, a0          ;; (set! v1 (*.si v1 a0))
    daddu v1, v0, v1          ;; (set! v1 (+.i v0 v1))
    sw s7, 108(v1)            ;; (s.w (+.i v1 108) s7)
```

I believe this means that if your array index is some super complicated expression involving branching, the equivalent of the `addiu v1, r0, 12` will happen before all of that. This makes it impossible to recognize this as an array access unless you consider multiple basic blocks, which is impossible without expression compaction across basic blocks...

I think this would get expression compacted into:
```
(+ v0 (* 12 (- s5 1)) 108)
```
which could be recognized as 
```
(+ object (* array-stride <index-expr>) offset-of-array)
```
pretty easily.

This means the algorithm in `TypeAnalysis.cpp` can't work - it needs to operate on the full nested IR, after expression compaction. However, all of the other stuff should still work fine on the nested IR.

It may be useful to keep the old algorithm around - it could possibly be used on functions containing inline assembly that can't have a nested IR built yet, or to double check the results on functions which don't hit this array access issue.

water111:
Merging this because most of it can work with the new algorithm, and the old algorithm is still worth keeping around for debugging/double check.

PR #96, opened 2020-10-25
Amalgamate FileIO and remove duplicate Timer header

### LINKED ISSUE #8: Move file I/O utilities to a single library
Because this repo started as a few separate projects, there is some duplicated code for reading/writing files in `decompiler/util/FileIO.h` and `goal/util/file_io.h`.  These should be moved to a single common library in `common/util` that should probably get its own `CMakeLists.txt` file.

The code for `Timer` is also duplicated a bunch and could go in this library as well.

xTVaser:
Cleaning up this longstanding issue since it's a quick one.

Closes #8


water111 (approved):
Perfect, thanks!

PR #102, opened 2020-10-26
Add very ROUGH script to format lisp code via emacs

xTVaser:
Currently this works through an elisp script which finds all relevant files with the option of specifying which batch / and batch-size it should process.  That allows us to have multiple emacs processes formatting in parallel to cut-down the time.

TODO:
- We need to do a 2-pass format.  EMac's semantic seems to recognize `,` as whitespace no matter what.  In the case of CL/Scheme's quasicomma, this is problematic. 

Related to - https://github.com/water111/jak-project/issues/63
Resolves - https://github.com/water111/jak-project/issues/75


xTVaser:
I'm going to abandon this for now, even `srefactor`, while really nice in many respects, will make obvious formatting mistakes (folding code into a common on the previous line for example)

PR #106, opened 2020-10-30
Try to speed up the build

water111:


water111:
The first attempted saved ~8 seconds... not great.  I think we just have to expect ~4 seconds per file when we include a bunch of STL headers, spdlog, and fmt.

water111:
Linux went from 5 minutes to 3m 47s, windows went from 5m 53s, 4m 17s,

PR #107, opened 2020-10-31
Set up the compiler to ptrace the runtime

water111:
Add a library `xdbg` which should hide the details of cross platform debugging.
Implement ptrace attach/detach for linux only in `xdbg`
Make the runtime send the compiler all the info needed to attach.
Make the Listener remember loads / reset messages for the debugger.
Some very basic logic to attach/detach, and a simple test.

Needs more documentation and a better way to manage debugger state before merging. Maybe we want a `Debugger` object to hold all this state?


water111:
This adds a very basic debugging framework.  The C Kernel now sends some info to the compiler so the debugger can find the correct process. The compiler can now "attach" to a target with `(dbg)`, and break/continue with `(:break)` and `(:cont)`.  We can also read/write memory by GOAL address and get x86 GPRs, but this isn't exposed at the compiler REPL yet, other than a `(:dump-all-mem <file>)` that dumps all GOAL memory to a file for later debugging.

Right now this only works in Linux, but the platform specific stuff is stubbed out in `xdbg.cpp`.  The debugger tests only run in linux.

PR #109, opened 2020-11-01
Add some debugger memory utilities

water111:
Add utilities to print memory in several ways, including disassembling.

Load the symbol table on break, and allow using symbols (or their values) as addresses in debugging forms.

Build Zydis as a shared library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,9 @@ add_subdirectory(third-party/spdlog)
 # build zydis third party library for disassembling x86
 option(ZYDIS_BUILD_TOOLS "" OFF)
 option(ZYDIS_BUILD_EXAMPLES "" OFF)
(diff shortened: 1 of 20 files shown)


water111:
This will close https://github.com/water111/jak-project/issues/86

This implements reading/writing memory and setting breakpoints by address.  There is a basic disassembly but it needs some work to correctly find instruction boundaries and include some GOAL info as well.

There are options to attach, continue, break, disassemble, print memory, and set/clear breakpoints (by address) implemented for the compiler.  There are tests for attach/continue/break/read/write/breakpoints.

I didn't realize this until after I started writing code, but you can't seem to read/write registers from a different thread than the thread that starts the `ptrace`. So the logic for figuring out how to continue (different for breakpoints vs. manual breaks) is a bit convoluted (it's deferred until the next debugger operation from the main thread). If we have trouble with this, we should just move all `ptrace` stuff to the watcher thread.

PR #111, opened 2020-11-05
Added missing import causing runtime_error to not be found.

Nexushunter:
### Problem
Upon initial setup, there was an issue with `runtime_error` not being picked up. It turns out an import `<stdexcept>`was missing.

### Solution
Add missing import.


water111:
Good catch, thanks!

If you are interested, we have a discord here for discussing development here: https://discord.gg/z4ChyVPJcx

PR #115, opened 2020-11-07
[Source Line Debugger] Tracking objects and IR

water111:


water111:
Example of using the debugger, after doing `(lt)` to connect to the target and `(blg)` to build and load all code.

``` lisp

;; first, attach the debugger, which stops the target.
gc > (dbg)
[Debugger] PTRACE_ATTACHED! Waiting for process to stop...
Target has stopped. Run (:di) to get more information.
Read symbol table (160000 bytes, 227 reads, 226 symbols, 2.38 ms)
rax: 0x0000000000000000 rcx: 0x00007f6fc4943361 rdx: 0x0000000000000000 rbx: 0x0000000000147d24 
rsp: 0x00007f6fc483ac80 rbp: 0x00007f6fc483acc0 rsi: 0x0000000000000000 rdi: 0x0000000000000000 
 r8: 0x0000000000000000  r9: 0x0000000000000008 r10: 0x00007f6fc483aca0 r11: 0x0000000000000000 
r12: 0x0000000000147d24 r13: 0x00007ffe421c509f r14: 0x00007ffe421c50a0 r15: 0x00007f6fc483afc0 
rip: 0x000055e3dbbc4af7
Not in GOAL code!
Debugger connected.

;; set a breakpoint at the value of the last symbol, which is the last function.
;; the sym-val is a special thing used in the debugger which can look up symbol values
;; from the symbol table of the stopped target.
gcs> (:bp (sym-val last))
;; resume!
gcs> (:cont)
;; call a function which calls last.
gcr> (print (last (list 'a 'b 'c 'd)))
Target has stopped. Run (:di) to get more information.

;; we trigger an ack timeout because the runtime hit a breakpoint and won't send the ack message.
;; we should suppress the ack timeout when we've detected that we hit a breakpoint.
  Error - target has timed out. If it is stuck in a loop, it must be manually killed.
Runtime is not responding. Did it crash?

;; display info
gcs> (:di)
Read symbol table (160512 bytes, 231 reads, 230 symbols, 2.04 ms)
rax: 0x0000000000169a32 rcx: 0x0000000000000000 rdx: 0x0000002000169a34 rbx: 0x0000000000161544 
rsp: 0x0000002007ffff88 rbp: 0x0000002000162e64 rsi: 0x0000000000169a30 rdi: 0x0000000000169a32 
 r8: 0x0000002000160ad4  r9: 0x0000002000000000 r10: 0x0000000000147e64 r11: 0x0000000000160a94 
r12: 0x0000000000147d24 r13: 0x0000002007ffbd64 r14: 0x0000000000147d24 r15: 0x0000002000000000 
rip: 0x0000002000162e65
In function last in segment 0 of obj gcommon, offset_obj 0x5b5, offset_func 0x1
- [0x2000162e64] jmp 0x0000002000162E71                  goto label-4
  [0x2000162e69] movsxd rax, dword ptr [r15+rdi*1+0x02]  mov igpr-2, [igpr-0 + 2]
  [0x2000162e6e] mov rdi, rax                            mov igpr-0, igpr-2
  [0x2000162e71] movsxd rax, dword ptr [r15+rdi*1+0x02]  mov igpr-3, [igpr-0 + 2]
  [0x2000162e76] mov rcx, r14                            mov-symptr igpr-4, '_empty_
  [0x2000162e79] add rcx, 0xFFFFFFFFFFFFFFF6
  [0x2000162e80] cmp rax, rcx                            j(igpr-3 != igpr-4) label-2
  [0x2000162e83] jnz 0x0000002000162E69
  [0x2000162e89] mov eax, [r15+r14*1]                    mov igpr-5, '#f
  [0x2000162e91] mov rax, rdi                            ret igpr-1 igpr-0
  [0x2000162e94] ret

;; remove the breakpoint
gcs> (:ubp (sym-val last))
;; continue execution
gcs> (:cont)
;; the result is printed!
(d)
1481218

;; oops, extra print. This is the ack we timed out on earlier, showing up way late.
[Listener] Got an ack message when we weren't expecting one.

gcr>
```

water111:
This tracks the loading of object files and stores debug info for where functions, instructions, and IR are in an object file. The runtime now sends better messages to the compiler so segments can be tracked.  Tracking stuff in `*listener*` and in top level segments is still a little bit iffy, but main/debug segments of object files works properly.

The debugger can show which function and IR you have stopped in, like in the example above.

Segfaults now stop properly in the debugger.
Resuming from a breakpoint now works correctly (previously would result off by one byte).
Code / object generation creates debug info for functions, IR, and x86 instructions in a segment.
Functions now have unique names within object files for purposes of identifying them to the debugger.
Tests make sure that all functions in `gcommon` can be disassembled by Zydis and x86 instructions / IR can be matched up. Also we test that stopping by breakpoint and finding the function leads to a successful disassembly that matches instruction boundaries from the debug info.


The next step is to connect IR to source lines and better utilities for getting addresses of functions/methods, which will be a change to the compiler front end.

On the lower-level side we need to support stack frames in the debugger and also hook up the register allocator to the debug info.

PR #118, opened 2020-11-11
Compiler logging

doctashay:
Adding more spdlog functionality to the compiler, with an emphasis on logging both to the terminal and a separate log file in duplicity, for easy review later. 

Still very much a work in progress.


doctashay:
spdlog is no longer required for the project. Closing.

PR #125, opened 2020-11-17
[Reader] Support escape codes for any character

water111:
Game text has weird non-printable characters that the reader can't handle. Now you can do something like `"\cab"` and it will insert the character for `0xab`. This involves fiddling with the string to number and number to string stuff, which is always a pain on Windows so I'm making this a separate PR for testing.


water111:
coveralls didn't work.

PR #142, opened 2020-11-28
Type Prop - Second Attempt

water111:
WIP second version of type analysis decompiler pass.


water111:
It succeeds everywhere in gcommon, excluding the 128-bit inspects, the debug register setting function, and the asm memcpy function.

Current stats:
```
1078/10179 functions attempted type analysis (10.59%)
618/1078 functions that attempted type analysis succeeded (57.33%)
618/10179 functions passed type analysis (6.07%)
```

PR #145, opened 2020-11-29
Fix and rename rgba16_to_rgba32

whiteh0le:
This fixes the conversion from rgba16 to rgba32.
I also renamed the function because according to the _GS User Manual, Section 2.3.1_ that _Color Data Format_ it's called `rgba16` and not `rgba15` hope you don't mind.


water111:
This is perfect, thanks!  The name change is a good idea too.

PR #146, opened 2020-11-29
[Compiler] Bitfield Types

water111:
Add bitfield types (for now, 64-bit only).

- [x] Fixed shifts
  - [x] Codegen
  - [x] IR
  - [x] Front-end (don't really need this)
- [x] Add a `Type`
- [x] Support in `TypeSystem`
- [x] Support in `deftype`
- [x] Support casting integers to bitfields
- [x] Support accessing bitfields
- [x] Static bitfields (how do these work? stored in memory or inline in code?)
- [x] Set a bitfield. And test these cases
  - [x] Lexical
  - [x] Global
  - [x] Field of type


water111:
md5 checksums of files, before I mess with the compiler a bunch:
```
770dee114b4360da85c78da50b831e1b  ./GAME.CGO
a5e40b0360510dbc609ccb9495ed0c85  ./KERNEL.CGO
```

water111:
Scary changes made to set in https://github.com/water111/jak-project/pull/146/commits/27b80ac23ea39fde7c58971ea456fe3ed511a38f
The files have the same checksum, so it should have no effect on existing behavior.

PR #147, opened 2020-12-01
Add some compiler features and documentation

water111:
Turn on type checking for "memory places", but add an exception for putting a signed integer in an unsigned integer. Fix up some small type issues (pointers vs. uint32).

Add a debug option to compile and display assembly of the result.

Update the GOAL doc.


water111:
Making some progress toward reasonable error messages!  It's not perfect, but it's so much better than before:
```
  > (m "goal_src/kernel/gcommon.gc")
-- Compilation Error! --
The symbol asdf was looked up as a global variable, but it does not exist.
Form:
asdf
Location:
?
Code:
asdf
Location:
?
Code:
(begin (label gensym9) asdf (if (or (eq? a b) (zero? a)) (return-from #f #t)) (s...
Location:
text from /home/water/jak-project/goal_src/kernel/gcommon.gc, line: 256
  (until (eq? a object)
Code:
(until (eq? a object) asdf (if (or (eq? a b) (zero? a)) (return-from #f #t)) (se...
REPL Error: Compilation Error
```

PR #149, opened 2020-12-04
Basic Inline Assembly

water111:
This still needs some tests and some work around register allocation. 

To get a "register", you can define it inside `rlet`, which is like a normal `let` but:
- The initial value is optional.  If none is provided, you get whatever happens to be in the register.
- You can specify the type (optional). By default it's the type of the initial value, or `object` if there's no initial value.
- You can specify the register (optional).
- You can specify the register "class" (gpr/xmm), (optional)

It will behave like a normal variable and use normal GOAL coloring tricks, but can only live in that given register and can't be spilled to the stack ever.  If it goes dead inside the `rlet`, the GOAL coloring system can use it as a temporary in any high level GOAL code. If you don't want this to happen, don't write higher level GOAL code in your `rlet`. This is usually not an issue because asm ops should interact correctly with the coloring system by default.

For even more control, use an `asm-func` a GOAL function where you have to do everything yourself. There's no prologue or epilogue.  You can still write GOAL code, but the compiler shouldn't use the stack or any callee saved registers, and as a result some stuff can fail if you run out of temporary registers.

We should consider a "nuclear option" to completely bypass the coloring system and access registers. In either case, this is a dangerous thing to do and can mess with any higher level GOAL expression.  Currently this is by setting `:color #f` inside of the asm operation, but it isn't clear that it works in a useful way yet.

It's not clear what should happen if you manually use a callee-saved register. I think it should by default work with the coloring system. It's too annoying to have a special mode for these where they can be used, but not default allocated, and not backed up only in asm-funcs. So it'll get backed up in a normal function, and error in an asm-function (unless you use the `:color #f` option). 

Test ideas:
- Support accessing registers that are "forbidden" in coloring (rsp, r15/offset)
- Allow constraining a register and not using it (it's annoying when debugging if this errors)


water111:
I think I finally found a solution I like. It's heavily biased toward places where we'll need to do MIPS -> x86 assembly translation, but still allows us to very easily access non-standard registers or do things manually/without the stack for a few kernel functions.  The goal doc has been updated to describe this behavior in more detail.

PR #150, opened 2020-12-04
Implement `gkernel`: Part 1

water111:
It's time to implement `gkernel.gc` in OpenGOAL!

This is going to have a lot of differences between the actual game and the port because we need to modify the kernel to work for x86-64.  But the non-assembly parts should be pretty similar.  They will be a good way to check the decompiler's correctness because they have a good combination of loops/branching/conditionals, etc...

For now, this is just the start, and I'm just making the PR so I can easily do a diff against master to look at the compiler changes.

Compiler changes:
- Better support for constants. Integer constants can now be GOAL constants, enum constants, etc instead of just plain numbers.
- Bugfix on `'#f` / `'#t`, code was correctly but wasn't as fast as it could be.
- Typechecking for `set!` allows `#f` to be used as a null value.

### DIFF
--- a/boot_kernel.sh
+++ b/boot_kernel.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# Directory of this script
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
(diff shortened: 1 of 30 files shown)


water111:
This PR is getting too large, so I'm going to do this as a Part 1 + a separate compiler PR + a Part 2.

PR #154, opened 2020-12-06
Improve Register Allocator

water111:
Faster and better on inline assembly.

Baseline before speed improvements:
Starting the compiler, building game 3x takes 4.6 seconds on my computer.

MD5's at ac52c7bca3a34d3fc316875cbff799d85362352c
```
a0c2a6111602300db6f441bf983040a5  iso/KERNEL.CGO
c2f58039fc824702ce903f4d76d5a142  iso/GAME.CGO
```

### DIFF
--- a/decompiler/config/all-types.gc
+++ b/decompiler/config/all-types.gc
@@ -189,6 +189,7 @@
 (declare-type process basic)
 (declare-type stack-frame basic)
 (declare-type cpu-thread basic)
+(declare-type state basic)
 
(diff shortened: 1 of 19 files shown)


water111:
Run time was decreased to 2.6s with exactly the same output. At this point, it's probably worth looking into other parts of the compiler or revising the register allocation algorithm if more speed is needed.

For fun I compiled the compiler with full optimization and this runs in 0.5 seconds!

PR #156, opened 2020-12-10
Add another kernel test, fix small bugs

water111:


water111:
The windows test failures found a very real bug in the kernel implementation that the linux tests didn't catch:
```
(set! temp sp)   ;; store stack pointer in temp
(.sub temp off)  ;; subtract off GOAL pointer offset
;; incorrect code: 
 (set! (-> obj sp) (the int sp)) ;; store GOAL pointer to stack
;; correct code:
 (set! (-> obj sp) (the int temp)) ;; store x86-64 pointer to stack
```

This happened to work on Linux because `(-> obj sp)` is a 32-bit integer, and the location of the GOAL main memory happened to be one where you can just mask off the upper 32-bits of an x86-64 address to get the GOAL address.  On windows this didn't work, because the offset wasn't all zeros in the lower 32 bits.

To make these bugs easier to find in the future, I've changed the address hint passed into `mmap` to one that will cause problems in these cases.

water111:
looks like an identical bug in `Debugger.cpp` with the address used for breakpoints.

PR #157, opened 2020-12-13
[Decompiler] Experimental Expression Stack

water111:
Adds a very experimental expression stack algorithm as a totally separate compile pass from function analysis, and fix up some small issues with type propagation.

There are some implementation issues, so it will probably be mostly rewritten at some point.

Most of the important information, like `consumes`, types, and register use are in terms of `IR_Atomic` operations. But we want to support the expression stack operation on non-atomic IR. So it is probably worth making a better interface for the information the expression stack needs and make sure it can be done on all IR.

Currently the expression building goes in an inside -> outside order, but this will have issues depending on where things are split for stuff like computing the condition of `if`s.  So going in the "compile order" of outside -> in might be a better approach?

The expression logic is the perfect place to convert multi-op loads into a single large expression. This means we'll need type information in the expression stack, and some way to detect and simplify these load expressions.

### DIFF
--- a/common/type_system/TypeSpec.h
+++ b/common/type_system/TypeSpec.h
@@ -51,6 +51,7 @@ class TypeSpec {
   size_t arg_count() const { return m_arguments.size(); }
 
   const TypeSpec& get_arg(int idx) const { return m_arguments.at(idx); }
+  TypeSpec& get_arg(int idx) { return m_arguments.at(idx); }
   const TypeSpec& last_arg() const {
     assert(!m_arguments.empty());
     return m_arguments.back();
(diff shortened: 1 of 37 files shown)


water111:
windows build failed because nasm website is down.

water111:
This PR is all over the place, I didn't want to merge it until windows tests were working again.

The end result is that it adds a very experimental expression stack algorithm.  It will need some significant changes to work properly when there are expressions with branching, like `(my-func (if a b c))` or similar.  Starting on this brought up some more issues with the type propagation algorithm, so it's in the process of being redone. The plan is to get the type propagation algorithm working 100% on KERNEL.CGO before coming back to the expression stack.

PR #162, opened 2020-12-19
[Compiler] Add static data features

water111:
Add support for pointers to other data in the same segment in the linker (`LINK_PTR`), in `ObjectGenerator` and in `StaticStructure`.

In the compiler, add static pairs, linked lists, etc. For now, it supports quoting only (no quasiquote), so you can't put in GOAL expressions. The only supported elements are: more pairs/lists, strings, symbols, and integers, which are signed 32-bit integers.

Make `StaticString` a child of `StaticBasic`, which makes including strings are pair elements, and eventually including strings a fields much easier.

Add the `StaticResult` class. It's kind of like a `RegVal`, but works with the static part of the compiler.  It would have been really cool if the static stuff could use `Val` like all the rest of the front end, but it seems like there would need to be too much special-casing. A `StaticResult` can be used to store an integer/float/bitfield-type-constant, reference, symbol, but not stuff like "the entire data of a type".

### DIFF
--- a/common/link_types.h
+++ b/common/link_types.h
@@ -14,6 +14,7 @@ enum LinkKind {
   LINK_TYPE_PTR = 2,                  //! link a pointer to a type.
   LINK_DISTANCE_TO_OTHER_SEG_64 = 3,  //! link to another segment
   LINK_DISTANCE_TO_OTHER_SEG_32 = 4,  //! link to another segment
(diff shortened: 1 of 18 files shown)


water111:
This adds support for many other statics, including constants, strings, integers, floats, enums, pairs, other static structures, inlined static structures. We also do full type checking on the fields.

PR #171, opened 2020-12-29
[Compiler] In-progress support for vector float (AVX)

water111:
Add basic support for vector floats.

- [x] Add support in `IGen` for register->register moves
- [x] Add support in `IGen` for usual GOAL loads and stores
- [x] Add support in `IGen` for `vmulps`. This is a common format of instruction.
- [x] Add support in `IGen` for `vshufps`.
- [x] Add support in `IGen` for RIP-relative loads and stores
- [x] Ability to load constant values (from a constant pool?)
- [x] Clean up the two different `RegKind`s. Maybe a `RegClass` and a `RegKind`?
- [x] Add support in `rlet` for creating a `vf` register
- [x] Add support for load and store assembly operations in the front-end and IR
- [x] Add support for basic 3-op assembly operations in the front-end and IR
- [x] Add support for blend operations in the front-end
- [x] Make sure coloring system is aware of `vf` registers. (Doesn't have to color them, just not clobber them)
- [x] Implement a few functions in `vector` or `vector-h` as an example (may have to add more operations).
- [x] Documentation for new asm forms.


water111:
Getting close. The generated code for loading a constant vector is really bad:
```
(rlet ((vf1 :class vf :reset-here #t))
      (.lvf vf1 (new 'static 'vector :x 1.0 :y 1.2 :z 1.4 :w 1.6))
      )
```

becomes 
```
- [0x10000] lea rax, [0x0000000000010007]          mov-sva igpr-1, static-structure
  [0x10007] sub rax, r15
  [0x1000a] vmovaps xmm0, [r15+rax*1]              mov ivf-0, [igpr-1 + 0]
```
it seems like we should support directly loading a static with `.lvf`.

water111:
Now it is a single instruction (not linked, so the offset is wrong):
```
[0x10000] vmovaps xmm0, [0x0000000000010008]     mov-svl ivf-0, [static-structure]
```

PR #175, opened 2020-12-31
[Compiler] Improve spills and register backups

water111:


water111:
This reduced the size of `KERNEL.CGO` from 84,672 bytes to 74,128 bytes!

PR #181, opened 2021-01-04
[Decompiler - New IR] Add AtomicOp

water111:
This PR adds a new `AtomicOp`. This will replace `IR_Atomic` in the new decompiler IR.

There are a few other changes:
- Added `InstructionParser` a utility to parse text containing MIPS assembly source code. This is intended to be used to generate test cases for future IR2 work.
- Made decompiler a library and cleaned up the build.
- Move decompiler into its own namespace to avoid conflicts between compiler/decompiler classes with the same name.
- Instead of passing `LinkedObjectFile` everywhere we may need label names, instead pass a `const std::vector<Label>&`.


water111:
It seems like there is a real problem of having both the decompiler and the compiler code in the test binary. They have classes with the same names. The ASan build found a real bug, and I found a bug recently with `Label`, which exists in both as well.

I will try to put the decompiler IR in a namespace to avoid this problem.

water111:
The Codacy failure is because there are "always true" conditions in the `AtomicOpBuilder` which isn't finished yet.

PR #189, opened 2021-01-10
Typo fixes & Windows QoL changes

ManDude:
Fixes some typos in some files and adds batch scripts for the decompiler/runtime/compiler. Also fixes a crash on Windows with the logging utility due to non-existent directory.


water111:
Thanks!

PR #190, opened 2021-01-10
Improved test namings for multi-operations, added new ones, still a few failures

xTVaser:
Current failures seem to mostly be off-by-one register counts involving shift operations.

These are the more interesting ones:
```
..\..\..\test\decompiler\test_AtomicOpBuilder.cpp(59): error: Expected equality of these values:
  op->to_string(prg.labels, &env)
    Which is: "(set! a3 (%.ui a1 a2))"
  output_lines.at(i)
    Which is: "(set! a3 (mod.ui a1 a2))"
[  FAILED  ] DecompilerAtomicOpBuilder.DIVU_MFHI (3 ms)
..\..\..\test\decompiler\test_AtomicOpBuilder.cpp(59): error: Expected equality of these values:
  op->to_string(prg.labels, &env)
    Which is: "(set! a3 (%.si a1 a2))"
  output_lines.at(i)
    Which is: "(set! a3 (mod.si a1 a2))"
[  FAILED  ] DecompilerAtomicOpBuilder.DIV_MFHI (1 ms)
..\..\..\test\decompiler\test_AtomicOpBuilder.cpp(59): error: Expected equality of these values:
  op->to_string(prg.labels, &env)
    Which is: "(set! a1 (shl a2 a3))"
  output_lines.at(i)
    Which is: "(set! a1 (sll a2 a3))"
```


water111:
For the ones in your comment, I think these are just typos in the ops. I think we should change `AtomicOp.cpp` so the operations are called `sll` and `mod`.  In GOAL, the function is called `mod`, not `%`.  The `shl` (x86 instruction name) vs. `sll` (MIPS instruction name) is less clear, but I would go with `sll` to be consistent with everywhere else. 

I think at least some of the other failures are because some registers are duplicated in the register info.  I think this is a real bug and I'll work on a fix.

water111:
The register counts issue was a real bug and should be fixed in https://github.com/water111/jak-project/pull/191

PR #195, opened 2021-01-14
[Decompiler] WIP Conversion to SSA and variable naming

water111:
Unfortunately, the conversion to SSA is turning out to be relatively slow. For now, I'm just focusing on getting the right answer with an algorithm that's not doomed to be impossibly slow, without worrying about implementation details.

After that, it's probably worth going back and improving the implementation. We can easily map "GOAL registers used in coloring allocator" to "integer between 0, 64", so we can use a 64-bit integer where each bit corresponds to represents sets of registers instead of a `set`/`unordered_set`. I did bitsets in the compiler (https://github.com/water111/jak-project/pull/154) and it worked great, though it was more complicated because you can have arbitrarily many variables in the compiler.

The algorithm is based off of _Simple Generation of Static Single-Assignment Form_ by John Aycock and Nigel Horspool, but there are some differences.  The "Improvement 1" has been modified so that the specially numbered variable is the variable assigned by the phi function, not the variables going into it.  We insert arguments to phi functions in the preceding blocks, and insert the phi function itself in the block that contains it.

### DIFF
--- a/common/type_system/TypeSystem.cpp
+++ b/common/type_system/TypeSystem.cpp
@@ -1083,7 +1083,7 @@ bool TypeSystem::typecheck_base_types(const std::string& expected,
 /*!
  * Get a path from type to object.
  */
-std::vector<std::string> TypeSystem::get_path_up_tree(const std::string& type) {
+std::vector<std::string> TypeSystem::get_path_up_tree(const std::string& type) const {
   auto parent = lookup_type(type)->get_parent();
   std::vector<std::string> path = {type};
   path.push_back(parent);
@@ -1101,7 +1101,7 @@ std::vector<std::string> TypeSystem::get_path_up_tree(const std::string& type) {
 /*!
  * Lowest common ancestor of two base types.
  */
-std::string TypeSystem::lca_base(const std::string& a, const std::string& b) {
+std::string TypeSystem::lca_base(const std::string& a, const std::string& b) const {
   if (a == b) {
     return a;
   }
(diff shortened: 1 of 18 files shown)


water111:
At this point, converting to "rough" SSA and simplifying as in the paper seems to work correctly.  Even this bad implementation isn't that slow compared to the type analysis pass, so I'm no longer worried about speed, at least for this PR.  It's likely that doing all 10k functions will take < 10 seconds, which is fast enough.

The variable finding part is partially implemented. We safely convert out of SSA by eliminating all phi functions.  Then we remap the remaining SSA entries as needed, using a variable ID of 0 to represent the initial value of the register coming into the function.  This form should split registers into different variables as much as possible.  It might be the case that this is the final assignment of variables we want. Although this may have "too many" variables, the expression propagation pass will eliminate small temporaries anyway so we don't need to worry about these. At least in the KERNEL.CGO, this looks pretty reasonable, though you can imagine weird situations where this wouldn't work well. 

Example: if you increment an integer like a loop counter, it works well.  But if you increment an integer outside of a loop, it will turn that into something like `j = i + 1` and then nothing else will use `i` after that.  In a language like GOAL where all variable declarations are scoped, this could become ugly. But it seems like people don't increment integers outside of loops very often, so we might be able to get away with this.  It seems like ghidra has the same limitation.  Of course, we can default to merging variables of the same type like this, but you would then get stuff like `i = j + k` where `i`, `j`, and `k` really should be separate variables, will come out like `i += k`. In the end, adding too many variables seems like the better option by default - if you're naming local variables you'll never have to give a single name to a variable which was originally two variables in the source.

The next step will be to determine the type of each variable, and give it a more pleasant name.  Ideally the name contains:
- something to indicate it's a local variable (`var`?)
- The type of what it holds
- The register it was based on (maybe? or maybe once we trust this pass more we don't want this?)
- If this is still ambiguous, add some integer to the end.
- Can be easily parsed. Eventually we might want a variable renaming GUI.
Example:
`var-string-v0-0`

water111:
The basic implementation is finished.  Left to do:
- [ ] Manually check a few places for bugs
  - [ ] Check case where instruction reads and writes the same register, but different variable, in the same instruction.
  - [ ] Check for edge cases that don't work well.
- [ ] Propagate the idea of `VarInfo` up a level to the `Env` instead of just storing string names. This will be useful later on.
- [ ] Improve the printing
  - [ ] Use much shorter variable names: `v0-3` is  better than `v0-(function (pointer uint32) string kernel-context none)-12`
  - [ ] Print out a table of local variables and their types in the decompilation output

PR #200, opened 2021-01-18
[Decompiler] Test framework for decompiler regression tests and gcommon tests

water111:


water111:
Having this type of test seems very valuable, and I already found a few bugs when creating this.  The biggest issue with these tests is it's really confusing to understand the output before all the expression stuff has been properly built (stuff like "hidden variables" set in branch delay slots, the ordering of body/condition in loops, etc).  But I think it's still worth it to come up with a set of test functions that we can test at various points in the decompiler, and it will only get easier to add more tests as the decompiler progresses.

PR #209, opened 2021-01-23
[Decompiler] Begin expression conversion, rearrange tests

water111:


water111:
I ended up rearranging a bunch of test stuff instead of doing much work on expressions. Merging this as a separate PR so the expression stuff is easier to look at.

PR #211, opened 2021-01-24
[Decompiler] Expression Building

water111:
- [x] Figure out how to fix https://github.com/water111/jak-project/issues/210
- [x] Figure out the best way to apply load-rewriting, and get it working for the `bfloat` example.

### DIFF
--- a/decompiler/Function/BasicBlocks.h
+++ b/decompiler/Function/BasicBlocks.h
@@ -3,9 +3,6 @@
 #include <vector>
 #include <memory>
 
-#include "CfgVtx.h"
-#include "decompiler/util/DecompilerTypeSystem.h"
-#include "decompiler/util/TP_Type.h"
 // for RegSet:
 #include "decompiler/analysis/reg_usage.h"
 
@@ -16,7 +13,6 @@ class Function;
 struct BasicBlock {
   int start_word;
   int end_word;
-  TypeState init_types;
 
   // [start, end)
   int start_basic_op = -1;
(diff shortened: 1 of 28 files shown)


water111:
In the process of fixing https://github.com/water111/jak-project/issues/210, I noticed that return variables don't have liveliness computed correctly. I fixed this with a hack so the register usage algorithm thinks there's a "fake instruction" at the end of the last block that just reads `v0`.  This works even in the case with jumps straight to the end because we insert a 0-length block at the end of the function, which can still have this fake instruction applied. It's a bit hacky, but it's totally contained in `reg_usage.cpp`, so it seems fine.  In the case where we don't know the return type of the function, we don't read `v0`.

A consequence of fixing this is that all return values end up in the same SSA variable.  This is kind of satisfying, and gives me a nice way to verify that the fix actually works. I don't think this matters in the case with multiple return expressions, as these variables will be eliminated.  But if you have code like:
```
...
(return-from #f x)
...
(return-from #f y)
...
```
It may end up forcing `x` and `y` to be in the same variable.  (Note - this only applies in the case where `x` and `y` are just plain variables with nothing else).  This may not be ideal, but I think it's rare enough to ignore, at least until we can find a case where it actually occurs.  It seems possible to not split here with some tricks in the SSA phase, but I'd want to have some more test cases before trying this.

water111:
Currently, there is no way to extract register use information for registers used in rewritten branch conditions that become `ConditionElement`s.  This seems like it can't be done from looking at a single instruction, so perhaps the cfg pass should figure this out and store this info in the `ConditionElement`.

water111:
Codacy itself is failing:
image

Seems like there is some race condition in quickly updating the branch.

PR #212, opened 2021-01-25
Add support for macOS (rough)

doctashay:
Incomplete and inconsistent support for macOS, with GitHub Actions integration. 

Will take some time to complete.


doctashay:
Recently took another crack at this. We're still having a few issues that prevent us from fully building on Mac. Here's the good news:
-Decompiler builds fine, but it doesn't run properly due to a failed assertion:
`Assertion failed: (gLogger.fp), function set_file, file /Users/shale/Documents/jak-project/common/log/log.cpp, line 85.`
Should be easily fixable after making sure macOS is handling file paths correctly. 

-DGO packer and unpacker build and run with no issues.

Now, the roadblocks we're facing:
-mmap related functions need to use the proper header on macOS, I believe this is <sys/mman.h>. If this is already declared, something else must be changed for these functions to work.

-Our existing OpenGL implementation will almost certainly break on macOS and need to be adapted. After resolving the file pathing issues and mmap functionality, this will be the next most pressing issue to address.

game/runtime.cpp
```diff
@@ -121,6 +124,14 @@ void ee_runner(SystemThreadInterface& iface) {
                   MAP_ANONYMOUS | MAP_PRIVATE, 0, 0);
   }
 
+  #elif __APPLE__
```
xTVaser, line 127: good callout by codacy here -- maybe this is your memory issue?

doctashay: The build hasn't quite made it to the runtime yet, this issue occurs within the compiler. I don't think Codacy caught any similar typo in the compiler, while it should prevent issues down the line which is great it won't solve this particular build issue. I haven't even checked to see if I define <sys/mman.h> in the compiler yet so I'll take a look at that next.

common/cross_os_debug/xdbg.h
```diff
@@ -18,7 +18,7 @@
 #endif
 
 namespace xdbg {
-#ifdef __linux
+#if defined(__APPLE__) || defined(__linux_)
```
water111, line 21: should be `__linux__` (two underscores at the end)

common/util/BinaryWriter.h
```diff
@@ -4,12 +4,13 @@
  * @file BinaryWriter.h
  * Write raw data like a stream.
  */
-
+#include <unistd.h>
```
water111, line 7: `unistd` will have to be mac/linux only, it's not on windows. I don't think there's anything in here that needs `unistd.h`, maybe you could add `#include <cstdio>` instead? (I'm assuming the issue was with `fopen`/`fwrite`/`fclose`, which it's better to use `cstdio` because that works on mac/windows/linux)

decompiler/util/DataParser.cpp
```diff
@@ -3,6 +3,10 @@
 #include "DataParser.h"
 #include "third-party/fmt/core.h"
 
+#ifdef __APPLE__
+#include <unordered_map>
```
water111, line 7: you don't need the `__APPLE__` around this - I should have put the `#include <unordered_map>` but it happened not to matter on linux/windows because some other header brought it in.

decompiler/util/TP_Type.h
```diff
@@ -7,6 +7,10 @@
 #include "common/common_types.h"
 #include "decompiler/Disasm/Register.h"
 
+#ifdef __APPLE__
+#include <unordered_map>
```
water111, line 11: same here, I should have had this in the first place. you can remove the `__APPLE__` check

game/runtime.cpp
water111: I think we should try to avoid having an apply specific thing here. Mac OS is a unix, it should support `mmap`.

third-party/11zip/extlibs/minizip/src/gzread.c
```diff
@@ -4,7 +4,7 @@
  */
 
 #include "gzguts.h"
-
+#include <unistd.h>
```
water111, line 7: this might need to be linux/mac only.

water111 (commented):
Looks good so far!

doctashay:
As a quick addendum to this, after making even with master, here is the list of projects we can actually fully build with no issues on macOS. I have not tested any of these applications at runtime so I'm sure there will be other issues we run into after we fix compilation bugs. 
```
fmt
replxx
libzstd_static
lzokay
common
decompiler
glfw
imgui
minizip
elzip
memory_dump_tool
dgo_packer
dgo_unpacker
level_tools
level_dump
```
As expected, since they are the largest components of the project, the compiler and runtime both run into issues. `decomp` also appears to be failing after merging the latest master branch, like something to do with the PrettyPrint system. Since these don't build, our offline_tests won't build either. It's been a few months since I last looked at this and I'm seeing all sorts of really simple syntax issues like water is pointing out. Taking it one step at a time to narrow down issues.

decompiler/util/TP_Type.h
doctashay: Perfect! I wasn't sure on the proper implementation for this

common/util/BinaryWriter.h
doctashay: It looks like there are some other cases where `unistd.h` was implemented over `cstdio`  that I don't remember implementing, `common/cross_sockets/xsocket.cpp#L11` for example. Are there certain cases where we need `unistd.h` specifically or would it be safe to use cstdio to replace everything?

third-party/11zip/extlibs/minizip/src/gzread.c
doctashay: How good of a practice is it to adapt a third party library to fix issues locally? Should I consider submitting a PR for this to the original repo, that way we could just pull it down?

common/util/BinaryWriter.h
water111: There are some things that are only in `unistd.h`, like `usleep` or `close`, so we do have to use it some cases.  In `xsocket` we need it for `close`, which closes a unix file descriptor (for the unix sockets)

third-party/11zip/extlibs/minizip/src/gzread.c
water111: if it's an actual issue with the library, it's fine to fix it.  I already made changes here:
https://github.com/water111/jak-project/pull/508/files#diff-d0cb7382079eadc966c157222d19e2e15cf484ac5b6fbb56e1af0868b0a32879R169
to fix this problem on linux, you should probably modify that line to include `unistd.h` on mac os as well.

xTVaser:
In the effort of cleaning up our PR list, I'm gonna close this for now, we've talked a few times about what has to be done to get something like this merged -- hopefully we can get all of the pieces of requirements for macOS support together one day and merged and this PR can be re-opened.

doctashay:
Yeah, agreed this is probably best for now, since I am a bit short on free time to work on this. We have quite a few pieces of the puzzle together and I'll re-open it when I have more time

PR #215, opened 2021-01-27
[Decompiler] Add more expressions (Part 4)

water111:


water111:
So far, all functions are looking good. Planning to fix https://github.com/water111/jak-project/issues/214 and https://github.com/water111/jak-project/issues/216 before adding any more features.

PR #218, opened 2021-01-27
[Decompiler] Expression 5

water111:


water111:
Before merging this, I need to fix the issue with an empty list of allowed object files when running the decompiler on everything.

PR #219, opened 2021-01-28
[Tools] Add DGO packer and unpacker

water111:


water111:
Codacy is failing because it doesn't like having a `_` in a filename which is stupid.

PR #221, opened 2021-01-29
Compiler - Implementing more VU Instructions (Part 1 of 2)

xTVaser:
Commits describe which instructions have been implemented, they cover the vast majority of decompiled usages (75%+), and the bulk of the remainder will be quickly covered with the addition of the ACC register equivalent.

I only focused on VU instructions that are actually spat out by our current Jak 1 black-label decompilation, so I wouldn't be surprised if later games require us to add support for a few more, but that's a later concern.

### What I have left to do before im happy with this PR
- I did very rough test coverage to verify the ASM instructions, but i want to add comprehensive tests
- Ideally, removing some of the code duplication where possible, it's hard to remove the `Atoms.cpp` map duplication for reasons ive put in TODO comments.

### For this part, I'm leaving out instructions that have to do with the following:
- ACC Register (the bulk of the remainder)
- Q Register (division / sqrt)
- R Register
- Fixed point conversions (both ways)
- VCLIP (main point of the instruction seems to be setting the `clipping flag`.  Do we support such flags yet?
- Instructions that interact with VU Memory

### So which instructions are left?
- Accumulator ones
  - `V[ADD|SUB|MUL]a[x-z].dest`
  - `V[ADD|SUB|MUL][a|aq].dest`
  - `VM[ADD|SUB|MUL]a[x-z].dest`
  - `VM[ADD|SUB|MUL][a|aq].dest`
  - `vopmsub.dest`
  - `vopmula.dest`
- Q Register
  - `vdiv`
  - `vsqrt`
- R Register
  - `vrget`
  - `vrnext`
  - `vrxorw`
- Flag support needed
  - `VCLIP`
- Uses VU Memory
  - `vsqi.dest`
  - `vlqi.dest`
- Integer related
  - `vftoi0` | `vftoi12` | `vftoi4` | `viaddi` | `viand` | `vior` | `vitof0` | `vitof12` | `vitof15` | `vmtir`
    - I think it was discussed that these could be handled not through asm instructions, but through some C?


goalc/compiler/IR.cpp
```diff
@@ -1058,6 +1058,27 @@ void IR_AsmRet::do_codegen(emitter::ObjectGenerator* gen,
   gen->add_instr(IGen::ret(), irec);
 }
 
+///////////////////////
+// AsmFNop
+///////////////////////
+
+IR_AsmFNop::IR_AsmFNop(bool use_coloring) : IR_Asm(use_coloring) {}
```
water111, line 1065: It's not important, but you could leave out the `use_coloring` stuff for this IR.  It's the system used to assign variables to actual hardware registers. The assembly has a few modes where can either use variables with assembly and the compiler takes care of it (`use_coloring = true`), or you can manually specify registers (`use_coloring = false`).  This instruction doesn't touch variables/registers, so it doesn't need this option.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -125,6 +125,18 @@ Val* Compiler::compile_asm_ret(const goos::Object& form, const goos::Object& res
   return get_none();
 }
 
+Val* Compiler::compile_asm_fnop(const goos::Object& form, const goos::Object& rest, Env* env) {
```
water111, line 128: Usually I name the `compile_xxx` function after the name of the GOAL keyword, not the instruction. So this would be `compile_asm_nop_vf`

goalc/compiler/IR.cpp
```diff
@@ -1274,27 +1295,42 @@ void IR_RegSetAsm::do_codegen(emitter::ObjectGenerator* gen,
...
+    case Kind::MAX:
+      function = ".max.vf";
+      break;
+    case Kind::MIN:
+      function = ".min.vf";
+      break;
     default:
       assert(false);
   }
+  return fmt::format("{}{} {}, {}, {}, {}", function, get_color_suffix_string(), m_dst->print(),
```
water111, line 1332: For printing the mask, can we do `{:#b}`? It will print it as binary with the `0b` prefix, which will be easier to read than a decimal.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -1,3 +1,4 @@
+#include "..\Compiler.h"
```
water111, line 1: remove

goalc/compiler/compilation/Asm.cpp
```diff
@@ -311,16 +324,194 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
   return get_none();
 }
 
+// TODO - get rid of duplication
```
water111, line 327: I'm not too worried about this duplication. It's the price we pay for having that single table of keywords which I think is a net benefit even if you sometimes have stuff like this.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -311,16 +324,194 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
...
+}
+
+Val* Compiler::compile_asm_mulw_vf(const goos::Object& form, const goos::Object& rest, Env* env) {
+  return compile_asm_vf_math3(form, rest, IR_VFMath3Asm::Kind::MUL,
+                              emitter::Register::XMM_ELEMENT::W, env);
+}
+
+Val* Compiler::compile_asm_abs_vf(const goos::Object& form, const goos::Object& rest, Env* env) {
+  auto args = get_va(form, rest);
+  va_check(form, args, {{}, {}, {}}, {{"color", {false, goos::ObjectType::SYMBOL}}});
```
water111, line 461: You could consider making `mask` an optional argument and having it default to `0b1111`. This `va_check` would be changed to 
```
va_check(form, args, {{}, {}}, {{"color", {false, goos::ObjectType::SYMBOL}}, {"mask", {false, goos::ObjectType::INTEGER}}});
```

and then you could get the mask with
```
u8 mask = 0xff;
if (args.has_named("mask") {
  mask = args.named.at("mask").as_int()
}
```

Then you would specify the mask in the operation like `(.abs.vf vf1 vf2 :mask #b0101)` or just leave it out.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -311,16 +324,194 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
...
+        src->print());
+  }
+
+  int64_t mask;
+  if (!try_getting_constant_integer(args.unnamed.at(2), &mask, env)) {
+    throw_compiler_error(
+        form,
+        "The value {} is invalid for a destination mask, it could not be evaluated as a "
+        "constant integer.",
+        args.unnamed.at(3).print());
```
water111, line 487: `.at(2)`

goalc/compiler/compilation/Asm.cpp
```diff
@@ -311,16 +324,194 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
...
+  }
+
+  if (mask < 0 || mask > 15) {
+    throw_compiler_error(
+        form, "The value {} is out of range for a destination mask (0-15 inclusive).", mask);
+  }
+
+  // There is no single instruction ABS on AVX, so there are a number of ways to do it manually,
+  // this is one of them. For example, assume the original vec = <1, -2, -3, 4>
```
water111, line 496: This implementation is really cool!

goalc/compiler/compilation/Asm.cpp
```diff
@@ -370,9 +562,10 @@ Val* Compiler::compile_asm_blend_vf(const goos::Object& form, const goos::Object
 Val* Compiler::compile_asm_vf_math3(const goos::Object& form,
                                     const goos::Object& rest,
                                     IR_VFMath3Asm::Kind kind,
+                                    emitter::Register::XMM_ELEMENT broadcastElement,
                                     Env* env) {
   auto args = get_va(form, rest);
-  va_check(form, args, {{}, {}, {}}, {{"color", {false, goos::ObjectType::SYMBOL}}});
+  va_check(form, args, {{}, {}, {}, {}}, {{"color", {false, goos::ObjectType::SYMBOL}}});
```
water111, line 568: Same comment as the `abs`, we could make this a `:mask` keyword argument. Feel free to leave this for later.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -399,7 +592,37 @@ Val* Compiler::compile_asm_vf_math3(const goos::Object& form,
...
+  auto temp_reg = env->make_vfr(dest->type());
+
+  // If there is a broadcast register, splat that float across the entire src2 register before
+  // performing the operation For example vaddx.xyzw vf10, vf20, vf30
+  // vf10[x] = vf20[x] + vf30[x]
+  // vf10[y] = vf20[y] + vf30[x]
+  // vf10[z] = vf20[z] + vf30[x]
+  // vf10[w] = vf20[w] + vf30[x]
+  if (broadcastElement != emitter::Register::XMM_ELEMENT::NONE) {
+    env->emit_ir<IR_SplatVF>(color, src2, src2, broadcastElement);
```
water111, line 618: I think this will modify `src2` and you need to splat into a temporary here.

goalc/compiler/compilation/Atoms.cpp
```diff
@@ -26,8 +27,44 @@ static const std::unordered_map<
         {".lvf", &Compiler::compile_asm_lvf},
         {".svf", &Compiler::compile_asm_svf},
         {".xor.vf", &Compiler::compile_asm_xor_vf},
+        // TODO - Readability wise, it would probably be nice to have all the varients like
+        // '.sub.xyzw.vf' and so on, but it's not required at this point and needs to be done in a
```
water111, line 31: That's a good point. If we do end up doing this, let's make them macros.

goalc/compiler/compilation/Atoms.cpp
```diff
@@ -26,8 +27,44 @@ static const std::unordered_map<
...
+        // clever way to avoid lots of duplication
+        //
+        // TODO - difficult to cut down duplication as the map wants a function _reference_.
+        // ie. - I'd like to just call a function that returns a function, but the path forward
+        // isn't obvious to me yet.
```
water111, line 36: This is somewhat intentionally constraining to keep things simple, so it's always very clear what part of the compiler is compiling a form. A previous iteration of the compiler had a more complicated way to parse forms and it was kind of a nightmare.

goalc/emitter/IGen.h
```diff
@@ -2155,14 +2162,10 @@ class IGen {
...
-    assert(src1.is_xmm());
-    assert(src2.is_xmm());
-    Instruction instr(0x59);
-    instr.set_vex_modrm_and_rex(dst.hw_id(), src2.hw_id(), VEX3::LeadingBytes::P_0F, src1.hw_id());
-    return instr;
-  }
+  // TODO - rip, i reimplemented this below, which is better? im biased now
+  // personally, i think the shuffle instruction is a pain to work with directly, so id prefer
+  // working with it via helper functions (my splat one for example) so it doesn't really matter to
+  // me
```
water111, line 2168: Keep both? I agree the helper is better in 99% of cases and yours is well designed, but it may be nice to have the "raw" version as well.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -399,7 +592,37 @@ Val* Compiler::compile_asm_vf_math3(const goos::Object& form,
...
+  }
+
+  // Perform the arithmetic operation on the two vectors into a temporary register
+  env->emit_ir<IR_VFMath3Asm>(color, temp_reg, src1, src2, mask, kind);
```
water111, line 622: I think we can add a check to see if `mask == 0b1111` and if so optimize this to a single `IR_VFMath3Asm` without the blend at the end.

goalc/emitter/Register.h
```diff
@@ -96,6 +96,13 @@ class Register {
 
   std::string print() const;
 
+  /*
+          Our XMM Registers are 4 packed single-precision floating points
+    In the order (from left->right a.k.a most significant to least significant):
+    W | Z | Y | X
+          */
+  enum class XMM_ELEMENT { X, Y, Z, W, NONE };
```
water111, line 104: Can we rename this to `VF_Element`? The floating point numbers also go in `xmm`s so I try to use `vf` when referring to an `xmm` that has a vector.

test/goalc/source_templates/with_game/test-vector-abs.gc
```diff
@@ -0,0 +1,23 @@
+(defun test-vector-abs ()
+  (let ((vector-0 (new 'stack 'vector))
+        (vector-1 (new 'stack 'vector)))
+		; TODO - a lib function get generate a vector would be useful
```
water111, line 4: Initialization of stack variables in GOAL is kinda weird and I don't 100% understand when they are/aren't initialized so for now you are stuck with this.  In the near future you will be able to use this syntax for stack variables too:
```
(new 'stack 'vector :x 1.0 :y 2.0 :z 3.0 :w 4.0)
```
but it isn't done yet.

test/goalc/source_templates/with_game/test-vector-abs.gc
```diff
@@ -0,0 +1,23 @@
...
+		; TODO - it's really annoying getting a compiler error for passing `1` instead of `1.0`, quality of life improvement perhaps?
```
water111, line 5: It was somewhat intentional that floats/ints don't convert automatically in `set!` to help catch decompiler bugs. But it may be worth reconsidering, or at least having an option to disable this for code written by humans.

test/test_emitter_test.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "goalc/emitter/IGen.h"
+
+using namespace emitter;
+
+TEST(EmitterExperimentalTest, VF_NOP) {
+  CodeTester tester;
+	tester.init_code_buffer(1024);
+  tester.emit(IGen::nop_vf());
+
+	// ASK - easy way to get the opcode listing (that you know is correct) other than hand-crafting it?
```
water111, line 12: https://defuse.ca/online-x86-assembler.htm#disassembly and pick x64

I'd recommend just adding quick checks for each thing you add in `IGen`, getting the instruction slightly wrong is easy to do and sometimes it looks like it works.

water111 (commented):
Awesome progress! This is coming together much faster than I thought it would.

test/goalc/source_templates/with_game/test-vector-abs.gc
xTVaser: ah ok, that's much nicer.  I might still make the function to make writings tests easier _now_ but then it can be easily refactored to the better syntax when available.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -312,20 +320,208 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
...
+    throw_compiler_error(
+        form, "Invalid first source register for a vector float 3-arg math form. Got a {}.",
+        src->print());
+  }
+
+  u8 mask = 0b1111;
+  if (args.has_named("mask")) {
+    try {
+      mask = args.named.at("mask").as_int();
+    } catch (std::exception) {
```
water111, line 480: I think you don't want a try/catch here and you can just do `mask = args.named.at("mask").as_int();` directly. The `va_check` at the beginning makes sure that if there's a named argument named `mask` it's already an integer.  If you did want to check for it being an integer, it's probably best to check `args.named.at("mask").is_int()` instead.

In general I try not to use try/catch like this inside the compiler because the inner exception thrown could be some other compiler error that is being hidden by this `throw_compiler_error`.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -312,20 +320,208 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
...
+      throw_compiler_error(
+          form,
+          "The value {} is invalid for a destination mask, it could not be evaluated as a "
+          "constant integer.",
+          args.named.at("mask").print());
+    }
+
+    if (mask < 0 || mask > 15) {
```
water111, line 488: Can we remove the `mask < 0` check? mask in unsigned so this ends up being a compiler warning at least on linux.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -352,27 +548,37 @@ Val* Compiler::compile_asm_blend_vf(const goos::Object& form, const goos::Object
...
-                         "The value {} is invalid for a blend mask, it could not be evaluated as a "
-                         "constant integer.",
-                         args.unnamed.at(3).print());
-  }
+  u8 mask = 0b1111;
+  if (args.has_named("mask")) {
+    try {
+      mask = args.named.at("mask").as_int();
+    } catch (std::exception) {
+      throw_compiler_error(
```
water111, line 556: same comment as above, I don't think we need try/catch

goalc/compiler/compilation/Asm.cpp
```diff
@@ -352,27 +548,37 @@ Val* Compiler::compile_asm_blend_vf(const goos::Object& form, const goos::Object
...
+          form,
+          "The value {} is invalid for a destination mask, it could not be evaluated as a "
+          "constant integer.",
+          args.named.at("mask").print());
+    }
 
-  if (mask < 0 || mask > 15) {
-    throw_compiler_error(form, "The value {} is out of range for a blend mask.", mask);
+    if (mask < 0 || mask > 15) {
```
water111, line 563: remove `mask < 0`

test/goalc/test_with_game.cpp
```diff
@@ -345,6 +342,219 @@ TEST_F(WithGameTests, StaticBoxedArray) {
...
+
+  std::string toGOALFormat() {
+    std::string answer = fmt::format("({:.4f}, {:.4f}, {:.4f}, {:.4f})", x, y, z, w);
+    // {fmt} formats negative 0 as "-0.000", just going to flip any negative zeros to positives as I
+    // don't think is an OpenGOAL issue
+    return std::regex_replace(answer, std::regex("-0.0000"), "0.0000");
+  }
+};
+
+struct VectorFloatTestCase {
```
water111, line 383: These tests are awesome!

goalc/compiler/compilation/Asm.cpp
```diff
@@ -399,7 +605,56 @@ Val* Compiler::compile_asm_vf_math3(const goos::Object& form,
...
+    }
+  }
+
+  // If there is a broadcast register, splat that float across the entire src2 register before
+  // performing the operation For example vaddx.xyzw vf10, vf20, vf30
+  // vf10[x] = vf20[x] + vf30[x]
+  // vf10[y] = vf20[y] + vf30[x]
+  // vf10[z] = vf20[z] + vf30[x]
+  // vf10[w] = vf20[w] + vf30[x]
+  // TODO - surely i can condense this...(only using a temp register when i have to)
```
water111, line 632: I don't think there's much more you can do here to make it simpler.

If it's possible for `temp_reg` to be the same register as `src2`, the compiler will do this automatically during the register allocation phase.

goalc/compiler/IR.cpp
```diff
@@ -1278,23 +1299,36 @@ IR_VFMath3Asm::IR_VFMath3Asm(bool use_color,
                              const RegVal* dst,
                              const RegVal* src1,
                              const RegVal* src2,
+                             u8 mask,
```
water111, line 1302: I think we can remove `mask` from this IR entirely?  This IR always operates on all four components and doesn't do anything with the mask, other than print it.

goalc/emitter/IGen.h
```diff
@@ -2187,11 +2180,68 @@ class IGen {
     //    return instr;
   }
 
+  /*
+    Generic Swizzle (re-arrangment of packed FPs) operation, the control bytes are quite involved.
```
water111, line 2184: great description!

goalc/compiler/compilation/Asm.cpp
```diff
@@ -399,7 +605,56 @@ Val* Compiler::compile_asm_vf_math3(const goos::Object& form,
         src2->print());
   }
 
-  env->emit_ir<IR_VFMath3Asm>(color, dest, src1, src2, kind);
+  u8 mask = 0b1111;
+  if (args.has_named("mask")) {
+    try {
```
water111, line 610: no need for try/catch, integer is checked in `va_check`

goalc/compiler/compilation/Asm.cpp
```diff
@@ -399,7 +605,56 @@ Val* Compiler::compile_asm_vf_math3(const goos::Object& form,
...
+      mask = args.named.at("mask").as_int();
+    } catch (std::exception) {
+      throw_compiler_error(
+          form,
+          "The value {} is invalid for a destination mask, it could not be evaluated as a "
+          "constant integer.",
+          args.named.at("mask").print());
+    }
+
+    if (mask < 0 || mask > 15) {
```
water111, line 620: `mask < 0` is a compiler warning, can remove

test/goalc/test_with_game.cpp
```diff
@@ -345,6 +342,219 @@ TEST_F(WithGameTests, StaticBoxedArray) {
...
+  std::string test = fmt::format("{:.4f}", -0.0);
+
+  std::vector<VectorFloatTestCase> cases = {};
+  for (int i = 0; i <= 15; i++) {
+    VectorFloatTestCase testCase = VectorFloatTestCase();
+    testCase.destinationMask = i;
+    cases.push_back(testCase);
+    // Re-add each case with each broadcast varient
+    for (int j = 0; j < 4; j++) {
+      VectorFloatTestCase testCase = VectorFloatTestCase();
```
water111, line 450: compiler warning on `testCase` here having the same name as `testCase` on line 445

test/goalc/test_with_game.cpp
```diff
@@ -345,6 +342,219 @@ TEST_F(WithGameTests, StaticBoxedArray) {
...
+  env.write(templateFile, data, outFile);
+  runner.run_test(testCategory, outFile,
+                  {fmt::format("{}\n0\n", testCase.getExpectedResult().toGOALFormat())});
+}
+
+// TODO - This test runs more often than the rest, should probably be split into it's own fixture
+// (broadcasting ignored!)
+TEST_P(VectorFloatParameterizedTestFixtureWithRunner, VF_ABS_DEST) {
+  VectorFloatTestCase testCase = GetParam();
+  testCase.operation = [](float x, float y) { return fabs(x); };
```
water111, line 543: compiler warning on unused `y`.

water111 (changes requested):
This is great work and I expect we'll start using these very soon - the 4th file in the game is `vector-h`.

I had a few tiny comments, mostly just to get rid of compiler warnings on linux. After those are fixed this is ready to merge!

PR #232, opened 2021-02-05
third-party/licensing: Switch miniLZO usage to LZOkay

xTVaser:
Commits will cleanup once VU PR is merged.

This switches our usage of miniLZO (which is GPLv2) to an MIT licenses implementation so we are free to license however we please going forward.

I decompressed COMMON.CGO and the output was identical:
```powershell
PS C:\Users\xtvas\Repositories\jak-project\decompiler_out> .\compare.ps1

PS C:\Users\xtvas\Repositories\jak-project\decompiler_out> .\compare.ps1

InputObject                 SideIndicator
-----------                 -------------
intentionally different.rtf <=
```

image

Also the updated Jak2 Script is super rough, it's largely copied from the existing Jak1 config file, just so i could get a decompression going.


third-party/lzokay/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+cmake_minimum_required(VERSION 3.10)
+include(CMakePackageConfigHelpers)
+
+project(lzokay VERSION 0.1 LANGUAGES CXX)
+add_library(lzokay lzokay.hpp lzokay.cpp)
```
water111, line 5: I think we should make this `add_library(lzokay SHARED lzokay.hpp lazkay.cpp)` and it will fix the weird build issue.

PR #246, opened 2021-02-09
[timer-h decomp] timer-h decomp

ManDude:
Only issue: the compiler spits a warning on the `get-last-frame-time-stamp` method:
`[TypeSystem] The method get-last-frame-time-stamp of type profile-bar was originally defined as (function profile-bar uint32), but has been redefined as (function profile-bar uint)`

Which I am not sure what it's trying to tell me, maybe the `defmethod` for it is redefining the return value in some unexpected way.


water111:
The issue is the line:
```
(get-last-frame-time-stamp (profile-bar) uint32 9)
```
should actually be `uint`. Once a `uint32` is loaded from memory it becomes a `uint`.

OpenGOAL and GOAL use different types for integers in memory and integers in registers.  Integers in memory can be `uint8`, `uint16`, ...  Integers in registers are always `uint` or `int` and are 64-bit.  I think they did this in GOAL so they didn't have to make every single operation work for every single integer size. For example, MIPS has no `add` instruction for 8 bit values, so you'd have to do a 32-bit add and then add some shifts to clear the upper bits.

Integers in types are used for:
- Fields in types
- Array elements (both `array`, `boxed-array`, and `inline-array`)

Integers in registers are used for:
- Variables defined with `let` or `local-vars`
- Function arguments
- Function return values

PR #258, opened 2021-02-13
compiler: Support the majority of the remaining VU VF instructions

xTVaser:
This adds support for the majority of the remaining VU instructions.  The remaining being:
- R Register
  - `vrget`
  - `vrnext`
  - `vrxorw`
- Flag support needed
  - `VCLIP`
- Uses VU Memory
  - `vsqi.dest`
  - `vlqi.dest`
- Integer related
  - `vftoi0` | `vftoi12` | `vftoi4` | `viaddi` | `viand` | `vior` | `vitof0` | `vitof12` | `vitof15` | `vmtir`
    - I think it was discussed that these could be handled not through asm instructions, but through some C?

I think I identified an issue with the VDIV/VSQRT instruction encoding, the gist being, `fsf` and `ftf` don't seem to be properly output.

In this PR, the following are supported one way or another (some will require the use of some clever decompiling), for instructions that use the ACC/Q, this will be supported by an additional vector argument, assuming the game does not rely on side-effects of these registers being mutated outside of the scope of operation, this is completely sufficient, but probably something we'd want the decompiler to flag us for.
  - `V[ADD|SUB|MUL]a[x-z].dest`
  - `V[ADD|SUB|MUL][a|q].dest`
  - `VM[ADD|SUB|MUL]a[x-z].dest`
  - `VM[ADD|SUB|MUL][a|q].dest`
  - `vopmula.dest` & `vopmsub.dest` via a single instruction
  - `vdiv`
  - `vsqrt`
  - `vwaitq`

Relates to - https://github.com/water111/jak-project/issues/201

### DIFF
--- a/goalc/compiler/IR.h
+++ b/goalc/compiler/IR.h
@@ -473,6 +473,16 @@ class IR_AsmFNop : public IR_Asm {
                   emitter::IR_Record irec) override;
 };
 
+class IR_AsmFWait : public IR_Asm {
+ public:
+  IR_AsmFWait();
+  std::string print() override;
+  RegAllocInstr to_rai() override;
+  void do_codegen(emitter::ObjectGenerator* gen,
+                  const AllocationResult& allocs,
+                  emitter::IR_Record irec) override;
+};
+
 class IR_GetSymbolValueAsm : public IR_Asm {
  public:
   IR_GetSymbolValueAsm(bool use_coloring, const RegVal* dest, std::string sym_name, bool sext);
@@ -517,7 +527,7 @@ class IR_RegSetAsm : public IR_Asm {
 
 class IR_VFMath3Asm : public IR_Asm {
  public:
-  enum class Kind { XOR, SUB, ADD, MUL, MAX, MIN };
+  enum class Kind { XOR, SUB, ADD, MUL, MAX, MIN, DIV };
   IR_VFMath3Asm(bool use_color,
                 const RegVal* dst,
                 const RegVal* src1,
@@ -533,6 +543,17 @@ class IR_VFMath3Asm : public IR_Asm {
   const RegVal* m_dst = nullptr;
   const RegVal* m_src1 = nullptr;
   const RegVal* m_src2 = nullptr;
+  Kind m_kind;
+};
+
+class IR_VFMath4Asm {
+ public:
+  enum class Kind { MUL_ADD, MUL_SUB };
+  IR_VFMath4Asm(Kind kind);
+  IR_VFMath3Asm::Kind first_op();
+  IR_VFMath3Asm::Kind second_op();
+
+ protected:
   Kind m_kind;
 };
 
@@ -556,7 +577,7 @@ class IR_SplatVF : public IR_Asm {
  public:
   IR_SplatVF(bool use_color,
              const RegVal* dst,
-             const RegVal* src1,
+             const RegVal* src,
              const emitter::Register::VF_ELEMENT element);
   std::string print() override;
   RegAllocInstr to_rai() override;
@@ -568,5 +589,34 @@ class IR_SplatVF : public IR_Asm {
   const RegVal* m_dst = nullptr;
   const RegVal* m_src = nullptr;
   const emitter::Register::VF_ELEMENT m_element = emitter::Register::VF_ELEMENT::NONE;
+};
+
+class IR_SwizzleVF : public IR_Asm {
+ public:
+  IR_SwizzleVF(bool use_color, const RegVal* dst, const RegVal* src, const u8 m_controlBytes);
+  std::string print() override;
+  RegAllocInstr to_rai() override;
+  void do_codegen(emitter::ObjectGenerator* gen,
+                  const AllocationResult& allocs,
+                  emitter::IR_Record irec) override;
+
+ protected:
+  const RegVal* m_dst = nullptr;
+  const RegVal* m_src = nullptr;
+  const u8 m_controlBytes = 0;
+};
+
+class IR_SqrtVF : public IR_Asm {
+ public:
+  IR_SqrtVF(bool use_color, const RegVal* dst, const RegVal* src);
+  std::string print() override;
+  RegAllocInstr to_rai() override;
+  void do_codegen(emitter::ObjectGenerator* gen,
+                  const AllocationResult& allocs,
+                  emitter::IR_Record irec) override;
+
+ protected:
+  const RegVal* m_dst = nullptr;
+  const RegVal* m_src = nullptr;
 };
 #endif  // JAK_IR_H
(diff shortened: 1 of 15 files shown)


goalc/compiler/IR.h
```diff
@@ -536,6 +546,17 @@ class IR_VFMath3Asm : public IR_Asm {
   Kind m_kind;
 };
 
+class IR_VFMath4Asm {
```
water111, line 549: This looks unused?

goalc/compiler/compilation/Asm.cpp
```diff
@@ -319,6 +327,113 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
...
 }
 
+void Compiler::check_vector_float_regs(const goos::Object& form,
+                                       Env* env,
+                                       std::vector<std::pair<std::string, RegVal*>> args) {
+  for (std::pair<std::string, RegVal*> arg : args) {
+    if (!arg.second->settable() || arg.second->ireg().reg_class != RegClass::VECTOR_FLOAT) {
+      throw_compiler_error(
+          form,
+          fmt::format("Invalid {} register for a vector float operation form. Got a {{}}.",
```
water111, line 337: The `throw_compiler_error` thing has formatting build in, I think this could be:
```cpp
throw_compiler_error(form, "Invalid {} register for a vector float operation form. Got a {}.", arg.first, arg.second->print());
```

goalc/compiler/compilation/Asm.cpp
```diff
@@ -429,50 +544,203 @@ Val* Compiler::compile_asm_mul_vf(const goos::Object& form, const goos::Object&
...
-  if (!dest->settable() || dest->ireg().reg_class != RegClass::VECTOR_FLOAT) {
-    throw_compiler_error(
-        form, "Invalid destination register for a vector float 3-arg math form. Got a {}.",
-        dest->print());
+  auto src1 = compile_error_guard(args.unnamed.at(1), env)->to_reg(env);
+  auto src2 = compile_error_guard(args.unnamed.at(2), env)->to_reg(env);
+  // This third register is intended for the ACC/Q/ETC, and is used to temporarily store the value
+  // that eventually goes into the destination
+  //
+  // For example VMADD:
```
water111, line 588: With the `+=`, this would be `VMADDA`, not `VMADD`, right?

I'm guessing we can use this form for both versions, and for `VMADDA`, we can have the decompiler add a `(mov acc vf_dest)`

goalc/compiler/compilation/Asm.cpp
```diff
@@ -429,50 +544,203 @@ Val* Compiler::compile_asm_mul_vf(const goos::Object& form, const goos::Object&
...
   }
 
-  auto src = compile_error_guard(args.unnamed.at(1), env)->to_reg(env);
-  if (src->ireg().reg_class != RegClass::VECTOR_FLOAT) {
-    throw_compiler_error(
-        form, "Invalid first source register for a vector float 3-arg math form. Got a {}.",
-        src->print());
+  // First we clear a temporary register
+  auto temp_reg = env->make_vfr(dest->type());
+  env->emit_ir<IR_VFMath3Asm>(color, temp_reg, temp_reg, temp_reg, IR_VFMath3Asm::Kind::XOR);
```
water111, line 609: Do we need this clear? I don't think we need the temp register to be zeroed first.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -507,125 +775,178 @@ Val* Compiler::compile_asm_abs_vf(const goos::Object& form, const goos::Object&
...
-  // If there is a broadcast register, splat that float across the entire src2 register before
-  // performing the operation For example vaddx.xyzw vf10, vf20, vf30
-  // vf10[x] = vf20[x] + vf30[x]
-  // vf10[y] = vf20[y] + vf30[x]
-  // vf10[z] = vf20[z] + vf30[x]
-  // vf10[w] = vf20[w] + vf30[x]
-  if (broadcastElement != emitter::Register::VF_ELEMENT::NONE) {
-    auto temp_reg = env->make_vfr(dest->type());
-    env->emit_ir<IR_SplatVF>(color, temp_reg, src2, broadcastElement);
+  // Perform the operation, you can't seem to VSQRTPS with the same register, which is why we needed
```
water111, line 886: Hmmm. I think the encoding of `vsqrtps` may be slightly wrong and causing this issue.  Can we try this without the temp register with the encoding fixed?

goalc/emitter/IGen.h
```diff
@@ -2297,15 +2312,20 @@ class IGen {
...
+    Instruction instr(0x5E);  // VDIVPS
+    instr.set_vex_modrm_and_rex(dst.hw_id(), src2.hw_id(), VEX3::LeadingBytes::P_0F, src1.hw_id());
+    return instr;
+  }
+
+  static Instruction sqrt_vf(Register dst, Register src) {
+    assert(dst.is_xmm());
+    assert(src.is_xmm());
+    Instruction instr(0x51);  // VSQRTPS
+    instr.set_vex_modrm_and_rex(dst.hw_id(), src.hw_id(), VEX3::LeadingBytes::P_0F, src.hw_id());
```
water111, line 2328: I think this is slightly wrong. The final `src.hw_id()` should be `0b11111` always, see https://www.felixcloutier.com/x86/sqrtps

"Note: VEX.vvvv and EVEX.vvvv are reserved and must be 1111b otherwise instructions will #UD."

test/test_emitter_avx.cpp
```diff
@@ -281,6 +288,33 @@ TEST(EmitterAVX, BlendVF) {
...
+
+TEST(EmitterAVX, SqrtVF) {
+  CodeTester tester;
+  tester.init_code_buffer(1024);
+  tester.emit(IGen::sqrt_vf(XMM0 + 3, XMM0 + 4));
+  tester.emit(IGen::sqrt_vf(XMM0 + 3, XMM0 + 14));
+  tester.emit(IGen::sqrt_vf(XMM0 + 13, XMM0 + 4));
+  tester.emit(IGen::sqrt_vf(XMM0 + 13, XMM0 + 14));
+
+  EXPECT_EQ(tester.dump_to_hex_string(true), "C5D851DCC4C10851DEC55851ECC4410851EE");
```
water111, line 315: Should be `C5F8...`. I don't know why, but this shows up as valid in https://defuse.ca/online-x86-assembler.htm#disassembly but I don't think it really is.

water111 (changes requested):
Nice work! I had a few small comments but it is very close.

goalc/compiler/compilation/Asm.cpp
xTVaser: Yeah for my example's purposes you're right, its a `VMADDA`, must have got confused as like you said, we can use it for both.

goalc/compiler/compilation/Asm.cpp
xTVaser: Hmm, good catch (with below info) however, I'm not able to get things to work.  I get Illegal Instructions, I'll commit what I have

water111 (approved):

PR #275, opened 2021-02-21
[Decompile] DMA and similar

water111:
Decompile:
- `gsound-h`
- `timer-h`
- `timer`
- `vif-h`
- `dma-h`
- `video-h`
- `vu1-user-h`
- `dma`
- `dma-buffer`
- `dma-bucket`


water111:
codacy is complaining because my docstring doesn't end with a period....

PR #287, opened 2021-02-26
[Decomp] Minor fixes plus some math camera decomp

water111:


water111:
will close https://github.com/water111/jak-project/issues/99

PR #289, opened 2021-02-27
decomp - vector.gc - Decompile vector.gc

xTVaser:
Just putting up the PR early, I've noticed some new instructions that I need to support on the compiler side / some ones i missed in the re-write stage, and there are 2 functions that fail to decompile before my new stage.


decompiler/config/jak1_ntsc_black_label.jsonc
```diff
@@ -3,14 +3,9 @@
 {
```
xTVaser, line 3: dont let me merge without reverting this!

decompiler/IR2/AtomicOp.cpp
```diff
@@ -471,6 +472,38 @@ goos::Object AsmOp::to_form(const std::vector<DecompilerLabel>& labels, const En
...
+  std::vector<goos::Object> forms;
+
+  std::vector<std::optional<RegisterAccess>> src;
+  for (int i = 0; i < m_instr.n_src; i++) {
+    auto v = m_src[i];
+    src.push_back(v);
+  }
+
+  OpenGOALAsm goalOp = OpenGOALAsm(m_instr, m_dst, src);
+  OpenGOALAsm::Function func = goalOp.func;
```
water111: remove line?

decompiler/IR2/Form.cpp
```diff
@@ -740,6 +794,69 @@ void EmptyElement::apply_form(const std::function<void(Form*)>&) {}
...
+
+RLetElement::RLetElement(Form* _body, RegSet _regs) : body(_body) {
+  for (auto& reg : _regs) {
+    sorted_regs.push_back(reg);
+  }
+  std::sort(sorted_regs.begin(), sorted_regs.end(), cmp);
+}
+
+void RLetElement::apply(const std::function<void(FormElement*)>& f) {
+  // note - this is done in program order, rather than print order. Not sure if this makes sense.
```
water111, line 816: remove this comment

decompiler/IR2/Form.cpp
```diff
@@ -740,6 +794,69 @@ void EmptyElement::apply_form(const std::function<void(Form*)>&) {}
...
+
+  std::vector<goos::Object> rletForm;
+  rletForm.push_back(pretty_print::to_symbol("rlet"));
+  rletForm.push_back(pretty_print::build_list(regs));
+
+  // NOTE - initialize any relevant registers in the body first
+  for (auto& reg : sorted_regs) {
+    if (reg.get_kind() == Reg::RegisterKind::VF && reg.to_string() == "vf0") {
+      rletForm.push_back(
+          pretty_print::to_symbol("(.lvf vf0 (new 'static 'vector :x 0.0 :y 0.0 :z 0.0 :w 1.0))"));
```
water111, line 840: Don't worry about it for this PR, but it might be a good idea to move this to a macro like `initialize-constant-vector!` or something.  There could be some clever way to do this initialization that's faster that a normal static load.

decompiler/IR2/FormExpressionAnalysis.cpp
```diff
@@ -1564,6 +1564,26 @@ void DerefElement::inline_nested() {
   }
 }
 
+///////////////////
+// RLetElement
+///////////////////
+
+void RLetElement::push_to_stack(const Env& env, FormPool& pool, FormStack& stack) {
```
water111, line 1571: I think we can leave out this function. These should only get used during expression propagation, which happens before your analysis pass. If you remove this, it will use the default for a `FormElement`, which is probably what you want.

decompiler/IR2/FormExpressionAnalysis.cpp
```diff
@@ -2158,6 +2178,10 @@ void AsmOpElement::push_to_stack(const Env&, FormPool&, FormStack& stack) {
   stack.push_form_element(this, true);
 }
 
+void OpenGoalAsmOpElement::push_to_stack(const Env&, FormPool&, FormStack& stack) {
```
water111, line 2180: Same as `RLetElement`, I think you can remove this.

decompiler/analysis/inline_asm_rewrite.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+      RLetElement* rlet = pool.alloc_element<RLetElement>(body, vf_regs);
+      top_level_form->clear();
+      top_level_form->push_back(rlet);
+    }
+  } catch (std::exception& e) {
+    std::string warning = fmt::format("ASM instruction re-writing failed in {}: {}",
+                                      f.guessed_name.to_string(), e.what());
+    lg::warn(warning);
+    // TODO - changed to what? f.warnings.append(";; " + warning);
```
water111, line 86: To put a warning in the source at the top of the function:
```
f.warnings.general_warning("some warning {}", 1);
```
It uses `fmt` formatting.

decompiler/config/jak1_ntsc_black_label/label_types.jsonc
```diff
@@ -1,282 +1,281 @@
...
-	],
-
-	"collide-edge-grab-h":[
-	    ["L1", "collide-edge-work", true]]
-
-
-
-}
\ No newline at end of file
+  "math": [
```
water111, line 2: Thanks for formatting this!

goal_src/engine/math/vector-h.gc
```diff
@@ -493,8 +493,11 @@
     (set! (-> ,v x) ,xv)
     (set! (-> ,v y) ,yv)
     (set! (-> ,v z) ,zv)
-    (set! (-> ,v w) ,wv))
-  )
+    (set! (-> ,v w) ,wv)))
+
+(defmacro print-vector! (v)
```
water111, line 498: I think they used `!` only for functions which mutate the input.
Also, in newer versions of the compiler, you can just do `(print x)` and it will actually call the print method.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -274,15 +276,15 @@ Val* Compiler::compile_asm_lvf(const goos::Object& form, const goos::Object& res
...
   MemLoadInfo info;
   info.sign_extend = false;
   info.size = 16;
   info.reg = RegClass::VECTOR_FLOAT;
-  if (as_co) {
+  if (args.has_named("offset")) {
+    int offset = args.named.at("offset").as_int();
     // can do a clever offset here
-    env->emit_ir<IR_LoadConstOffset>(dest, as_co->offset, as_co->base->to_gpr(env), info, color);
+    env->emit_ir<IR_LoadConstOffset>(dest, 0, src->to_gpr(env), info, color);
```
water111, line 287: I think this isn't quite right. The `MemoryOffsetConstantVal` represents a memory location plus some constant offset.  We want to propagate constant memory offsets at compile to avoid using temporary registers and addition instructions when possible.  

There should be 6 cases:
- `as_co` and no `:offset` - use the old `as_co->offset` and `as_co->base`
- `as_co` and `:offset` - use `as_co->offset + offset` and `as_co->base`
- No `as_co` and no `:offset` - use `0` and `src->to_gpr(env)`
- No `as_co` and `:offset` - use `offset` and `src->to_gpr(env)`
- `as_sv` and no `:offset` - do a `IR_StaticVarLoad`
- `as_sv` and `:offset` - throw a compiler error.  Loading directly from a `(new 'static ...)` isn't something that the original GOAL supported - I only added it to make it faster to load constant vectors/floats.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -317,9 +321,11 @@ Val* Compiler::compile_asm_svf(const goos::Object& form, const goos::Object& res
   info.sign_extend = false;
   info.size = 16;
   info.reg = RegClass::VECTOR_FLOAT;
-  if (as_co) {
+  // TODO - i don't think this is correct (the gpr allocation)
```
water111, line 324: This should be similar to the `lvf` cases, but leave out the last two.

water111 (changes requested):
The rewriting pass looks good!  I had a few comments on the `:offset` argument added to lvf/svf, but other than that it is very close.

It seems like I'll need to do some more work on the decompiler before we're really ready to decompile vector.gc, so we can worry about that later.

goal_src/engine/math/vector-h.gc
xTVaser: Do vectors have a print method? I forgot I even made this change, was back when i was doing some initial tests, might not be relevant if there's a simpler way to print a vector as you mention

doc/goal_doc.md
```diff
@@ -1317,11 +1317,11 @@ Inserts a `FWAIT` assembly instruction, x86 does not require as much synchroniza
...
+(.lvf dst-reg src-loc [:color #t|#f] [:offset <int>])
 ```
-Load a vector float register from `src-loc`. The `dst-reg` must be a vector float register. The `src-loc` can be a gpr containing a GOAL pointer or expression which gives a GOAL pointer. There is no type checking on the `src-loc` so be careful. The load uses `vmovaps`, so the source must be 16-byte aligned. 
+Load a vector float register from `src-loc`. The `dst-reg` must be a vector float register. The `src-loc` can be a gpr containing a GOAL pointer or expression which gives a GOAL pointer. There is no type checking on the `src-loc` so be careful. The load uses `vmovaps`, so the source must be 16-byte aligned.
 
-If the source is in the form `base-reg + constant-offset`, like from a `(&-> my-object my-inline-vector-field)`, the constant offset will be folded into the load instruction like `vmovaps xmm1, [r15 + rax + 12]`.
+If the source is in the form `base-reg + constant-offset`, like from a `(&-> my-object my-inline-vector-field)`, specify the offset explicitly with `:offset`, this will be folded into the load instruction like `vmovaps xmm1, [r15 + rax + 12]`.  The default `:offset` is assumed to be `0`.
```
xTVaser, line 1324: reminder: change this to reflect whatever updates i make it lvf/svf

goal_src/engine/math/vector-h.gc
water111: Yep: https://github.com/water111/jak-project/blob/master/goal_src/engine/math/vector-h.gc#L319
but it used to be annoying to call due to `print` being the name of both a method (which does the right thing) and a function (which doesn't).  Now the method should win over the function.

PR #298, opened 2021-03-02
repl: Add a (repl-help) command for forgetful people like me

xTVaser:
As discussed, a brief help command.  Originally I wanted to do it in GOAL code, but doing it in C++ allows me to print without requiring any setup / better coloring and formatting.  I'll try adding the persistent history next.

image


water111:
Nice, this is great!

PR #301, opened 2021-03-03
repl: Support cross-session history

xTVaser:
Linenoise had all the tools already to support this, the problem (and i could very well have solved it in a less than ideal way) is that the history is stored in a `static` vector.  Being as we have to interact with the history in different points of the app (ie, load the history on startup / save it on exit / etc) these are all in different files, and when doing so the history would be empty.

Wrapping everything to have a single linenoise include, "solves" this problem, and history now works as expected.

It will save a `.opengoal.repl.history` file in your (linux) `$HOME` or (windows) `$USERPROFILE` directory.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -8,6 +8,7 @@ add_library(common
         goos/PrettyPrinter.cpp
         goos/Reader.cpp
         goos/TextDB.cpp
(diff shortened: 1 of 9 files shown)


water111:
This is awesome, it makes such a difference to have the history saved.

The static history is really annoying, but I think your wrapper is a good solution.

PR #318, opened 2021-03-09
Implement runtime display (test)

ManDude:
Implements a display window using GLFW and OpenGL (using glad).

There is currently a bug where manually closing the display window will cause some issues with the listener check. I don't really know how to fix them.


ManDude:
The system works perfectly now. The runtime is ready to have actual graphics rendered in it, and handling user input.

PR #336, opened 2021-03-25
Support 128-bit variables

water111:
Going down the list fixing issues and checking 128-bit support.  I'm not going to worry about bitfield access for now, but eventually we will want this.

Todo:
- Try out some stuff in `vector
- Make the types of the failed stuff in `vec4s` correct (everything other than accessing bitfields should work)


water111:
Will close https://github.com/water111/jak-project/issues/335 and https://github.com/water111/jak-project/issues/334 as well

PR #341, opened 2021-03-28
[Decompiler] Make matrix decompile

water111:
Will fix https://github.com/water111/jak-project/issues/339

See also https://github.com/water111/jak-project/issues/340


water111:
The `matrix-4x4-determinant` function is giving the wrong answer on this matrix:
```
      (src (new 'static 'matrix :data (new 'static 'array float 16
                                           1. 3. 3. 4.
                                           8. 6. 7. 8.
                                           9. 10. 11. 12.
                                           13. 14. 15. 17.))))
```

(and seems to give the wrong answer on all matrices).
It compute up with -463. 

It seemed like this would be a good function to expose an error in register allocation in the compiler, so I checked the function in common lisp and it also got -463, so I believe the compiler is working properly and the issue is either in the game code or in the decompiler.

PR #344, opened 2021-03-29
[Decomp] Math library: transform and quaternion

water111:
Also fixes:
- `outer.product` when src = dst (https://github.com/water111/jak-project/issues/346)
- allows macros to be used in static field definitions
- `sqrt` to `sqrtf` inconsistently named between compiler/decompiler
- In rare cases, decompiler misses casts
- Moving from `vf` gives you a float


water111:
Pulling out sets from an `if` drops the casts, causing StringToInt to fail the offline test. Also some offline tests still need cleaning up and I want to double check that nothing broke when I'm a little less tired.

PR #350, opened 2021-04-03
[Decompiler] bitfield support and clean up of DMA

water111:
For now, it supports accessing bitfields, both signed/unsigned, on 64-bit types.  Interestingly, GOAL uses the `srl` instruction (not `dsrl`) to save an instruction when a bitfield ends at bit 32 and appears to rely on this truncating bits 32-63.  The manual says this is undefined behavior of the CPU, but it seems to work for them?

All of the stuff up to `dma.gc` (minus `trigonometry` and `geometry`) worked and are added to the test. There are some bitfield access in the inspect methods that make a nice test.

The next trick to handle is "check for bitfield being zero".  The asm is:
```asm
    lwu a1, 0(a0)
    andi a1, a1, 256
    bne a1, r0, L47
    sll r0, r0, 0
```
at least when the bit mask can fit in the imm16 of an `andi`.

It's current decompiled as:
```lisp
(nonzero? (logand (the-as int (-> arg0 chcr)) 256)
```
which is reasonable and correct.  The current plan is to hook `logand` for operating on a bitfield, have it transform into a `partial-bitfield-access` then have `nonzero?` or similar check for the `partial-bitfield-access`.


water111:
Setting a bitfield value, probably from this code
```lisp
(set! (-> (the-as vif-bank #x10003800) err me0) 1)
```
Generates this asm
```asm
    lui v1, 4096
    ori v1, v1, 14336
    lwu v1, 32(v1)            ;; get err
    addiu a0, r0, -3          ;; inverted mask for me0
    and v1, v1, a0            ;; get rid of me0's value
    ori v1, v1, 2             ;; set new value (from constant)
    lui a0, 4096              ;; and store it back (note the expression is evaluated again!)
    ori a0, a0, 14336
    sw v1, 32(a0)             
```

Interestingly, the compiler totally re-evaluates the expression to get the address containing the location of the bitfield. Which makes a lot of sense based on how we believe the GOAL compiler was implemented.

So it's more like:
```lisp
(let ((temp (-> bank err me0)))
  (set! (-> temp me0) 1)
  (set! (-> bank err) temp)
  )
```

I'm going to aim for getting the above output on this first pass because I think eliminating that temp is probably better to do somewhere else.

water111:
It is now working for everything in `dma.gc`. I want to see how it does on some more files before merging this, but so far it seems promising.

water111:
The case above works properly now:
```lisp
(set! (-> (the-as vif-bank #x10003800) err me0) 1)
```

PR #359, opened 2021-04-14
prepare for decompiling display

ManDude:
Needs some work on the decompiler (mostly bitfield stuff) before a more proper decomp can be done


ManDude:
Probably done now. I'll wait until bitfield types support is complete to continue messing around with getting a basic renderer set up.

PR #372, opened 2021-04-19
Allow enum names to serve as types (using the enum's type)

ManDude:


common/type_system/TypeSystem.cpp
```diff
@@ -70,6 +70,46 @@ Type* TypeSystem::add_type(const std::string& name, std::unique_ptr<Type> type)
...
+    if (kv->second != type) {
+      // exists, and we are trying to change it!
+      fmt::print("[TypeSystem] Enum {} was originally\n{}\nand is redefined as\n{}\n", name,
+                 kv->second, type);
+
+      throw std::runtime_error("Enum type was redefined.");
+    }
+  } else {
+    // Should forward declaring enums be forbidden? would it even necessarily break anything? I'll
+    // keep it allowed for now, though it is a weird thing to do...
```
water111, line 93: Let's make this an error. I don't think there's a reason to forward declare enums, and we want to prevent the case where somebody forward declares a type as a structure, and then uses that same name for an enum.

common/type_system/TypeSystem.cpp
```diff
@@ -70,6 +70,46 @@ Type* TypeSystem::add_type(const std::string& name, std::unique_ptr<Type> type)
...
+    // if (m_forward_declared_types.find(name) != m_forward_declared_types.end()) {
+    //  fmt::print(
+    //      "[TypeSystem] Enum type {} was forward-declared, enums cannot be forward-declared\n");
+    //}
+
+    // newly defined!
+
+    if (m_types.find(t->get_parent()) == m_types.end()) {
```
water111, line 101: We can remove this check.  It's only needed when initially setting up the built-in types and all types should have valid parents by now.

common/type_system/defenum.cpp
```diff
@@ -0,0 +1,154 @@
...
+    case 8:
+      return true;
+    default:
+      assert(false);
+      return false;
+  }
+}
+
+template <typename T>
+void for_each_in_list(const goos::Object& list, T f) {
```
water111, line 62: I think this exists in `common/goos/ParseHelpers` already

water111 (changes requested):
Looks good! I had a few small comments.

Does this work for function arguments? I suspect it will give you a typechecking error about using an `int` where an `int32` is expected but I'm not sure.

common/type_system/defenum.cpp
ManDude: I can only find it in `goalc/compiler/compilation/Static.cpp`

water111 (approved):

ManDude:
It does give that typechecking error, I'll fix it meanwhile.

ManDude:
Actually I probably won't, since it seems more of an issue with `defun` than with the enums.

PR #374, opened 2021-04-21
[Compiler/Decompiler] Better support for Bitfield and Enum types

water111:
This makes enums real types, like bitfields.

So far:
- Fixed bugs where boxed arrays of enums/bitfields were broken
- Made enum types real types in the compiler. They can be used for fields, pointers, function args/return values, locals, etc, just like bitfields. Just like bitfields, they require casting to and from int.  Most of the reason for doing this was to make the type systems of the compiler/decompiler match more closely.
- Support for decompiling bitfield enum constants that appear in code

To Do:
- Support bitfield enum constant in static data too (should be easy)
- Fix bad casting with `(logior <bitfield-enum> <constant>)`.  The type pass correctly determines that the result of this is also a bitfield enum, but somehow things go wrong later on.
- Try to find some examples of non-bitfield enums and add those too.

Looks good
```asm
    lwu v1, 4(s4)             ;; (set! v1-17 (l.wu (+ arg0 4)))[37] [s4: process ] -> [v1: process-mask ]
    ori v1, v1, 512           ;; (set! v1-18 (logior v1-17 512))[38] [v1: process-mask ] -> [v1: process-mask ]
    sw v1, 4(s4)              ;; (s.w! (+ arg0 4) v1-18)[39] [v1: process-mask s4: process ] -> []
```

Also good:
```lisp
    (set! v1-17 (-> a1-0 mask))
    (set! v1-18 (logior v1-17 512))
    (set! (-> a1-0 mask) v1-18)
```

Wrong:
```lisp
set! (-> arg0 mask) (logior (the-as int (-> arg0 mask)) 512))
```
if it attempted to cast the `512` to a `process-mask` it would get decompiled properly.


decompiler/config/jak1_ntsc_black_label.jsonc
```diff
@@ -3,20 +3,20 @@
 {
   "game_version":1,
   // the order here matters (not sure that this is true any more...). KERNEL and GAME should go first
-  "dgo_names":["CGO/KERNEL.CGO","CGO/GAME.CGO",
+  "dgo_names_":["CGO/KERNEL.CGO","CGO/GAME.CGO",
```
water111, line 6: revert this!

water111:
A bunch of bitfield enum things now work properly

operations with `logand` and `logior`
```lisp
;; old
(nonzero? (logand (-> arg0 mask) 256))
;; new
(nonzero? (logand (-> arg0 mask) (process-mask process-tree)))
;; old
(set! (-> pp mask) (logior (-> pp mask) 1024))
;; new
(set! (-> pp mask) (logior (-> pp mask) (process-mask going)))
```

operations with `logand` and `lognot` that were constant folded:
```lisp
;; old
(set! (-> obj mask) (logand -961 (the-as int (-> arg0 mask))))
;; new
(set!
   (-> obj mask)
   (logand
    (lognot (process-mask sleep sleep-code process-tree heap-shrunk))
    (-> arg0 mask)
    )
   )
```

Setting a value:
```lisp
;; old
(set! (-> a0-57 mask) (the-as uint #x4011c))
;; new 
(set! (-> a0-57 mask) (process-mask pause menu progress process-tree camera))
```

Still needs support for static data, and for non-bitfield enums.

PR #382, opened 2021-04-24
[Decompiler] WIP: Stack Spills

water111:
The stack slot identification and type analysis passes seem to do something reasonable. The next step is to build better expressions.  Maybe there should be separate stack access functions?

It also doesn't help that all the example functions are super complicated and full of casts and bitfields.


water111:
The `disasm-vif-tag` function is looking promising.  Remaining to-do:
- [ ] Option to rename stack variables.
- [ ] Decompiler figures out the type of stack variables
- [ ] Casts when setting/getting stack variables
- [ ] Option to retype stack variables?

This function is always going to be full of magic integer constants until https://github.com/water111/jak-project/issues/380 is resolved.

water111:
It is much closer now. `disasm-vif-tag` is working and `disasm-dma-list` is very close. The bitfield access is all matched.
- [ ] There is a missing cast setting a stack variable that's blocking `disasm-dma-list` from compiling
- [ ] Some general cleanup of types might make things more readable
- [ ] Some dead set eliminations are missing when the source variable is a stack variable
- [ ] There's a missing trick for checking `(= thing <some-enum-constant>)`:
```lisp
(zero? (+ (the-as uint (-> current-tag id)) (the-as uint -3)))
```
from
```asm
    lq v1, 112(sp)
    dsll32 v1, v1, 1
    dsrl32 v1, v1, 29
    daddiu v1, v1, -3
    daddiu a0, s7, 8
    movn a0, s7, v1
```

PR #394, opened 2021-04-28
[Decompile] vector.gc (finally)

water111:
I _think_ we now have everything to decompile `vector.gc`.

Current this is about half of it + tests.  Everything out of the decompiler seems to be correct.  Most of the work is figuring out their crazy naming conventions (what does `vector-v*float++!` mean??) and writing reasonable doc strings.


water111:
`vector-cross!` will return junk in `w` - need to update the tests to not print this.

PR #405, opened 2021-05-01
repl: Add auto-listening capabilities `-auto-lt`

### LINKED ISSUE #403: [REPL] Flag to start the repl and auto-connect to the listener

xTVaser:
Closes #403

Speeds up the workflow a bit.  Wasn't sure about the `break;` in the arg parsing, hope that doesn't break an assumption/expectation?

image


water111:
Thanks! I think the `break` wasn't supposed to be there in the first place, so no harm in removing it.

PR #410, opened 2021-05-01
[Decompile] connect, text-h, settings-h, capture, memory-usage-h

water111:
Previously I had skipped `capture` due to bitfields/stack spills, but now it works!


water111:
Todo before merging this:
- [ ] Figure out what's going on with `relocate`.  This method is defined for `object` (root of type tree) but this slot clearly holds methods with slightly different types (`relocate` process by the kernel takes an offset and `relocate` something loaded by the linker takes a `kheap`).
- [ ] Update `texture-h.gc` and all of the other OpenGOAL code for type changes
- [ ] Fix up bugs in `texture.gc` as needed
- [ ] Write up git issue for what went wrong in the decompiler for `texture`.

water111:
- Fixed wrong types in `mem-usage` method
- Fixed decompiler crash when stack variable is set to 0 and other stack spill related issues
- Added `make-u128` in the decompiler for 128-bit constants generated in GOAL with a `ld`, `ld`, `pcpyld` sequence (and added a macro in the compiler)
- Fixed boxed array `new` using `array` instead of `boxed-array` in the decompiler
- Removed some old debug information from the decompiler that was not used
- Fixed some "could not construct variable from `r0` errors"
- Partial decompilation of `texture.gc`
- Decompile `capture.gc`
- Fix compiler error when declaring a type with a 128-bit bitfield member
- Offline test for connect, text-h, settings-h, capture, and memory-usage-h

PR #420, opened 2021-05-06
decomp: `lights.gc`

xTVaser:


goal_src/engine/gfx/lights.gc
```diff
@@ -5,3 +5,97 @@
...
+         )
+    (set! (-> out levels x) (+ f0-2 (* clamped-alpha (- f1-2 f0-2))))
+    )
+   )
+  out
+  )
+
+(defun
+  light-group-slerp
+  "Linearlyl interpolate between each of the 4 vector components of two light-groups.
```
water111, line 29: The docstring has to come after the arguments in a `defun`.

Also, I usually like to put all the arguments on the same line. Something like
```lisp
(defun light-group-slerp ((out light-group) (a light-group) (b light-group) (alpha float))
  "docstring"
   body
  )
```

If you use emacs, you can use alt + shift + 6 to quickly combine two lines.

goal_src/engine/gfx/lights.gc
```diff
@@ -5,3 +5,97 @@
...
+   Alpha is clamped between 0 - 1."
+  ((out light-group) (a light-group) (b light-group) (alpha float))
+  (dotimes (vec-index 4)
+   (light-slerp
+    (the-as light (+ (+ (* 48 vec-index) 0) (the-as int out)))
```
water111, line 34: This is a little bit weird.  It's basically taking an object, converting it to an integer, adding an integer, and converting it back to an object.  This will do the right thing, but probably isn't what they wrote.

To me, this looks like an array access. It's basically doing `(+ object (* index array-stride))`.  In GOAL its possible and somewhat common to have overlapping fields to give you multiple ways to access the same data. For example, in this case, they have 4 lights: `dir0`, `dir1`, `dir2`, `ambi`. My guess is that they also defined an array of 4 lights that overlaps with these, so they could access `dir2` by doing `(-> group dir2)` or by doing `(-> group lights-array 2)`.  They could optionally hide these overlay fields from the inspect methods, so sometimes you have to add it yourself. You could add an array field like this:
```lisp
(deftype light-group (structure)
  ((dir0 light :inline :offset-assert 0)
   (dir1 light :inline :offset-assert 48)
   (dir2 light :inline :offset-assert 96)
   (ambi light :inline :offset-assert 144)
   (lights light 4 :inline :offset 0) ;; this array starts at the same offset as dir0
   )
  :method-count-assert 9
  :size-assert         #xc0
  :flag-assert         #x9000000c0
  )
```

I suspect that the arguments will change from `(the-as light (+ (+ (* 48 vec-index) 0) (the-as int out)))` to `(-> out lights vec-index)`.

goal_src/engine/ps2/memcard-h.gc
```diff
@@ -32,6 +32,25 @@
   :flag-assert         #x900000044
   )
 
+(defmethod inspect mc-file-info ((obj mc-file-info))
+  (format #t "[~8x] ~A~%" obj 'mc-file-info)
```
water111, line 36: The compiler autogenerates these so there's no need to add them again, even though they show up in the decompilation.

water111 (commented):
This is getting close!

If you want to test compiling locally, you can run the goal compiler and just do `(build-game)` at the REPL.  This will be faster and have better error messages than running the full test.

goal_src/engine/gfx/lights.gc
xTVaser: oh cool, that's pretty interesting.  and yeah the code definitely did not look like a sane person would have wrote it agreed lol

PR #422, opened 2021-05-06
[decomp] `prototype-h`, a few enums and other random things

ManDude:


ManDude:
`res` still needs a lot of work and `fact-h` has some weird things (maybe `go` stuff)

PR #424, opened 2021-05-07
[decomp] headers for generic, mercneric, tieneric, merc, cpu shadow renderers, art, joint and bone

ManDude:


water111:
Nice, that's a lot of types!

PR #428, opened 2021-05-09
Decompilation: `shadow-h` and `target-h`

xTVaser:
Based off https://github.com/water111/jak-project/pull/427 so i could use the automated testing changes i made (last PR i do like this)


goal_src/engine/target/target-h.gc
```diff
@@ -5,3 +5,58 @@
...
+  :size-assert         #x248
+  :flag-assert         #x1500000248
+  (:methods
+    (dummy-20 () none 20)
+    )
+  )
+
+;; failed to figure out what this is:
+(when (or (not *target*) (zero? *target*))
+  (set! *target* #f)
```
water111, line 45: This is probably a `define-perm`. See `goal-lib.gc`.

PR #439, opened 2021-05-11
decomp some of `res` to get started

ManDude:


water111:
I'm pretty sure this fails because I haven't done https://github.com/water111/jak-project/issues/443 yet.

The `obj` arguments to the `res-tag` methods and the `(tag res-tag)` argument to `dummy-14` need to use 128-bit registers, but right now they use 64-bit registers, so it causes an error.

PR #441, opened 2021-05-11
decomp: `level-info`

xTVaser:
This decompiled seemingly without issues, output looks pretty readable, just a bunch of keyword-value data.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -29,7 +29,9 @@ tasks:
       - task: decomp
   decomp-list:
     cmds:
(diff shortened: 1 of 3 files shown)


water111:
I'd like to wait until https://github.com/water111/jak-project/issues/442 is fixed before we decompile this.  Right now the decompiler spits out `(quote x)` instead of `'x` for all these statics.

And also something appears broken in the compiler, based on the tests.

xTVaser:
No longer needed

PR #473, opened 2021-05-13
[decomp] some of `fact-h`

ManDude:


goal_src/engine/util/types-h.gc
```diff
@@ -54,17 +54,25 @@
...
   )
 
+(defglobalconstant TICKS_PER_SECOND 300) ;; 5 t/frame @ 60fps, 6 t/frame @ 50fps
+
 (defmacro seconds (x)
   "Convert number to seconds unit."
   (cond
     ((float? x)
-     (* 300.0 x)
+     (* (* 1.0 TICKS_PER_SECOND) x)
```
water111, line 63: I think this can be `(* 1.0 TICKS_PER_SECOND x)`.
Both GOOS and OpenGOAL will use the type of the first argument as the math type.

PR #488, opened 2021-05-15
[decompiler] Get rid of FormStack peek and fix virtual method calls

water111:
I was expecting this to be much harder, so I'm a little suspicious that this isn't going to work 100% of the time.

### DIFF
--- a/decompiler/IR2/AtomicOp.h
+++ b/decompiler/IR2/AtomicOp.h
@@ -636,12 +636,10 @@ class CallOp : public AtomicOp {
   void collect_vars(RegAccessSet& vars) const override;
   const std::vector<RegisterAccess>& arg_vars() const { return m_arg_vars; }
   RegisterAccess function_var() const { return m_function_var; }
-  bool is_virtual_method() const { return m_is_virtual_method; }
 
  protected:
   TypeSpec m_call_type;
   bool m_call_type_set = false;
-  bool m_is_virtual_method = false;
 
   std::vector<RegisterAccess> m_arg_vars;
   RegisterAccess m_function_var;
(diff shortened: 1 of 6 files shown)


water111:
Yeah this didn't work. The issue would be if the method call is falsely "split".

```lisp
(defmacro set-vector! (vec x y z w)
  `(begin
     (set! (-> vec x) ,x)
     (set! (-> vec y) ,y)
     (set! (-> vec z) ,z)
     (set! (-> vec w) ,w)
   )
  )

;; method call with a macro used to generate an argument
(some-method obj (set-vector! (new 'stack 'vector) 1. 2. 3. 4.))

(let ((x (method-of-object obj some-method))
      (vec (new 'stack 'vector)))
  (set! (-> vec x) 1.)
  (set! (-> vec y) 2.)
  (set! (-> vec z) 3.)
  (set! (-> vec w) 4.)
  (x vec) ;; missing obj
  )
```

I _really_ don't like the peek system.  It seems like it would be better to ditch it and treat a method call as normal use of variables.  Then, if it makes a mess with multiply accessing an object for the vtable and the `this` argument, we can clean that up manually in a later pass (or part of `let` pass).  When it was first implemented, there was no `let` detection, which would make the clean up much harder.  Now I think this is the better approach, and would let us delete all the "peek" junk.

water111:
Will close https://github.com/water111/jak-project/issues/255

This finally ditches the awful peek system and weird special casing for virtual methods.  The new strategy is to treat a virtual method just like any other method, but when popping the object, we allow it to be used twice.  This wasn't possible in the original framework because we didn't have SSA based use/def counts. 

I originally tried to fix this up at the `let` stage, but that didn't work when the method call was embedded in another expression - the temp var for the `this` blocked other unrelated variables from propagating properly.

PR #492, opened 2021-05-16
decomp: Decompile `transformq`

xTVaser:
I left out the reference test until https://github.com/water111/jak-project/issues/487 is resolved

Also, I need to brush up (and by brush up I mean i know nothing 😄 ) on my matrix / quaternion related math to accurately name these methods.  Left them with stub names for now.


goal_src/engine/math/transformq.gc
```diff
@@ -5,3 +5,536 @@
...
+          (a1-2 (-> *standard-dynamics* gravity-normal))
+          (v1-2
+           (vector-normalize!
+            (vector-flatten! (new 'stack-no-clear 'vector) a1-2 gp-0)
+            1.0
+            )
+           )
+          )
+    (vector-dot v1-2 s5-1) ; added manually, function was inlined.
+    (let ((f0-2 f0-1))
```
water111, line 274: I think one of these was set by `vector-dot`.  Right now this uses the `f0-1` and `f1-2` local vars without ever setting them to a value.

goal_src/engine/math/transformq.gc
```diff
@@ -5,3 +5,536 @@
...
+    s3-0
+    (vector-normalize-copy! (new 'stack-no-clear 'vector) arg0 1.0)
+    (vector-y-quaternion! (new 'stack-no-clear 'vector) s3-0)
+    )
+   )
+  )
+
+;; definition for method 11 of type trsqv
+(defmethod TODO-RENAME-11 trsqv ((obj trsqv))
+  (local-vars (a1-0 none) (a2-0 none) (a3-0 none))
```
water111, line 127: this has more arguments (`vector`, `float`, `int`)

goal_src/engine/math/transformq.gc
```diff
@@ -5,3 +5,536 @@
...
+  (the-as quaternion (-> obj rot))
+  )
+
+;; definition for method 18 of type trsqv
+(defmethod set-quaternion!-18 trsqv ((obj trsqv) (arg0 quaternion))
+  (quaternion-copy! (get-quaternion obj) arg0)
+  )
+
+;; definition for method 21 of type trsqv
+(defmethod swap-dir-targ-and-rot-quaternions!-21 trsqv ((obj trsqv))
```
water111, line 50: This isn't really a swap, it's just setting `dir-targ` to `rot`.

goal_src/engine/math/transformq.gc
```diff
@@ -5,3 +5,536 @@
...
+   (set! (-> s2-0 z) (cos arg0))
+   (set! (-> s2-0 w) 1.0)
+   (s3-0 obj s2-0 arg1 arg2)
+   )
+  )
+
+;; definition for method 14 of type trsqv
+(defmethod TODO-RENAME-14 trsqv ((obj trsqv) (arg0 float))
+  (let ((s5-0 (method-of-object obj TODO-RENAME-19))
+        (s4-0 (new 'stack-no-clear 'quaternion))
```
water111, line 196: probably a `vector`?

water111 (changes requested):
Looks good! I had a few comments on types, but it is close.

For now stub names are fine. Some of these call functions in `geometry` that I haven't figured out yet, so it would be hard to figure out what all these do.  Types that use a lot of methods are always really confusing and this was one of the first method-heavy files we've done.

PR #494, opened 2021-05-16
`(init-vf0-vector)` to make vector code a bit more readable

xTVaser:
After some discussion, we are probably abandoning https://github.com/water111/jak-project/issues/445 because we can't really get it to performance any better.  From Agner Fog's writings:
> There is no instruction for moving a constant into avector register. The default way of putting a constant into a vectorregister is to load it from a memory constant. This is also the most efficient way if cache misses are rare. But if cache misses are frequent then we may look for alternatives

But we can atleast make it a bit more readable, turns out i already had a TODO around this anyway.

I'm not sure if i defined the macro in the write place, because i then had to put it in the forward-declarations for tests to pass.


test/goalc/source_templates/with_game/test-vf-constant-loading.gc
```diff
@@ -0,0 +1,14 @@
+(defun test-vector-constant ()
+  (let ((vector-1 (new 'stack 'vector)))
```
water111, line 2: Is this file used anywhere?

xTVaser: my bad, was a left over when i was testing the new ASM instructions that have been decided against.

PR #499, opened 2021-05-17
decomp: decompile `trajectory`

xTVaser:
I'm pretty confident about the signature for `add-debug-line` but might be worth an extra pair of eyes.


goal_src/engine/physics/trajectory.gc
```diff
@@ -5,3 +5,172 @@
...
+          (a0-6 #t)
+          (a1-1 68)
+          (a2-1 s5-0)
+          (a3-0 s4-0)
+          (t0-0 #x800000ff)
+          (t1-0 #f)
+          )
+     (let ((t2-0 -1))
+      )
+     (t9-1 a0-6 a1-1 a2-1 a3-0 (the-as int t0-0) t1-0)
```
water111, line 171: based on the code here, i'd guess that `add-debug-line` takes one more argument (possibly unused inside the function).   
If you try that, does this get a lot nicer?

goal_src/engine/physics/trajectory.gc
```diff
@@ -5,3 +5,172 @@
...
+    (let
+     ((f0-1 (* (-> obj time) (/ (+ 1.0 (the float s2-0)) (the float s3-0)))))
+     ((method-of-type trajectory TODO-RENAME-9) obj f0-1 s4-0)
+     )
+    (let ((t9-1 add-debug-line)
+          (a0-6 #t)
+          (a1-1 68)
+          (a2-1 s5-0)
+          (a3-0 s4-0)
+          (t0-0 #x800000ff)
```
water111, line 166: This looks like a rgba color to me. If you change the type of `add-debug-line` to make this argument an `rgba`, it should convert this to `(new 'static 'rgba :r #xff :a #x80)` automatically.

goal_src/engine/physics/trajectory.gc
```diff
@@ -5,3 +5,172 @@
...
+   )
+  (none)
+  )
+
+;; definition for method 15 of type trajectory
+;; INFO: Return type mismatch int vs none.
+;; Used lq/sq
+(defmethod TODO-RENAME-15 trajectory ((obj trajectory))
+  (let ((s5-0 (new 'stack-no-clear 'trajectory))
+        (s4-0 (new 'stack-no-clear 'trajectory))
```
water111, line 147: I'm guessing these are `vector` instead of trajectory?

decompiler/config/all-types.gc
```diff
@@ -34131,7 +34131,7 @@
 ;;(define-extern *debug-sphere-table* object) ;; unknown type
 (define-extern add-debug-sphere-from-table function)
 (define-extern debug-sphere-table type)
-(define-extern add-debug-line function)
+(define-extern add-debug-line (function symbol int trajectory trajectory int symbol int))
```
water111, line 34134: I think this should be 
```
(define-extern add-debug-line (function symbol int vector vector rgba symbol int int))
```

water111 (commented):
I think `add-debug-line` might not be right but everything else looks good

PR #506, opened 2021-05-20
[decomp] Add tool to print type information from a memory dump

water111:


water111:
codacy failure is a false positive. The check is here:
https://github.com/david-a-wheeler/flawfinder/blob/293ca17d8212905c7788aca1df7837d4716bd456/flawfinder#L1057 which mistakenly recognizes me defining a template method called `read` as the POSIX `read` function being called.

PR #507, opened 2021-05-20
Automatically watch PCSX2 savestate directory and analyze the EE Memory

xTVaser:
I can convert elzip to a submodule if desired, it's not _too_ big though.

Usage:
```
task watch-pcsx2 SAVESTATE_DIR="PATH_TO_PCSX2_SAVESTATE_DIR"
```
> Ideally run from the root of the repo

Everytime a save-state is saved, it will be processed and `ee-analysis.log` will be updated in the root of the repo


water111:
Nice!

I think having a zip library available is a good idea.  

I'm not a huge fan of the `11zip` wrapper, but it sounds like it works and the `minizip` library it uses internally seems like a good choice, so lets go with this for now.  If we ever compress assets (DGOs/textures compress really well!)  we might want to revisit this.

PR #510, opened 2021-05-20
[OpenGOAL] Support for 128-bit arguments and return values (WIP)

water111:
Checking to see if windows tests pass...


water111:
codacy failure is "scope of variable can be reduced" which is true, but I don't want to.

PR #516, opened 2021-05-22
[decompiler] support break

water111:
Currently as a `begin` ending with a `goto`.


water111:
- Add some features to `CfgVtx.cpp` to support getting the destination of a `break` as a block index and figuring out which blocks need labels
- Add labels as needed in cfg_builder based on the data from `CfgVtx.cpp`
- Fix a bug where accessing a bitfield with a single right shift would give `int` in the decompiler always
- Convert a `break` into code + a `goto`.  I think in all cases random `goto` forwards end up turning into `break`s, so trying to convert these into some block-like structure isn't always possible.
- Recognize converting a symbol into a string.
- Don't crash on failed stack spills
- Clean up dead code in a `break` like we did for return
- Decompile method 19 of res-lump. This now can be copy-pasted from the decompiler!

PR #523, opened 2021-05-25
Organize all types

xTVaser:
Closes - https://github.com/water111/jak-project/issues/455


water111:
So I don't forget - we need to update the symbol order so that types go with the `deftype`, not the first use. Also it would be nice to filter out symbols that have no type and are only use like `'blah`.

PR #527, opened 2021-05-25
[decomp] more of `res` + change a few macros

ManDude:


CMakeSettings.json
```diff
@@ -22,12 +22,12 @@
...
 			"name": "Release",
 			"generator": "Ninja",
-			"configurationType": "RelWithDebInfo",
+			"configurationType": "Release",
 			"buildRoot": "${projectDir}\\out\\build\\${name}",
 			"installRoot": "${projectDir}\\out\\install\\${name}",
 			"cmakeCommandArgs": "",
 			"buildCommandArgs": "",
-			"addressSanitizerEnabled": true,
+			"addressSanitizerEnabled": false,
```
xTVaser, line 30: should we make a separate config to use in the windows CI build to still use ASan? or do we no longer care.

water111: I think it would be nice to have asan tests on windows still, but that can wait for later.  We still have them under the clang linux build.

ManDude: I wasn't able to notice any difference with asan on/off for the github testing VM (that makes sense, 90% of the time is spent compiling and the tests take less than a minute) so it could probably be left on for those

PR #529, opened 2021-05-26
[decomp] even more `res`

ManDude:


ManDude:
Remaining 5 methods require decompiler fixes.

goalc/compiler/compilation/Math.cpp
```diff
@@ -14,6 +14,10 @@ MathMode Compiler::get_math_mode(const TypeSpec& ts) {
     return MATH_FLOAT;
   }
 
+  if (m_ts.tc(m_ts.make_typespec("pointer"), ts)) {
+    return MATH_POINTER;
```
water111, line 18: I don't think we want to add another math mode.  If we did, we'd need to update all the other math forms to throw an appropriate error if they got a pointer (right now, they would `assert(false)`) The idea behind math modes was that the "mode" of a math operation is determined by the type of the first argument, and then all the rest of the arguments are converted to the same mode with `to_math_type`. 

This doesn't really apply to `logand` after this change so it might be easier to do something like:
`if (is_integer(first->type()) ` and similar.

test/decompiler/reference/kernel/gkernel_REF.gc
```diff
@@ -164,26 +164,14 @@
...
-                                     pointer
-                                     (+
-                                      (+
-                                       v1-2
-                                       (the-as int (-> type-to-make size))
-                                       )
-                                      arg2
-                                      )
-                                     )
+                                    (&+ (&+ v1-2 (-> type-to-make size)) arg2)
```
water111, line 172: Nice! This looks much better now that it doesn't convert to an int!

goal_src/engine/data/res.gc
```diff
@@ -402,13 +429,184 @@
...
+          (let* ((tag (-> obj tag (-> tag-pair lo)))
+                 (tag-type (-> tag elt-type))
+                 (data (make-property-data obj time tag-pair buf-addr))
+                 )
+              (if tag-addr
+                (set! (-> tag-addr 0) tag)
+                )
+              (cond
+                ((type-type? tag-type float)
+                 (set! default (the-as uint128 (deref float data)))
```
water111, line 542: I want to double check the code generated for this case and make sure it matches GOAL in terms of what happens in the upper bits. I believe the correct behavior is that bits 32-64 are sign extended and 64-128 are 0.  But I suspect that bits 32-64 will end up zero in OpenGOAL.  I'm not sure if it really matters - the only thing I think of is comparing `float`s as integers will behave differently.

water111 (commented):
Nice, this file is getting close!

goal_src/engine/data/res.gc
ManDude: Reading the disassembly is a bit too complex for me, but here is the disassembled output of the entire `res` file:
[test_disasm.txt](https://github.com/water111/jak-project/files/6580027/test_disasm.txt)

ManDude: According to the Intel manuals (and if I'm reading right), `movss` zero-extends bits 32 to 127 when reading from memory, but leaves them unmodified if reading from registers, so the end result for the `set!` is the latter, but its return value is zero-extended. Other than that, everything seems fine.

water111: Thanks! I took a look and the 128-bit stuff is working as expected. It's correctly treating `default` as a 128-bit variable and doing all the conversions as expected.  As written, if the size is 16, it will return a 128-bit value.  I'm not 100% sure that the original GOAL code does this correctly - there may have been a GOAL compiler bug that causes the upper 64-bits to be lost due to `or` being used instead of `por` in some cases.  The decompiler should eventually catch this for us (https://github.com/water111/jak-project/issues/549) so we can know for sure.

There are a few very minor issues (https://github.com/water111/jak-project/issues/548 https://github.com/water111/jak-project/issues/550), but I think we should merge this PR first and deal with those later.

PR #533, opened 2021-05-28
decomp: `credits`

xTVaser:
Decompiler output is mishandling the `*font-work*`


decompiler/config/all-types.gc
```diff
@@ -34619,7 +34621,7 @@
 ;;(define-extern mc-create-file object) ;; unknown type
 ;;(define-extern get-heap object) ;; unknown type
 ;;(define-extern mc-save object) ;; unknown type
-(define-extern print-game-text function)
+(define-extern print-game-text (function string font-context symbol int int float)) ; TODO decomp error, this seems correct though
```
xTVaser, line 34624: `;; WARN: Failed to convert to atomic ops: Variable could not be constructed from register s7` when trying to decompile this function in `text`

goal_src/engine/ui/credits.gc
```diff
@@ -5,3 +5,240 @@
...
+  (dotimes (v1-0 4)
+   (let ((f0-2 (* 64.0 arg0)))
+    (if (< 128.0 f0-2)
+     (set! f0-2 128.0)
+     )
+    (let
+     ((a1-6
+       (logior
+        (logand
+         (l.wu (+ (+ (* v1-0 4) (the-as int *font-work*)) 2496))
```
xTVaser, line 22: this fails to compile, maybe needs a nudge with a type cast?  This is a global and im pretty sure it is indeed a `font-work` not a numeric type as the code here is trying to coerce it into

decompiler/config/all-types.gc
water111: Issue here: https://github.com/water111/jak-project/issues/537

goal_src/engine/ui/credits.gc
water111: I'm not sure what's going on here. This is setting a bitfield in a `char-color` in the `color-table` of the font work, but the decompiler can't figure out the load from the color-table.  I probably won't have a chance to take a closer look for a while, so I'm making an issue here: https://github.com/water111/jak-project/issues/538

water111: This should be fixed now.  Previously:
```lisp
(let
 ((a1-6
   (logior
    (logand
     (l.wu (+ (+ (* v1-0 4) (the-as int *font-work*)) 2496))
     (the-as uint -256)
     )
    (the-as uint (shr (shl (the int f0-2) 56) 56))
    )
   )
  (a2-6 (+ (* v1-0 4) (the-as int *font-work*)))
  )
 (s.w! (+ a2-6 2496) a1-6)
 )
```

now:
```lisp
(set! (-> *font-work* color-table 32 color v1-0 r) (the int f0-2))
```

decompiler/config/all-types.gc
```diff
@@ -4885,7 +4885,7 @@
 (define-extern *text-group-names* (array string))
 (define-extern kheap type)
 (define-extern *common-text-heap* kheap)
-(define-extern *common-text* symbol)
+(define-extern *common-text* game-text-info) ; TODO guess, but its definitely not just a symbol!
```
water111, line 4888: Yeah that one was my bad guess a long long time ago.

PR #539, opened 2021-05-30
Decompilation: Next batch of files

xTVaser:
- `collide-frag-h`
  - the array types are weird/confusing leading to what seems to be an incorrect definition (inspect method looks odd!)
- `collide-func`
  - expression building fails in variable pass `variable pass failed on ray-plane-intersect: invalid unordered_map<K, T> key`


test/decompiler/reference/engine/collide/collide-frag-h_REF.gc
```diff
@@ -0,0 +1,111 @@
...
+;; definition for method 3 of type drawable-inline-array-collide-fragment
+(defmethod
+  inspect
+  drawable-inline-array-collide-fragment
+  ((obj drawable-inline-array-collide-fragment))
+  (format #t "[~8x] ~A~%" obj (-> obj type))
+  (format #t "~Tid: ~D~%" (-> obj id))
+  (format #t "~Tbsphere: ~`vector`P~%" (-> obj bsphere))
+  (format #t "~Tlength: ~D~%" (-> obj length))
+  (format #t "~Tdata[1] @ #x~X~%" (&-> obj data 4))
```
water111, line 93: There's something weird here.  The 4 should probably be a 0.

water111 (commented):
I don't think it's possible to have an `int32` as a label value.  So far, all the 32-bit constants I've seen are generated without a load and use `lui` and `ori` instead.

xTVaser:
> 
> 
> I don't think it's possible to have an `int32` as a label value. So far, all the 32-bit constants I've seen are generated without a load and use `lui` and `ori` instead.

Here's the code that made me think I had to:
```mips
    lw v1, res-lump(s7)       ;; (set! v1-0 res-lump)[3] [] -> [v1: <the etype res-lump> ]
    lwu t9, 56(v1)            ;; (set! t9-0 (l.wu (+ v1-0 56)))[4]
                              ;; [v1: <the etype res-lump> ] -> [t9: <method (function res-lump symbol symbol int symbol symbol pointer int)> ]
    lwu v1, 116(s5)           ;; (set! v1-1 (l.wu (+ a2-0 116)))[5] [s5: process-drawable ] -> [v1: draw-control ]
    lwu v1, 8(v1)             ;; (set! v1-2 (l.wu (+ v1-1 8)))[6] [v1: draw-control ] -> [v1: art-joint-geo ]
    lwu a0, 12(v1)            ;; (set! a0-1 (l.wu (+ v1-2 12)))[7] [v1: art-joint-geo ] -> [a0: basic ]
    daddiu a1, s7, effect-name;; (set! a1-1 'effect-name)[8] [] -> [a1: symbol ]
    daddiu a2, s7, interp     ;; (set! a2-1 'interp)[9] [] -> [a2: symbol ]
    lw a3, L6(fp)             ;; (set! a3-0 (l.w L6))[10] [] -> [a3: <uninitialized> ]
...
L6:
    .word 0xce6e6b28
```

Doesn't seem like a float, and the 4th arg of that res-lump method is supposedly an `int`

water111:
I think you're right, this is an `int32`.  I have no idea why it's different here.

water111:
Oh, I bet it is actually a `float` and not a `int32`.

decompiler/config/all-types.gc
```diff
@@ -10925,82 +10925,67 @@
...
+   (total-qwc       uint8          :offset 30)
+   (unused          uint8          :offset 31)
+   )
+  :method-count-assert 9
+  :size-assert         #x20
+  :flag-assert         #x900000020
+  )
 
+(deftype collide-fragment (drawable)
+  ((mesh    basic          :offset 8)
```
water111, line 10952: I think this is a collide-frag-mesh

decompiler/config/all-types.gc
```diff
@@ -10925,82 +10925,67 @@
...
+    (dummy-13 () none 13)
+    (dummy-14 () none 14)
+    (dummy-15 () none 15)
+    (dummy-17 () none 17)
+    )
+  )
 
-;; - Symbols
+(deftype drawable-inline-array-collide-fragment (drawable-inline-array)
+  ((data    uint8       36) ; NOTE - guess!
```
water111, line 10969: I think I figured out the pattern for these `drawable-inline-array`s.  This would be:
```lisp
(deftype drawable-inline-array-collide-fragment (drawable-inline-array)
  ((data    collide-fragment 1 :inline      :offset-assert 32)
   (pad uint32) ;; ending in inline basic always results in 4 byte pad.
   )
...
```

This should make the size correct and make the inspect for `data` look correct.

goal_src/engine/camera/cam-update-h.gc
```diff
@@ -5,3 +5,58 @@
...
+  ;; definition (perm) for symbol *camera-smush-control*, type smush-control
+  (define-perm *camera-smush-control* smush-control
+    (set-zero! (new 'global 'smush-control))
+    )
+
+  ;; failed to figure out what this is:
+  (when (or (not *camera-other-root*) (zero? *camera-other-root*))
+    (let ((v1-34 (new 'global 'vector)))
+    (.svf (&-> v1-34 quad) vf0)
+    (set! *camera-other-root* (the-as transform v1-34))
```
water111, line 55: I think `*camera-other-root*` is a `vector` and not a `transform`.  It wouldn't make sense to cast a vector to a transform - a transform is bigger.

goal_src/engine/collide/collide-frag-h.gc
```diff
@@ -5,3 +5,66 @@
...
+;; definition for method 3 of type drawable-inline-array-collide-fragment
+(defmethod
+  inspect
+  drawable-inline-array-collide-fragment
+  ((obj drawable-inline-array-collide-fragment))
+  (format #t "[~8x] ~A~%" obj (-> obj type))
+  (format #t "~Tid: ~D~%" (-> obj id))
+  (format #t "~Tbsphere: ~`vector`P~%" (-> obj bsphere))
+  (format #t "~Tlength: ~D~%" (-> obj length))
+  (format #t "~Tdata[1] @ #x~X~%" (&-> obj data 4))
```
water111, line 60: The 4 here is a sign that the `data` field is in the wrong spot.

water111 (commented):
I had a few suggestions on types but otherwise looks good.

Issue #540, opened 2021-05-30
[decompiler] ray-plane-intersect fails the variable pass


water111:
Hits a map exception.

Issue #538, opened 2021-05-29
[decompiler] array access within an inlined structure not working.


water111:
```
(deftype char-color (structure)
  ((color rgba 4 :offset-assert 0)
   )
  :method-count-assert 9
  :size-assert         #x10
  :flag-assert         #x900000010
  )

(deftype font-work (structure)
  ((font-tmpl          dma-gif-packet :inline :offset-assert 0)
  ... 
   (color-table        char-color       64  :inline    :offset-assert 1984)
  ...
```

This doesn't work:
```
(-> *font-work* color-table <some-constant> color <some-variable>)
```

In `set-credits-font-color` in `credits.gc`

Issue #537, opened 2021-05-29
[decompiler] Setting a stack variable to false doesn't work.


water111:
In `text.gc`, `print-game-text`

```asm
    sw s7, 208(sp)
    sw s7, 212(sp)
```

The decompiler tries to use `s7` as a variable and fails to recognize it as false. It's an issue at the instruction -> atomic op pass.

water111:
This now shows up as
```
    sw s7, 208(sp)            ;; (stack-store #f :offset 208 :sz 4)[179]
    sw s7, 212(sp)            ;; (stack-store #f :offset 212 :sz 4)[180]
```

Issue #535, opened 2021-05-29
[decompiler] eliminated forms in top level form turn into `(empty-form)`


water111:
they should just be completely eliminated.

See `or v1, s7, r0` in `B3` of `eye-h.gc`'s top level.

PR #541, opened 2021-05-30
[decomp] Small fixes

water111:
- Fixes issue where `local-vars` doesn't show up in the top-level (#498)
- Fixes issue with return value casts missing (#489)
- Fixes stack variable set to false (#537)
- Removes eliminated forms from top level (#535)
- Adds config option for CondWithElse rewrite logic (#530)
- Add missing case in field lookup for nested inline-array where the inner lookup is dynamic and the offset into the outer one is constant propagated to the actual load/store instruction (#538)
- Fix bug causing exception to be thrown in variable pass when Q/ACC is used in certain functions with >1 basic blocks (#540)


water111:
Codacy got confused and failed. After running it again manually it passed (https://app.codacy.com/gh/water111/jak-project/commit?cid=557437275&fileBranchId=23798640).  It seems like codacy fails more often than it finds something :)

PR #544, opened 2021-05-31
docs: Significantly reduce size of image struggling to load

xTVaser:
The background image was taking way too long to load, changed it to a jpg


ManDude:
You can also try something like [PNGGauntlet](https://pnggauntlet.com/) to squeeze PNGs. Tried it on the original background and it went from 6600KB to 1798KB.
image

PR #546, opened 2021-05-31
[decompiler] Add pass to generate a symbol definition map file

water111:
In the config, set `"generate_symbol_definition_map": true,` and run the decompiler. It will generate a `symbol_map.json` file in the output folder.
You can kill the decompiler after seeing `[info] Built symbol map in 224.18 ms` if you don't want to wait for the whole thing to run.

The order within files should now be correct, and `deftype`s will win over load/store use of symbols.  Symbols that aren't `deftype` args and are never loaded/stored won't show up.

Example:
```
  "gkernel-h": [
    "process-tree",
    "stack-frame",
    "uint64",
    "kernel-context",
    "thread",
    "cpu-thread",
    "dead-pool",
    "dead-pool-heap-rec",
    "dead-pool-heap",
    "catch-frame",
    "protect-frame",
    "handle",
    "state",
    "event-message-block"
  ],
```

### DIFF
--- a/decompiler/CMakeLists.txt
+++ b/decompiler/CMakeLists.txt
@@ -11,6 +11,7 @@ add_library(
         analysis/insert_lets.cpp
         analysis/reg_usage.cpp
         analysis/stack_spill.cpp
+        analysis/symbol_def_map.cpp
         analysis/type_analysis.cpp
         analysis/variable_naming.cpp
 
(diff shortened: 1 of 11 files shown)


water111:
Codacy is complaining that I could simplify an `if` condition. It's technically correct but I think it makes the code more confusing to read.

PR #553, opened 2021-06-02
Fixing organization error in all-types

xTVaser:
Patched up the last problem with all-types, and improved the script so it can be consistently ran on `all-types.gc`


ManDude:
I think you need to pull changes, fix the conflicts, and then push again.

xTVaser:
mhm, but i didn't want to cause a bunch of conflicts while things were in flux.  I think I'll wait until https://github.com/water111/jak-project/pull/596 now as well, as my script can easily make the switch to using the parent type instead of defaulting to `basic`

PR #573, opened 2021-06-09
[decompiler] fix missing casts issue

water111:
Due to a stupid error, the "read type" and "write type" of variables weren't merged. Many parts of the later passes assume that a variable has a consistent type.

In the example from https://github.com/water111/jak-project/issues/570, in all places where the variable was read, the type in the register was `sound-rpc-get-irx-version`. At all writes, the register had a `pointer`.
The `let` believed that it was correctly setting the variable to the `pointer`, and all uses believed that it was reading a `sound-rpc-get-irx-version`, so no casts were inserted.

There's an interesting decision to make here on how to pick the final type of the variable.  Two approaches seem to make sense:
- Pick the lowest common ancestor of the read uses and write uses. This makes the variable as general as possible and avoids any possible casts across branches of the type tree.  This is what I meant to do originally.
- Pick the write type.
I went with the first one for now.

Fixed a bug where (cdr <pair>) would incorrectly have type "object" instead of pair when sloppy pair typing was enabled.

Fixed a bug where a double cast would appear when:
- There is a manual cast for a register
- That register is used directly in a `let` as a source
- The destination variable has a different type than the cast.

Fixed some bugs around function calls that didn't return a value, but GOAL still moved a value from the return register.  These type of bugs would look like random insertions of `(the-as none <thing-that-happens-to-be-in-v0>)`'s or failure to properly build expressions shortly after a call to a function returning none.  These should now be fixed.

Unfortunately, fixing these changed some internal details of the decompiler and the numbering of variables has now changed. It only affects variables in register `v0`.


water111:
As a bonus, the `none` function stuff also fixes https://github.com/water111/jak-project/issues/434

water111:
Also fixed a bug where variables appearing in the source of a let definition wouldn't show up in `local-vars` if they needed to.  This allows `enter-state` to work in the decompiler now, though https://github.com/water111/jak-project/issues/493 is still an issue

PR #575, opened 2021-06-10
Decompile: `prototype` | `video` | `vol-h` | `air-h` | `nav-enemy-h` | `rigid-body-h`

xTVaser:
The last 3 header files have issues compiling, but it decompiles fine.  They are just mostly simple type-defs so this seems strange


decompiler/config/all-types.gc
```diff
@@ -23110,26 +23103,26 @@
...
-;    (x-length     float     :offset-assert 16)
-;    (z-length     float     :offset-assert 24)
-;    (sin-angle    float     :offset-assert 28)
-;    )
-;   :method-count-assert 9
-;   :size-assert         #x20
-;   :flag-assert         #x900000020
-;   )
+(deftype air-box (structure)
+  ((vecs         vector 2 :offset-assert 0)
```
water111, line 23107: This is probably `:inline`.

goal_src/engine/game/video.gc
```diff
@@ -5,3 +5,114 @@
...
+     )
+    )
+   )
+  (set!
+   (-> *font-default-matrix* vector 0 x)
+   (-> *video-parms* relative-x-scale)
+   )
+  (set-hud-aspect-ratio arg0 (the-as video-parms (get-video-mode)))
+  (when *progress-process*
+   (let ((s4-1 (method-of-object (-> *progress-process* 0) dummy-23)))
```
water111, line 100: Same here, I think this will look nicer if `dummy-23` took another argument

water111 (commented):
I think those three files compile fine, but aren't decompiled as part of the offline test because we're using the `ENGINE.CGO` file, and those files are only found in `GAME.CGO`. If you change `"CGO/ENGINE.CGO"` to `"CGO/GAME.CGO"` inside of `offline_test_main.cpp`, does it work?

test/decompiler/reference/engine/game/video_REF.gc
```diff
@@ -0,0 +1,111 @@
+;;-*-Lisp-*-
+(in-package goal)
+
+;; definition for function set-video-mode
+;; INFO: Return type mismatch int vs none.
+(defun set-video-mode ((arg0 video-parms))
```
water111, line 6: I think `arg0` here is likely a `symbol` here and not `video-parms`.  It's compared to `'ntsc` and `'pal`.  Both the compiler/decompiler do very little type checking for comparing things for equality, but it doesn't make sense to compare a structure to a symbol.

test/decompiler/reference/engine/game/video_REF.gc
```diff
@@ -0,0 +1,111 @@
...
+     (set! (-> *video-parms* relative-x-scale) 0.75)
+     (set! (-> *video-parms* relative-x-scale-reciprical) 1.3333334)
+     )
+    )
+   )
+  (set!
+   (-> *font-default-matrix* vector 0 x)
+   (-> *video-parms* relative-x-scale)
+   )
+  (set-hud-aspect-ratio arg0 (the-as video-parms (get-video-mode)))
```
water111, line 90: The `set-hud-aspect-ratio` function probably takes a symbol as its second argument, not a `video-parms`.

test/decompiler/reference/engine/game/video_REF.gc
```diff
@@ -0,0 +1,111 @@
...
+  (when *progress-process*
+   (let* ((s5-1 (-> *progress-process* 0))
+          (s4-1 (method-of-object s5-1 TODO-RENAME-23))
+          )
+    (get-video-mode)
+    (s4-1 s5-1 arg0)
```
water111, line 96: Hmm, this still doesn't look quite right.  I'd expect this to look something like:
```lisp
(TODO-RENAME-23 (-> *progress-process* 0) s5-1 arg0 (get-video-mode))
```
I think this function takes an additional argument.

This is a case where it might help to look at the assembly for the method call. The GOAL compiler generally evaluates expressions left-to-right:

First, it gets the object that it will call the method on: `(-> *progress-process*)`
```
    lw v1, *progress-process*(s7)
    lwu s5, 0(v1)
```

Then, it gets the method from the method table:
```
    lwu v1, -4(s5)            ;; (-> s5 type)
    lwu s4, 108(v1)           ;; (-> v1 method-table 23)
```

The first two arguments are just variables and don't require any code to evaluate.  So this is code for the third argument: `(get-video-mode)`
```
    lw t9, get-video-mode(s7) ;; (set! t9 get-video-mode)[39]
    jalr ra, t9               ;; (call!)[40]
    sll v0, ra, 0
```
Next, it moves the result into the argument register:
```
    or a2, v0, r0             ;; (set! a2 v0)[41]
```

And the final step is to insert a bunch of moves that move existing variables into the right registers (the compiler may omit these if the variable happens to be in the right register already):
```
    or t9, s4, r0             ;; (set! t9 s4)[42]
    or a0, s5, r0             ;; (set! a0 s5)[43]
    or a1, gp, r0             ;; (set! a1 gp)[44]
```
and then call the method
```
    jalr ra, t9               ;; (call!)[45]
    sll v0, ra, 0
```

PR #582, opened 2021-06-12
[decompiler] bitfield support for `sound-name`

water111:
Now the decompiler will handle `sound-name=` correctly.

`sound-name` constants are handled as a special case in the decompiler and will decompiled to strings like this:
```lisp
(sound-bank-load (static-sound-name "empty2"))
```
The `static-sound-name` is a GOOS macro that converts a string to the appropriate `sound-name` constant.


water111:
I also added some of the loader RPC in OVERLORD. It's just enough to respond to the irx version check and send a message to the ISO thread when it tries to load a sound bank. For now the ISO thread just ignores the sound bank load and prints a message.

In order to actually load these, we'll need to add them to fake iso.  This would also require
- Some step to copy them from `iso_data`, or maybe we just read them directly from `iso_data`?
- Some option to ignore them if we're running without assets/game data on github actions tests.

PR #590, opened 2021-06-13
[decompiler] Fix pretty print bug and add some documentation on states

water111:


joegoldin:
Thank you for the `process` documentation! Was just noodling around with them today, this clears up a lot. 🙏

PR #594, opened 2021-06-15
[decomp] prepare `joint` stuff

ManDude:
`joint` has giant asm functions, some weird function "splitting" stuff going on, scratchpad usage, some unused types AND a couple more things the decompiler doesn't like. What a pain, but we'll have to get back to it later.


ManDude:
`effect-control-h` will fail the offline compile test because the compiler can't know `entity` is a child of `res-lump`

ManDude:
nasm is down so the windows github build can't run, although all tests pass for me locally
image

PR #602, opened 2021-06-18
Update remote third-party/googletest

0x715C:
#603


water111:
Thanks! It's annoying that googletest sets `-Werror`.

PR #611, opened 2021-06-20
decomp: `generic`

xTVaser:


goal_src/engine/gfx/generic/generic.gc
```diff
@@ -5,3 +5,150 @@
 ;; name in dgo: generic
 ;; dgos: GAME, ENGINE
 
+; This is actually defined in generic-vu1.gc but this file builds first
+(define-extern generic-init-buf (function dma-buffer int uint none))
```
water111, line 9: In general, when there's stuff like this I try to put them in the `-h.gc` files, kind of like you would with C `.h` files, instead of in the file its used in.  It's not a huge deal but I think it will keep things a little bit more organized.

So in this case, it could go in `generic-h.gc`.

goal_src/engine/gfx/generic/generic.gc
```diff
@@ -5,3 +5,150 @@
...
+  (the-as
+   generic-dma-foreground-sink
+   (-> *level* level-default foreground-sink-group 1 merc-sink)
+   )
+  )
+
+;; failed to figure out what this is:
+(set!
+  (-> *generic-foreground-sinks* 6)
+  (new 'static 'generic-dma-foreground-sink :bucket 30)
```
water111, line 80: Not an issue for this PR but we may want to switch these to the `bucket-id` enum at some point.

water111 (approved):

PR #616, opened 2021-06-21
decomp: `ocean-transition`

xTVaser:
A lot of the ocean functions are very similar, after doing this one some of the other ocean files seem a lot easier (though a few of them do have microcode).

Blocked by:
- https://github.com/water111/jak-project/issues/615

There is also some weird patterns here, where those same `uint8`s are passed as arguments to functions that try to convert them into `float`s.  Might be missing something, or just might not be understanding it fully yet.


test/decompiler/reference/engine/gfx/ocean/ocean-trans-tables_REF.gc
```diff
@@ -1123,29 +1123,27 @@
...
-   0.0
-   1.0
-   1.0
-   0.0
-   1.0
-   1.0
-   1.0
-   0.0
+  (the-as (inline-array vector)
+   (new 'static 'array float 16
```
water111, line 1130: we should update the label type to be `(inline-array vector)` here and it should get rid of this cast

water111: looks like you already did this in the non `REF` and maybe this file just needs updating?

decompiler/config/all-types.gc
```diff
@@ -6906,7 +6903,7 @@
 (define-extern *ocean-trans-up-left-table* (pointer float))
 (define-extern *ocean-trans-corner-table* (pointer float))
 (define-extern *ocean-trans-strip-array* (pointer float))
```
water111, line 6905: maybe this is type `ocean-trans-strip-array`?

water111 (commented):
This is really close, just a few small type changes and it should be good to go.

Can you try decompiling `ocean-tables` with these changes to make sure that still works?

decompiler/config/all-types.gc
xTVaser: that would definitely make sense, but it doesn't decompile
```mips
    lw t0, *ocean-trans-strip-array*(s7);; [161] (set! t0-5 *ocean-trans-strip-array*)
                              ;; [] -> [t0: ocean-trans-strip-array ]
    daddu a3, a3, t0          ;; [162] (set! a3-17 (+ a3-16 t0-5))
                              ;; [a3: <integer 0 + (160 x int)> t0: ocean-trans-strip-array ] -> [a3: ocean-trans-strip ]
    dsll32 a3, a3, 1          ;; [163] (set! a3-18 (sll a3-17 33)) [a3: ocean-trans-strip ] -> [a3: <uninitialized> ]
    dsrl a3, a3, 1            ;; [164] (set! a3-19 (srl a3-18 1)) [a3: <uninitialized> ] -> [a3: <uninitialized> ]
    or a2, a2, a3             ;; [165] (set! a2-18 (logior a2-17 a3-19))
                              ;; [a2: <uninitialized> a3: <uninitialized> ] -> [a2: <uninitialized> ]
```
> Function ocean-trans-add-upload-strip failed type prop at op 163: Cannot get_type_int2: (sll a3 33), args ocean-trans-strip and <integer 33>

If I make it a pointer to a ocean-trans-strip-array it works but it still looks like a badly formed array access.

water111: ok, let's leave it at `(pointer float)` for now then.

goal_src/engine/gfx/ocean/ocean-transition.gc
```diff
@@ -5,3 +5,1266 @@
...
+     (let
+      ((a0-11 (+ (* (+ v1-8 1) 16) (the-as int *ocean-trans-corner-table*))))
+      (.lvf vf6 (&-> (the-as (pointer int128) a0-11)))
+      )
+     (let
+      ((a0-14 (+ (* (+ v1-8 5) 16) (the-as int *ocean-trans-corner-table*))))
+      (.lvf vf7 (&-> (the-as (pointer int128) a0-14)))
+      )
+     (let
+      ((v1-11 (+ (* (+ v1-8 6) 16) (the-as int *ocean-trans-corner-table*))))
```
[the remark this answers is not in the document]
water111: this is fine for now, we should probably wait until we understand what this code is doing before getting too specific with the types anyway

PR #654, opened 2021-06-30
decomp: `task-control`

xTVaser:


decompiler/config/all-types.gc
```diff
@@ -15872,40 +15874,41 @@
...
+    (third-flag-bit? (_type_) symbol 13)
+    (TODO-RENAME-14 (_type_) int 14)
+    (clear-all-but-first-flag-bit (_type_) int 15)
     )
   )
 
 (deftype task-control (basic)
   ((current-stage int16  :offset-assert 4)
-   (stage         basic  :offset-assert 8)
+   (current-stage32 int32 :offset 4) ;; I added this, in the top level, it tries to access the first field as an int32
```
water111, line 15895: I don't think this is right. If you remove this, and add this type cast, it should work:

```
  "(top-level-login task-control)" :[
    [165, "v1", "symbol"]
  ],
```
(will require https://github.com/water111/jak-project/pull/668)

The array contains both `task-control`s and `symbol`s, and in this case it was accessing the `type` field of a `symbol`.

water111 (commented):
There are a bunch of static fields like:
```lisp
     :game-task #x60
     :status #x2
     :flags #x6
```

that I think are actually enums.  The `game-task` enum has a `game-task->string` function that you could use to figure out and `status` is probably `task-status`, which also has `task-status->string`. 

It might be interesting to add these and the decompiler should fill out all the static constants automatically.

PR #681, opened 2021-07-07
decomp: `shadow`

xTVaser:


goal_src/engine/gfx/shadow/shadow.gc
```diff
@@ -5,3 +5,339 @@
...
+    81920.0
+    )
+   )
+  0
+  (none)
+  )
+
+;; failed to figure out what this is:
+(set!
+  (-> *part-id-table* 362)
```
water111, line 225: It looks like there's a crash when this code is loaded. Maybe we haven't decompiled the file that creates `*part-id-table*` yet? You could try to find where it's defined and just add that part, or just comment this out for now (and maybe leave a comment near the top of the file so we don't forget this)

xTVaser: ah yeah we havn't. that's either `sparticle` or `sparticle-launcher`

water111:
In order to prevent the crash, you'll need the actual `define`, not just a `define-extern`. The `define-extern` just tells the compiler the type of `*part-id-table*`, but doesn't actually store anything in the symbol.  The top-level code in `shadow` attempts to modify `*part-id-table*`, which crashes because nothing has defined that symbol. The default value of a symbol is 0, and in OpenGOAL, the first 1 MB of memory after 0x0 is intentionally unmapped so it crashes when trying to set a value in `*part-id-table*`.

You can add this to `sparticle-launcher.gc` and it should fix the crash:
```lisp
(define *part-id-table* (new 'global 'boxed-array sparticle-launcher 3584))
(define *part-group-id-table* (new 'global 'boxed-array sparticle-launch-group 1024))
```

PR #688, opened 2021-07-11
decomp: `target-part`

xTVaser:


decompiler/config/jak1_ntsc_black_label/type_casts.jsonc
```diff
@@ -2533,6 +2533,22 @@
   "cam-master-effect": [
     [[0, 999], "s6", "camera-master"]
   ],
+  
+  "birth-func-vector-orient": [
+    [[7, 24], "s3", "sprite-vec-data-2d"]
+  ],
+
+  "process-drawable-burn-effect": [
+    [[0, 999], "s6", "target"],
```
water111, line 2542: does something break if you make this a `:behavior target`?

goal_src/engine/target/target-part.gc
```diff
@@ -5,3 +5,9718 @@
...
+  (local-vars (v1-10 float) (v1-11 float) (sv-16 target))
+  (rlet ((vf0 :class vf)
+         (vf1 :class vf)
+         (vf2 :class vf)
+         )
+   (init-vf0-vector)
+   (set! sv-16 *target*)
+   (let ((s3-0 (new 'stack-no-clear 'vector)))
+    (new 'stack-no-clear 'vector)
+    (let ((s5-0 (new 'stack-no-clear 'vector)))
```
water111, line 136: I think this should be a quaternion.

goal_src/engine/target/target-part.gc
```diff
@@ -5,3 +5,9718 @@
...
+   )
+  )
+
+;; definition for function process-drawable-burn-effect
+;; Used lq/sq
+(defbehavior process-drawable-burn-effect target ((arg0 int))
+  (sound-play-by-name
+   (the-as
+    sound-name
+    (make-u128 #x6465 (-> (new 'static 'array uint64 1 #x6e7275622d746567) 0))
```
water111, line 9281: this looks wrong.  I would set the label cast as `["uint64", true]`, not as an array.

goal_src/engine/target/target-part.gc
```diff
@@ -5,3 +5,9718 @@
...
+        s5-1
+        s2-1
+        (/ (the float v1-8) (the float arg0))
+        )
+       (vector-lerp!
+        (-> (the-as process-drawable (-> s4-1 0)) draw color-mult)
+        s5-1
+        s2-1
+        (-
+         (-> (new 'static 'array float 1 1.0) 0)
```
water111, line 9321: same here, this should probably be `["L...", "float", true]`.  The automatic labeling can't tell the difference between the two on files with far labels.

decompiler/config/jak1_ntsc_black_label/label_types.jsonc
```diff
@@ -443,6 +443,15 @@
     ["L71", "vector"]
   ],
 
+  "target-part": [
+    ["L336", "(pointer float)", 1],
+    ["L337", "(pointer float)", 1],
+    ["L338", "(pointer float)", 1],
+    ["L339", "(pointer float)", 1],
+    ["L340", "(pointer float)", 1],
+    ["L341", "(pointer uint64)", 1]
```
water111, line 452: I think these should drop the `pointer` and be like `["L336", "float", true]`

water111 (commented):
Looks good, just a few comments on label types.
I'm hoping we can make the particle definitions better at some point, they are not really readable at this point.

PR #715, opened 2021-07-24
give `defstate` a parent type and allow anonymous behaviors inside

ManDude:


goal_src/kernel/gstate.gc
```diff
@@ -85,56 +85,84 @@ There are several ways to "go"
...
     )
-    
   )
-  
+
+(defmacro behavior (bindings &rest body)
+  "Define an anonymous behavior for a process state. This may only be used inside a defstate!"
+
+  (let ((behavior-type (first *defstate-type-stack*)))
+      (fmt #t "*defstate-type-stack* top = {} {}\n" behavior-type body)
```
water111, line 161: leftover debugging print?

decompiler/config/all-types.gc
```diff
@@ -37561,10 +37551,10 @@
...
 
 (define-extern *sun-iris-door-sg* skeleton-group)
-;;(define-extern sun-iris-door-open object) ;; unknown type
-;;(define-extern sun-iris-door-closed object) ;; unknown type
-;;(define-extern sun-iris-door-opening object) ;; unknown type
-;;(define-extern sun-iris-door-closing object) ;; unknown type
+(define-extern sun-iris-door-open state)
+(define-extern sun-iris-door-closed state)
+(define-extern sun-iris-door-opening state)
+(define-extern sun-iris-door-closing state)
```
water111, line 37557: Can we make these `(state sun-iris-door)` or does that cause decompiler issues?

goal_src/kernel/gstate.gc
```diff
@@ -85,56 +85,84 @@ There are several ways to "go"
...
-                             :trans #f
-                             :post #f
-                             :enter #f
-                             :event #f
-                             )
+  "Define a new state!"
+
+  (with-gensyms (new-state)
+    (let ((defstate-type (first parents)))
+      (def-state-check-behavior event defstate-type)
```
water111, line 110: Should we `(set! *defstate-type-stack* '())` at the beginning?  If you hit a syntax error in the middle of compiling a behavior it will stop compiling immediately and leave stuff in the stack.

ManDude: I totally forgot about that. I had it set up differently originally and forgot about this when I changed things, I definitely want it to be cleared every time.

decompiler/config/all-types.gc
ManDude: This still works the same as before, I just wanted to check the top-level code for states on an actual random game object.

PR #725, opened 2021-07-27
make some macros for DMA stuff

ManDude:


water111:
These macros are awesome! My only concern is making sure that we don't make a typo when adding them, but I'm not too worried about that.  I checked a few of these randomly and confirmed that `disasm-dma-list` prints the same thing with and without these changes.

PR #741, opened 2021-08-07
decomp: the majority of `navigate` and `rolling-lightning-mole`

xTVaser:


goal_src/engine/nav/navigate.gc
```diff
@@ -5,3 +5,3050 @@
...
+(defmethod TODO-RENAME-18 nav-control ((obj nav-control) (arg0 vector))
+  (dummy-28
+    (-> obj mesh)
+    (vector-! (new 'stack-no-clear 'vector) arg0 (-> obj mesh origin))
+    (-> obj nearest-y-threshold)
+    (the-as (pointer nav-control-flags) #f)
+    )
+  )
+
+(defmethod dummy-20 nav-control ((obj nav-control) (arg0 basic) (arg1 vector) (arg2 vector))
```
water111: arg0 should be `nav-poly` here

goal_src/engine/nav/navigate.gc
```diff
@@ -5,3 +5,3050 @@
...
+            )
+          )
+        )
+      )
+    0
+    )
+  (none)
+  )
+
+(defmethod TODO-RENAME-28 nav-control ((obj nav-control) (arg0 int))
```
water111: `arg0` can be a `collide-kind`

goal_src/levels/common/nav-enemy.gc
```diff
@@ -768,7 +768,7 @@ nav-enemy-default-event-handler
                   )
              (>= (- (-> *display* base-frame-counter) (-> self state-time)) (-> self state-timeout))
              (nonzero? (-> self draw))
-             (logtest? (-> self draw status) (draw-status drwf03))
+            ;;  (logtest? (-> self draw status) (draw-status drwf03))
```
water111, line 771: can you put this in `TARGET_STARTUP_HACKS` as well, just so we don't forget about it?

goal_src/levels/common/nav-enemy.gc
```diff
@@ -832,7 +832,7 @@ nav-enemy-default-event-handler
   (behavior ()
     (when (>= (- (-> *display* base-frame-counter) (-> self state-time)) 30)
       (when (>= (- (-> *display* base-frame-counter) (-> self state-time)) (-> self state-timeout))
-        (if (and (nonzero? (-> self draw)) (logtest? (-> self draw status) (draw-status drwf03)))
+        (if (and (nonzero? (-> self draw)))
```
water111, line 835: same here

goal_src/levels/maincave/mother-spider.gc
```diff
@@ -1558,7 +1558,14 @@
           (set-vector! s1-0 (sin f28-1) 0.0 (cos f28-1) 1.0)
           )
         (vector-float*! s2-1 s1-0 f30-0)
-        (dummy-35 (-> obj nav) (-> obj nav travel) (-> obj root-override trans) s2-1 s1-0 f30-0)
+        (dummy-35
+          (-> obj nav)
+          (-> obj nav travel)
+          (-> obj root-override trans)
+          s2-1
+          (the-as clip-travel-vector-to-mesh-return-info s1-0)
```
water111, line 1566: should probably fix `mother-spider.gc` so this `s1-0` is a `clip-travel-vector-to-mesh-return-info` - otherwise there will be stack corruption when this code runs.

goal_src/levels/rolling/rolling-lightning-mole.gc
```diff
@@ -5,3 +5,1518 @@
...
+          )
+        (vector-normalize-copy! arg1 s3-0 (-> arg0 flee-info deflection-max-dist))
+        )
+      )
+    (set! (-> arg0 nav travel quad) (-> gp-0 quad))
+    s1-0
+    )
+  )
+
+(defun fleeing-nav-enemy-adjust-travel ((arg0 fleeing-nav-enemy))
```
water111, line 298: I think this takes an additional argument that is unused, you could just add an `arg1` `object`.

goal_src/levels/rolling/rolling-lightning-mole.gc
```diff
@@ -5,3 +5,1518 @@
...
+      (t9-8 a0-15)
+      )
+    (fleeing-nav-enemy-chase-post-func)
+    (camera-line-rel
+      (-> self collide-info trans)
+      (-> self desired-travel)
+      (new 'static 'vector4w :x #x80 :w #x80)
+      )
+    (camera-line-rel (-> self collide-info trans) (-> self nav travel) (new 'static 'vector4w :z #x80 :w #x80))
+    (camera-line (-> self collide-info trans) (-> self nav target-pos) (new 'static 'vector4w :z #x80 :w #x80))
```
water111, line 740: probably a project for later, but I think all these `camera-line` things are some weird rgba type.

goalc/compiler/compilation/Asm.cpp
```diff
@@ -272,6 +272,22 @@ Val* Compiler::compile_asm_mov(const goos::Object& form, const goos::Object& res
   return get_none();
 }
 
+Val* Compiler::compile_asm_slt(const goos::Object& form, const goos::Object& rest, Env* env) {
```
water111, line 275: It looks like you ended up implementing it as a macro instead, maybe we should remove this function

test/decompiler/test_FormExpressionBuild2.cpp
```diff
@@ -1161,10 +1161,10 @@ TEST_F(FormRegressionTest, Method4ResTag) {
   std::string type = "(function res-tag int)";
   std::string expected =
       "(the-as int (if (zero? (-> arg0 inlined?))\n"
-      "            (* (-> arg0 elt-count) 4)\n"
-      "            (* (-> arg0 elt-count) (-> arg0 elt-type size))\n"
+      "                (* (-> arg0 elt-count) 4)\n"
+      "                (* (-> arg0 elt-count) (-> arg0 elt-type size))\n"
```
water111, line 1165: why did this change? This looks like the right thing after the pretty printer change, but why wasn't this failing tests earlier?

water111 (commented):
Looks good! I found a few small issues that might explain why it wasn't working correctly yet.

Also, my notes on possible method names (I didn't get all the way through):
```

nav-mesh:
9: tri-centroid-world
finds the centroid of the given triangle, in the "world" coordinate system.

10: tri-centroid-local
finds the centroid of the given triangle, in the local nav-mesh coordinate system.

11: get-adj-poly

12: setup-portal
sets up a portal between two polys.

13: initialize-mesh!

14: move-along-nav-ray!
not sure on the name, but I think this updates the current position in a nav-ray, and updates which
triangle you're in.

15: try-move-along-ray
this takes in a point/direction/distance, and see what would happen if you tried to move this way.
it returns the distance you can go before one of these happens:
- you reach the destination
- you hit a nav mesh boundary/gap
- you cross 15 triangles.

17: update-route-table
(initialization related)

18: ?? something to do with routes.

20: debug-draw-poly

21: point-in-poly?
is the point inside of the triangle?

22: find-opposite-vertices
given two triangles that share an edge, get the indices of the two vertices that aren't part of the edge.

24: closest-point-on-boundary
find the closest point on the perimeter of the triangle.

25: project-point-into-tri-3d
will move a 3D point in space to the surface of this nav-poly

26: project-point-into-tri-2d
Looking from the top down, is the point inside the nav-poly?
if the point is inside the triangle, returns that point.
if the point is outside the triangle, move it to the closest point (will be on the edge)

27: find-poly-fast
finds which triangle the given point is in.
also has some caching stuff so if you look up the same point multiple times, it won't redo the work.
I _think_ this is only an approximate check that may return #f even if you are inside.
But, if it returns a poly, it will be right.

28: find-poly
The accurate version of find-poly (tries find-poly-fast first)

29: is-in-mesh?
checks to see if the triangle is in the mesh or not.
not sure why it's separate from 27 (and such a different implementation). there might be some details I'm missing here.


nav-control
9: debug-draw

17: project-onto-nav-mesh
moves point to nav-mesh.

18: find-poly

20: project-point-into-tri-3d

25: is-in-mesh?
```
you could use these if you want, or leave them as dummy for now.

goal_src/engine/nav/navigate.gc
```diff
@@ -5,3 +5,3050 @@
...
+         (vf15 :class vf)
+         (vf2 :class vf)
+         (vf3 :class vf)
+         (vf4 :class vf)
+         (vf5 :class vf)
+         (vf6 :class vf)
+         (vf9 :class vf)
+         )
+    (let ((t0-0 (-> arg0 poly arg1)))
+      (when (zero? (logand (-> t0-0 pat) 1))
```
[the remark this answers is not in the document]
xTVaser: nah too big, `pat` is only a single byte here

test/decompiler/test_FormExpressionBuild2.cpp
xTVaser: ugh i don't know -- i think i knew at the time but ive forgotten.  my changes seem functionally equivalent except whitespace, maybe it's no longer a problem

PR #746, opened 2021-08-08
decomp: finish `twister`

xTVaser:


goal_src/levels/beach/twister.gc
```diff
@@ -5,3 +5,197 @@
...
+  (let ((s4-0 (new 'stack-no-clear 'matrix)))
+   (dotimes (s3-0 (-> obj num-joints))
+    (let
+     ((s2-0
+       (-> arg0 node-list data (+ (-> obj first-joint) s3-0) bone transform)
+       )
+      )
+     (matrix-rotate-y!
+      s4-0
+      (-> (the-as twister (+ (the-as uint obj) (* s3-0 16))) data 0)
```
water111, line 191: and this would make a lot of sense if this was grabbing `ry`

water111 (changes requested):
Just one small type change and I think this will be ready.

PR #749, opened 2021-08-08
decomp: finish `misty-teetertotter` | `misty-warehouse`

xTVaser:


water111 (approved):
Looks good! I will definitely use this file as an example for working on state stuff in the decompiler.

PR #752, opened 2021-08-09
Decompile font renderer and implement "DirectRenderer"

water111:
This needs cleanup and docs before it can be merged.

But we created our first frame!
image


game/graphics/texture/TexturePool.cpp
```diff
@@ -0,0 +1,195 @@
...
+        texture_record->mip_level = mip_idx;
+        texture_record->w = ww;
+        texture_record->h = hh;
+        texture_record->data_segment = tex.segment_of_mip(mip_idx);
+        texture_record->data.resize(size_bytes);
+
+        m_tex_converter.download_rgba8888(texture_record->data.data(), tex.dest[mip_idx],
+                                          tex.width[mip_idx], ww, hh, tex.psm, tex.clutpsm,
+                                          tex.clut_dest, size_bytes);
+        set_texture(tex.dest[mip_idx], std::move(texture_record));
```
ManDude, line 149: `texture_record` is used after this line for the debug texture dumping, and this seems to be causing some sort of undefined behavior (even if debug output is disabled).

game/graphics/display.cpp
```diff
@@ -117,7 +118,7 @@ void KillDisplay(std::shared_ptr<GfxDisplay> display) {
 
   if (GetMainDisplay() == display) {
     // killing the main display, kill all children displays too!
-    for (int i = 1; i < g_displays.size(); ++i) {
+    for (size_t i = 1; i < g_displays.size(); ++i) {
       KillDisplay(g_displays.at(i));
     }
```
ManDude, line 123: This loop is actually totally wrong, my bad... This is only gonna kill every second sub-display.

game/graphics/texture/TexturePool.cpp
water111: Good catch this is definitely wrong.

water111:
Todo before merging:
- Fix the issue with `link` called from windows (5 arguments)
- Figure out why textures aren't working 100% of the time on windows
- Investigate changing between fonts. This behaves strangely. For example, using `*stdcon*` and having `PAUSE` displayed at the same time usually causes `*stdcon*` to be displayed with the big font.
- dump to file and run from file?

water111:
Remaining issues:
- windows textures
- clang debugger issue
- save to file

The font issue was actually a renderer bug where it was using the wrong TBP.  There were also some memory bugs related to the primitive buffer that are now fixed. 

I added synchronization on the `sync-path` to make sure the rendering is done before attempting to draw the profile bar.  This probably sacrifices a tiny bit of performance but we can get accurate renderer CPU usage now, and the game did this too, for a reason I don't understand.

water111:
Performance analyzer screenshot of the code after sync-path:
image

In this case it's about 2% of a frame.  They cheated 4% off of the profile bar, so this seems reasonable.

PR #762, opened 2021-08-15
decomp: finish `basebutton`

xTVaser:


goal_src/levels/common/basebutton.gc
```diff
@@ -5,3 +5,1127 @@
...
+     )
+    )
+   )
+  )
+
+;; definition for method 29 of type basebutton
+(defmethod
+  TODO-RENAME-29
+  basebutton
+  ((obj basebutton) (arg0 symbol) (arg1 process))
```
water111, line 450: I think arg1 is an entity here, not process

PR #764, opened 2021-08-15
add support for non virtual states

water111:


water111:
There is some additional complexity with virtual states.  You need to be able to do
```lisp
(-> (method-of-type some-type some-state) enter)
```
and get an `enter` that has the appropriate function type, in both the compiler and decompiler.  The type of `(method-of-type some-type some-state)` has enough information to make this possible without a special `state-handler-of-virtual-state` form.

Remaining
- [ ] support `defstate-virtual-hook` in the compiler
- [ ] support `go-virtual-hook` in the compiler
- [ ] macros for the above two
- [ ] compiler access to state enter/code with more detailed types.
- [ ] decompiler access to state enter/code with more detailed types

water111:
Remaining:
- [ ] Some slightly nicer format for virtual gos.
- [ ] Use the `behavior` thing to avoid putting the full lambda?
- [ ] Document the return type of `event` issue.

PR #776, opened 2021-08-20
[decompiler] maybe fix symbol decompile issue

water111:


xTVaser:
fwiw, this does fix a label load issue i was having in `collectables` (it was unable to load the deftype's flags and such).  Not sure if that was the only issue!

PR #778, opened 2021-08-20
decomp: `eye`

xTVaser:


water111:
Just looked really quickly - I think if you made `arg1` in `convert-eye-data` a `uint128`, the issue would go away.

PR #780, opened 2021-08-22
[decomp] `progress`

ManDude:
an issue with far labels is causing `progress` to not decompile properly


ManDude:
will fix #816

ManDude:
will fix #787 and #813

PR #786, opened 2021-08-29
OpenGL Cleanup

doctashay:
Kind of a test PR, renaming GL window and some debug prints.


README.md
```diff
@@ -80,6 +80,11 @@ this decreases the compile and link time from ~10 seconds to ~4 seconds.
 Install packages and init repository:
 
 ```sh
+=======
+Install Packages and Init Repository
+
+```bash
```
water111, line 86: I think if you add this here, you'll need to remove the `sh` on line 82.

doctashay: Not sure how that made its way in there, should be even with master now

PR #812, opened 2021-09-02
decomp: decompile almost all the platform related files

xTVaser:
- plat
- baseplat
- plat-button
- plat-eco (almost, pure states #792 )
- citb-drop-plat-CIT/L1 (l1 omitted intentionally)
- plat-flip
- wedge-plats
- wall-plat


goal_src/levels/citadel/citb-drop-plat-CIT.gc
```diff
@@ -5,3 +5,921 @@
...
+                                                                   (a1-5
+                                                                    drop-plat-init-by-other
+                                                                    )
+                                                                   (a2-4 gp-0)
+                                                                   )
+                                                                  (->
+                                                                   self
+                                                                   duration
+                                                                   )
+                                                                  ((the-as
```
water111, line 702: I think this cast is wrong, it should be `(function process function vector uint uint int)`.

goal_src/levels/common/baseplat.gc
```diff
@@ -5,33 +5,564 @@
...
+       )
+      (set! (-> a0-9 param 1) (-> self speed))
+      (joint-control-channel-group-eval!
+       a0-9
+       (the-as art-joint-anim #f)
+       num-func-seek!
+       )
+      )
+     (if (and gp-0 (rand-vu-percent? 0.5))
+      (spawn-projectile-blue *target*)
```
water111, line 354: looks like the mysterious blue projectile is the blue eco effect that goes from Jak to the door that's being opened.

goal_src/levels/common/plat.gc
```diff
@@ -5,3 +5,603 @@
...
+  (the-as (function none :behavior plat) anim-loop)
+  :post
+  (the-as (function none :behavior plat) plat-post)
+  )
+
+;; definition for method 11 of type plat
+(defmethod copy-defaults! plat ((obj plat) (arg0 res-lump))
+  (set!
+   (-> obj mask)
+   (the-as process-mask (the-as int (logior #x80000 (-> obj mask))))
```
water111, line 547: Is this the output with the latest version of the decompiler? I think this should become a constant now.

goal_src/levels/jungleb/plat-flip.gc
```diff
@@ -5,3 +5,419 @@
...
+
+;; definition for method 11 of type plat-flip
+;; Used lq/sq
+(defmethod copy-defaults! plat-flip ((obj plat-flip) (arg0 res-lump))
+  (local-vars
+   (r0-0 uint128)
+   (a0-17 uint128)
+   (a0-26 uint128)
+   (a1-11 uint128)
+   (sv-16 int)
```
water111, line 230: sv-16 needs to be a res-tag here, otherwise this stack variable will be too small.  You can add a `type_casts.json` entry:
```json
["_stack_", 16, "res-tag"],
```

goal_src/levels/jungleb/plat-flip.gc
```diff
@@ -5,3 +5,419 @@
...
+                                                                     (the-as
+                                                                      uint128
+                                                                      sv-16
+                                                                      )
+                                                                     )
+                                                                    )
+                                                                   (.pcpyud
+                                                                    a0-17
+                                                                    a0-16
+                                                                    r0-0
```
water111, line 296: this use of r0 won't work properly, but I think this should go away when sv-16 is a res-tag

goal_src/levels/jungleb/plat-flip.gc
```diff
@@ -5,3 +5,419 @@
...
+;; definition for method 11 of type plat-flip
+;; Used lq/sq
+(defmethod copy-defaults! plat-flip ((obj plat-flip) (arg0 res-lump))
+  (local-vars
+   (r0-0 uint128)
+   (a0-17 uint128)
+   (a0-26 uint128)
+   (a1-11 uint128)
+   (sv-16 int)
+   (sv-32 int)
```
water111, line 231: sv-32 and sv-48 also needs to be a res-tag

water111 (changes requested):
It's great to see so much gameplay code getting decompiled!

There are just a few suggestions I had for types.  I think you might have reverted the googletest submodule update we did recently as well. I think if you go into `third-party/googletest` and do a `git fetch` and a `git checkout 955c7f83` it will go back to the right version.

goal_src/levels/jungleb/plat-flip.gc
xTVaser: oh interesting, ill have to keep my eye out for this

PR #819, opened 2021-09-02
decomp: `hud-classes`

xTVaser:
- [ ] hud-classes only has a single function remaining
- [ ] collectables is less close, currently has some issues with `defstates`


goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+       )
+      2
+      )
+     (font-color bright-green)
+     (font-flags shadow kerning large)
+     )
+    )
+   (let ((a3-7 (-> s5-0 base)))
+    (let ((v1-8 (the-as object (-> s5-0 base))))
+     (set! (-> (the-as (pointer int64) v1-8)) #x20000000)
```
water111, line 208: you could cast `v1-8` to a `dma-packet` for this store as well, I think it would look nicer

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+        )
+       s5-0
+       )
+      (font-color dark-yellow)
+      (font-flags shadow kerning middle large)
+      )
+     )
+    (let ((a3-12 (-> s3-0 base)))
+     (let ((v1-32 (the-as object (-> s3-0 base))))
+      (set! (-> (the-as (pointer int64) v1-32)) #x20000000)
```
water111, line 687: could also be a `dma-packet`

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+            (s3-0 (when s2-0
+                   (let ((t9-1 (method-of-type manipy activate)))
+                    (t9-1
+                     (the-as manipy s2-0)
+                     obj
+                     'manipy
+                     (the-as pointer #x70004000)
+                     )
+                    )
+                   (let ((t9-2 run-function-in-process)
```
water111, line 752: you shouldn't need a cast here any more

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+            (s4-0 (when s3-0
+                   (let ((t9-1 (method-of-type manipy activate)))
+                    (t9-1
+                     (the-as manipy s3-0)
+                     obj
+                     'manipy
+                     (the-as pointer #x70004000)
+                     )
+                    )
+                   (let ((t9-2 run-function-in-process)
```
water111, line 1051: I think you can remove the cast here

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+                    (t9-3
+                     (the-as manipy s3-0)
+                     obj
+                     'manipy
+                     (the-as pointer #x70004000)
+                     )
+                    )
+                   ((the-as
+                     (function process function object object object object object)
+                     run-function-in-process
```
water111, line 1660: shouldn't need to cast this

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+                                                                   arg0
+                                                                   'hud-pickups
+                                                                   (the-as
+                                                                    pointer
+                                                                    #x70004000
+                                                                    )
+                                                                   )
+                                                                  )
+                                                                 ((the-as
+                                                                   (function process function object object)
```
water111, line 2516: shouldn't require a cast any more

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+                                                               arg0
+                                                               'hud-money
+                                                               (the-as
+                                                                pointer
+                                                                #x70004000
+                                                                )
+                                                               )
+                                                              )
+                                                             ((the-as
+                                                               (function process function object object)
```
water111, line 2555: shouldn't require a cast any more

goal_src/engine/ui/hud-classes.gc
```diff
@@ -7,10 +7,3015 @@
...
+    (-> arg0 root scale)
+    (inc-value arg1)
+    (dec-value arg1)
+    (inc-value arg1)
+    1.0
+    )
+   (set! (-> arg0 draw dma-add-func) dma-add-process-drawable-hud)
+   enter-state
+   (process->handle arg1)
+   (go hud-collecting)
```
water111, line 3013: I think the `hud-collecting` state should take a single argument (`handle`)

water111 (changes requested):
very close, just a few comments on types.  Also I think the gtest submodule got changed by accident here.

goal_src/engine/ui/hud-classes.gc
xTVaser: you're right, it does. However, doing so causes the handler in `hud` to crash.  Probably something that needs to be looked into (file isn't added yet so no harm in adding the arg now)

PR #841, opened 2021-09-09
[game] fix text rendering bugs

water111:


water111:
Centered text is fixed (also credits and the key binding demo now work again)
image

Text width calculation is now correct (previously it was wrong for special characters)
image

Also the sprite bucket is temporarily hooked up to a direct renderer.  The volume bars use this. Eventually this will need to change to also include an actual sprite renderer.

image

PR #842, opened 2021-09-12
[decomp] add mips2c converter

water111:


water111:
It's a miracle, the windows assembly worked on the first try!

PR #846, opened 2021-09-13
[decomp] `flying-lurker`

ManDude:


common/goos/PrettyPrinter.cpp
```diff
@@ -33,7 +33,7 @@ goos::Object float_representation(float value) {
   u8 exp = (int_value >> 23) & 0xff;
   u32 mant = int_value & 0x7fffff;
   if ((exp == 0 && mant != 0) || exp == 0xff) {
-    lg::warn("PS2-incompatible float (0x{:08X}) detected! Writing as the-as cast.", int_value);
+    // lg::warn("PS2-incompatible float (0x{:08X}) detected! Writing as the-as cast.", int_value);
```
xTVaser, line 36: should we just remove this log all together?  I also had it commented out...is this some sort of hack that we are weary of -- because it seems to happen alot

water111: As far as I can tell it only happens in the particle definitions which are kind of a disaster right now

decompiler/config/jak1_ntsc_black_label.jsonc
```diff
@@ -4,6 +4,7 @@
   // if you want to filter to only some object names.
   // it will make the decompiler much faster.
   "allowed_objects": [],
+  "banned_objects": ["crates"],
```
xTVaser, line 7: whats the point of banning crates? can't we just not add it to our tests until its ready to go

ManDude: It's for when I want to run the decompiler on all files for xreferencing symbol usage.

water111 (changes requested):
Looks good! Just one `int` that needs to be a `res-tag`, then it will be good to merge

PR #851, opened 2021-09-21
[decomp] handle `handle->process` inside an `and`

water111:


xTVaser (approved):
solved my issue as predicted

PR #854, opened 2021-09-22
Add zstd library

water111:
The graphics dumps and memory dumps compress extremely well and very quickly with this library.


water111:
Takes 0.5 seconds to compress graphics dumps from ~18 MB to 5 MB.
Takes 0.6 seconds to compress 128 MB main memory dump to 7.3 MB.

PR #865, opened 2021-09-29
decomp: decompile remaining particle files

xTVaser:
### TODO
- the ones that use the `sound-spec` field
- `racer-part` needs `hud-h` fixed

### Done
- [x] beach-part
- [x] citadel-part
- [x] robotboss-part
- [x] sage-finalboss-part
- [x] jungle-part
- [x] flut-part
- [x] firecanyon-part
- [x] village2-part2
- [x] swamp-part
- [x] maincave-part
- [x] sunken-part
- [x] sunken-part2
- [x] sunken-part3
- [x] sunken-part4
- [x] sunken-part5
- [x] ogre-part
- [x] robocave-part
- [x] snow-part
- [x] collectables-part


decompiler/config/all-types.gc
```diff
@@ -32391,7 +32170,7 @@
 (define-extern *BALLOONLURKER-bank* balloonlurker-bank) ;; unknown type
 (define-extern balloonlurker-die (state balloonlurker)) ;; unknown type
 (define-extern *balloonlurker-rudder-joint-home* vector) ;; unknown type
-(define-extern balloonlurker-mine-explode (state balloonlurker)) ;; unknown type
+(define-extern balloonlurker-minfvillageb-part-explode (state balloonlurker)) ;; unknown type
```
water111, line 32173: typo?

decompiler/util/data_decompile.cpp
```diff
@@ -353,16 +353,20 @@ goos::Object decomp_ref_to_inline_array_guess_size(
   int end_label_idx =
       index_of_closest_following_label_in_segment(start_label.offset, my_seg, labels);
 
+  int end_offset = all_words.at(my_seg).size() * 4;
   if (end_label_idx < 0) {
-    throw std::runtime_error(
+    lg::warn(
```
water111, line 358: feel free to remove this warning if you want

goal_src/engine/game/collectables-part.gc
```diff
@@ -5,3 +5,3405 @@
 ;; name in dgo: collectables-part
 ;; dgos: GAME, ENGINE
 
+;; why is this file built first....
+;; TODO - also for entity and beach-rocks
+(deftype collectable (process-drawable)
```
water111, line 10: We shouldn't move the `deftype`s around - these actually generate code that will allocate the type record and add stuff to the symbol table, and we want to keep that stuff in the same order.  Is it possible to replace these with something like 
```
(declare-type collectable process-drawable)
(declare-type eco-collectable collectable)
```
?

goal_src/engine/ui/progress/progress-part.gc
```diff
@@ -174,75 +151,62 @@
...
+         (= (-> s5-0 particle-state s4-0) 2)
+         (= (-> s5-0 particle-state s4-0) 1)
+         )
+        (set! (-> arg2 vector 0 x) 245760.0)
+        )
+       ((= v1-0 -1)
+        (set! (-> arg2 vector 0 x) 0.0)
+        (set! (-> arg2 vector 0 y) 0.0)
         )
+       (else
```
water111, line 186: This `fuel-cell-progress-hud-orbit-callback` function has some changes so that it doesn't crash due to missing `manipy` stuff.  Can you replace this with the old version?

water111 (changes requested):
nice, that's a lot of particle effects!  A few comments on types

goal_src/engine/game/collectables-part.gc
xTVaser: i can double check, but not at this time i think -- I believe it complained about it not being fully defined.

decompiler/util/data_decompile.cpp
xTVaser: i figured id keep it, i feel like this will probably be correct in most cases -- but its rare enough that the warning shouldn't bother us, and in the event it is wrong somewhere, its more obvious

goal_src/engine/game/collectables-part.gc
xTVaser:
```
-- Type Error! --
Type eco-collectable is not fully defined.
Code:
((lambda :inline-only #t (v1-1)
 (let ((a0-3 (-> v1-1 root-override root-prim prim-core)))
  (set! (-> arg2 x) (-> a0-3 world-sphere x))
  (set! (-> arg2 y) (-> a0-3 world-sphere y))
  (set! (-> arg2 z) (-> a0-3 world-sphere z))
  )
 (if (zero? (logand (-> v1-1 state-flags) 1))
  (set! (-> arg1 next-time) (the-as uint (* (-> *sp-frame-time* x) 2)))
  )
 )
  (the-as eco-collectable (-> arg1 key proc))
  )
From macro: let
Used in:
goal_src/engine/game/collectables-part.gc, line: 89
```

water111 (approved):

PR #877, opened 2021-10-03
decomp: finish `battlecontroller`

xTVaser:


decompiler/config/all-types.gc
```diff
@@ -24636,20 +24639,20 @@
...
+   (enabled       symbol   :offset-assert 28)
    )
   :method-count-assert 9
   :size-assert         #x20
   :flag-assert         #x900000020
   )
 
 (deftype battlecontroller-creature-type (structure)
-  ((type2            basic  :offset-assert 0)
+  ((type2            nav-enemy  :offset-assert 0) ; a guess
```
water111, line 24655: probably a `type`

goal_src/levels/common/battlecontroller.gc
```diff
@@ -5,3 +5,1043 @@
...
+   (spawn-count                 int16                                    :offset-assert 198)
+   (die-count                   int16                                    :offset-assert 200)
+   (target-count                int8                                     :offset-assert 202)
+   (spawner-count               int8                                     :offset-assert 203)
+   (creature-type-count         int8                                     :offset-assert 204)
+   (spawner-array               battlecontroller-spawner       8 :inline :offset-assert 208)
+   (spawn-period                uint64                                   :offset-assert 464)
+   (path-spawn                  path-control                             :offset-assert 472)
+   (creature-type-array         battlecontroller-creature-type 4 :inline :offset-assert 476)
+   (final-pickup-type           int32                                    :offset-assert 604)
```
water111, line 51: this is probably a `pickup-type`

goal_src/levels/common/battlecontroller.gc
```diff
@@ -5,3 +5,1043 @@
...
+    (when
+     (and s4-0 (logtest? (-> (the-as nav-enemy s4-0) nav-enemy-flags) 2048))
+     (cond
+      ((< (-> s5-0 state) (-> s5-0 path curve num-cverts))
+       (when (or (-> self noticed-player) (= (-> s5-0 state) 1))
+        (let ((s3-0 (new 'stack-no-clear 'vector)))
+         (eval-path-curve-div!
+          (-> s5-0 path)
+          s3-0
+          (the float (-> s5-0 state))
```
water111, line 194: `eval-path-curve-div!` should probably take an `int` for this argument, but we don't have to worry about that now

goal_src/levels/common/battlecontroller.gc
```diff
@@ -5,3 +5,1043 @@
...
+  0
+  (none)
+  )
+
+(defbehavior battlecontroller-battle-end battlecontroller ()
+  (case (-> (level-get-target-inside *level*) name) 
+   (('citadel)
+     (level-hint-spawn
+      (game-text-id misty-battle-finished)
+      "sksp0378"
```
water111, line 499: do you know what this says?

goal_src/levels/common/battlecontroller.gc
```diff
@@ -5,3 +5,1043 @@
...
+    obj
+    arg0
+    )
+   )
+  )
+
+(defmethod deactivate battlecontroller ((obj battlecontroller))
+  (with-pp
+   (let ((gp-0 pp))
+    (let ((s6-1 obj))
```
water111, line 815: this is changing the process pointer, which is something we don't handle correctly in the decompiler yet.  can you leave a note at the top of the file about this? https://github.com/water111/jak-project/issues/882

goal_src/levels/common/battlecontroller.gc
xTVaser: "the doors open, lets get going"

goal_src/levels/common/battlecontroller.gc
xTVaser: hmm are you sure, it seems to explicitly convert to a float?
```asm
    mtc1 f0, s5               ;; [ 20] (set! f0-0 (gpr->fpr s5-0)) [s5: int ] -> []
    cvt.s.w f0, f0            ;; [ 21] (set! f0-1 (i2f f0-0)) [] -> []
    mfc1 a2, f0               ;; [ 22] (set! a2-0 (fpr->gpr f0-1)) [] -> [a2: float ]
```

goal_src/levels/common/battlecontroller.gc
water111: It's under `citadel` in the case statement - could this be for the room with all the annoying hopping enemies?

goal_src/levels/common/battlecontroller.gc
water111: oops, you're right - it should definitely stay the way it is. I read it as `(the-as float` instead.

Issue #889, opened 2021-10-08
[decompiler] say what object file is currently being worked on


ManDude:
Often times the decompiler will crash without warning but with no indication of the cause or even which file it crashed on. Makes it very difficult to debug quickly.

ManDude:
sticking a print in `for_each_obj` does the job but also spams the console ridiculously hard... need something more elegant

doctashay:
#890 will send this info to the decompiler.log file in a much more elegant fashion, if it is too verbose we can always log it to a separate file

ManDude:
fixed(?) in #895

PR #890, opened 2021-10-10
decompiler: show what object file is being worked on

doctashay:
This should address #889


decompiler/ObjectFile/ObjectFileDB.h
```diff
@@ -122,6 +122,7 @@ class ObjectFileDB {
     for (const auto& name : obj_file_order) {
       for (auto& obj : obj_files_by_name.at(name)) {
         f(obj);
+        lg::debug("Processing object {}", name);
```
ManDude, line 125: put it before the function call in case the function goes wrong

PR #895, opened 2021-10-12
[decompiler] clean up offline test program, reorganize decomp order

water111:
Still a work in progress, but `offline-test2` is working, at least on linux.  If it works on windows, I'm hoping to just get rid of the original.  The new version:
- uses the new order of passes for less RAM usage
- prints the file name, so you know where it crashes (and prints it just once)
- colored diffs without the crap in front
- json config file (`config.json` in the same file)

Still to-do before merge
- [x] compile
- [ ] exclude list for states (Vaser has this in a PR already, I just need to make sure the input's easily available for that)
- [x] verify that the normal decompiler still works
- [x] fix crash bugs and remove stuff from `banned_objects`


xTVaser:
Old system (debug):
- 7.5minutes before it even starts comparing files!
- over 18GB of RAM used total!
- around 13minutes to do the full suite.

New system (debug):
- way better output
- still works
- around 1GB of RAM
- Starts comparing files immediately
- total time is within the same ballpark

I imagine if we really cared, since we are doing file-by-file we could likely thread this?

PR #898, opened 2021-10-15
[goalc] improve reliability of debugger tests

water111:
the runtime will only flush its output buffer on the first kernel run after the first incoming message from the compiler (`ListenerStatus`).  I didn't realize this and assumed we'd get the `reset` message immediately on connection.  This PR waits for the response before continuing with the test.  I think this will fix the recent debugger test failures we've been seeing.


water111:
looks like that's not the only issue:
```
[2021-10-15 22:57:01:237] [debug] [DECI2] Waiting for listener...
[2021-10-15 22:57:01:288] [debug] [DECI2] Connected!
[Listener] Failed to connect
[DECI2] Got message: 24 0 0xe042 H -> E
[Listener] Socket connected established! (took 0 tries). Waiting for version...
Got version 0.8 OK!
[2021-10-15 22:57:01:374] [info] connected: true

[2021-10-15 22:57:01:374] [info] poking...

[Debugger] Context: valid = true, s7 = 0x147d24, base = 0x7efca45a2000, tid = 5810

[2021-10-15 22:57:01:375] [info] break...

goalc-test: /home/runner/work/jak-project/jak-project/goalc/debugger/Debugger.cpp:439: void Debugger::read_symbol_table(): Assertion `is_valid() && is_attached() && is_halted()' failed.
```

water111:
First issue: failing an `assert` could cause the test to exit without flushing stdout, so I missed print statements.  I don't really know how this is possible because I thought stdout would be flushed with every newline.

Second issue: The "watcher" thread runs in the background and checks to see if the process segfaults, disappears, etc, then prints a message to user if it has.  I was starting this too early, and it was racing `ptrace(PTRACE_SEIZE`.  If the watcher got to the `waitpid` before the ptrace finished, it would look like the process died and the debugger state would be reset.  Then, the debugger initialization would go to read the symbol table and the context would be invalid.

PR #958, opened 2021-11-10
[graphics] partial tfrag implementation

water111:
known missing features:
- mipmap (want to do differently on PC port)
- fog (want to do differently on PC port)
- ~~the bridge in rock village is transparent but shouldn't be (?)~~
- performance is not good (too many draw calls due to texture changes)
- visibility information is not being computed in many places, you have to fake it (not yet implemented)
- haven't done tfrag near
- haven't done many of the other tfrag subroutines
- tfrag transparent/low-res/etc aren't fully done

The good news
- It works! https://www.youtube.com/watch?v=h01W1wzKxIY
- The time-of-day color stuff works! I was a little worried that this wouldn't work, or we were missing something.
- The texture cache stuff all worked and we had the right textures with the right formats associated with the right TBP's.  After the level is loaded, we never have to reconvert textures.  I don't think this texture cache approach is what we want long term for performance reasons, but it's at least a known working thing to work off of.
- The terrible C++ rewrite of VIF UNPACK and the VU program is pretty fast. I think with a low amount of effort we could get at least 2x faster, and a _lot_ faster if we really wanted to and moved more to shaders
- There's nothing sneaky going on with the drawing setup. The "direct renderer" is inefficient but can handle this just fine.


water111:
In testing, we found that VI1.DGO no longer fits in the level heap (about 1k too big).
Also, a scratchpad dma assert was triggered in the drawing function (mips2c) where sadr was larger than 0x4000. I suspect we may need more than 2k alignment on it and I was getting lucky.

PR #959, opened 2021-11-11
improve debugger disasm, `:sym-name` and fix Windows builds

ManDude:


ManDude:
Codacy is not aware that `uintptr_t` on Windows is `unsigned long long` and thus `llx` would be appropriate...

CMakePresets.json
```diff
@@ -21,7 +21,7 @@
       "displayName": "Windows x64 Debug",
       "description": "Target Windows with the Visual Studio development environment.",
       "generator": "Ninja",
-      "binaryDir": "${sourceDir}/out/build/${presetName}",
+      "binaryDir": "${sourceDir}/out/build/Debug",
```
xTVaser, line 24: better to just change the "name" above to `Debug` or `Release` below instead of hard-coding it.  I've made this change in a few of my PRs as well.

PR #964, opened 2021-11-13
[decomp] entity birth

ManDude:
It works! Assuming the actual thing it tries to spawn is functional. Actor compaction does not work.

image
image
image

Rendering errors are because of the `sprite` renderer.


ManDude:
fix #966 and #917

PR #965, opened 2021-11-13
game: Get camera code working

xTVaser:
Have some issues around switching the camera mode, but the default camera stuff is working
Added support for analog sticks as well

https://user-images.githubusercontent.com/13153231/141607735-a7f04221-fb22-403a-ac33-554a61ff7e30.mp4


common/goos/Interpreter.cpp
```diff
@@ -1643,4 +1644,20 @@ Object Interpreter::eval_string_to_symbol(const Object& form,
...
+
+Object Interpreter::eval_get_env(const Object& form,
+                                 Arguments& args,
+                                 const std::shared_ptr<EnvironmentObject>&) {
+  vararg_check(form, args, {ObjectType::STRING}, {{"default", {false, ObjectType::STRING}}});
+  const std::string var_name = args.unnamed.at(0).as_string()->data;
+  const char* env_p = std::getenv(var_name.c_str());
+  if (env_p == NULL) {
+    if (args.has_named("default")) {
+      return StringObject::make_new(args.get_named("default").as_string()->data);
```
water111, line 1656: I think this can just be `return args.get_named("default");`

game/graphics/pipelines/opengl.cpp
```diff
@@ -10,6 +10,8 @@
 #include "third-party/imgui/imgui.h"
 #include "third-party/imgui/imgui_impl_glfw.h"
 #include "third-party/imgui/imgui_impl_opengl3.h"
+#define STB_IMAGE_IMPLEMENTATION
+#include "third-party/stb_image.h"
```
water111, line 14: thanks for adding this library, this is a useful one.  I think we're going to want it in more places (for loading/saving texture files, screenshots, etc).  Can you move the implementation to `common`?  I think you would just remove the `#define STB_IMAGE_IMPLEMENTATION` from here, and then make a new file in `common` that just contains:
```
#include "third-party/stb_image.h"
#define STB_IMAGE_IMPLEMENTATION
```

goal_src/engine/debug/default-menu.gc
```diff
@@ -21,12 +21,16 @@
 
 (define *dm-cam-mode-interpolation* (the-as debug-menu 0))
 
-(defun dm-cam-mode-func ((arg0 uint) (arg1 int))
+(defun dm-cam-mode-func ((arg0 (state camera-slave)) (arg1 int))
   (if (and (= arg1 4) arg0)
-   (send-event *camera* 'change-state arg0 *dm-cam-mode-interpolation*)
+   (begin
+    (format 0 "DEBUG <Vaser>: Changing State Arg 0 '~P' Arg 1 '~P'~%" arg0 arg1)
```
water111, line 27: probably okay to remove these now?  Also, you can use `(when ..` instead of `(if (begin ..` to save some typing if you want

goal_src/engine/debug/default-menu.gc
```diff
@@ -2933,144 +2939,6 @@
          (stub-item (new 'debug 'debug-menu-item-submenu "Instance" stub-menu)))
     stub-item
     )
-  ;; TODO - fix me!
```
water111, line 2936: did you mean to remove this?

goal_src/kernel/gkernel-h.gc
```diff
@@ -34,11 +34,11 @@
 
 ;; the size of the execution stack (~14 kB) shared by all threads
 ;; OpenGOAL NOTE: increased to 32kB
-(defconstant DPROCESS_STACK_SIZE #x8000)
+(defconstant DPROCESS_STACK_SIZE #x10000)
```
water111, line 37: update comment to say 64 kB

water111 (commented):
Nice that the camera is working! Soon we can get rid of my hacked-together camera script.

goal_src/kernel/gkernel-h.gc
xTVaser: i walked these back, since they weren't actually required

PR #977, opened 2021-11-18
[goos/goal] user profiles

ManDude:
Implements "user profiles" to GOOS/GOAL, allowing you to have extra custom code loaded at start-up that isnt part of the standard library. Also can be used for conditional compilation to tune settings for specific users.

A username can be selected with the `-user <username>` parameter in `gc`. The `-user-auto` parameter can also be used, and it reads the username from the file `goal_src/user/user.txt` (it reads the first uncommented valid word as the username).

There's a GOOS `user?` macro that will return `#t` if the current user matches any of its arguments. For example, `(user? mark john)` will return `#t` if the current user is either `mark` or `john`, and `#f` otherwise. You can use this for conditional compilation.

There is a `*user*` variable in both GOAL and GOOS which holds the username. Do not write to it, it won't magically reload user profiles and will just mess with the `user?` macro.

The default profile is `#f` and does not load any user code. It should be used for proper shareable builds, obviously.

### DIFF
--- a/common/goos/Interpreter.cpp
+++ b/common/goos/Interpreter.cpp
@@ -26,10 +26,10 @@ Interpreter::Interpreter(const std::string& username) {
   define_var_in_env(goal_env, goal_env, "*goal-env*");
   define_var_in_env(goal_env, global_environment, "*global-env*");
 
-  // set user profile bool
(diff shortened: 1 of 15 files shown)


ManDude:
test failure is some strange codacy warning that probably means absolutely nothing

water111:
yeah codacy goes insane if you name a function `read`

PR #978, opened 2021-11-22
[graphics] tfrag3 renderer

water111:
There are a few things to take care of before we can merge this:
- make the extractor print an error if you don't have the DGO
- handle the final boss missing texture somehow
- handle the connection from the game to the renderer - right now the levels/time of day are not sent to the renderer.
- make a nicer "debug" mode that lets you turn on many levels at once, or override time of day from the game
- general clean up of the extractor

Known issues:
- the loading causes a short pause.
- some alpha blending modes are not supported.
- ice and trans do not appear to render at all
- the texture wrapping in boggy swamp _might_ be wrong
- there is a bright white thing in fj temple (might just be the alpha not handled right)
- there is an issue with texture coordinate clamping - we may need separate modes for clamp u/clamp v.

Future to-do:
- frustum culling (the bspheres are in the format already and the index data is arranged so this is easy to check)
- vertex deduplication
- some more checks of the z buffer scaling/accuracy. The stuff super far in the distance is close to being clipped and there's some fighting on the weird mountain things in finalboss.


water111:
texture clamp bug:
image
missing transparent:
image
boggy swamp:
image


In good news the texture fighting on some things is much better. For example, the chain in rock village is perfect now (this is bad even in PCSX2):
image

Also the mysterious texture corruptions, weird colors when looking up, and missing tfrags from previous renderers are completely gone.

water111:
tfrag3 with sprites:
image

water111:
The texture clamp issue is just that we need separate u/v clamps. Adding this increased draw calls by ~15%.  I _think_ if we really needed, we could manually inspect textures and pick clamps to reduce the total number of draw types. But probably not worth it for now, when we are under 100 draw calls/level.

The transparent stuff is luckily not an issue with extraction. The blue mesh is the transparent tfrag tree.
image

PR #984, opened 2021-11-24
[decomp] make use of the generated load boundary data

ManDude:


goal_src/goal-lib.gc
```diff
@@ -332,6 +332,28 @@
...
+  (let ((val (first bindings))
+        (arr (second bindings)))
+
+        `(let* ((,len (-> ,arr length))
+                (,i 0)
+                (,val (-> ,arr ,i)))
+             (while (< ,i ,len)
+                ,@body
+                (1+! ,i)
+                (set! ,val (-> ,arr ,i))
```
water111, line 348: I think this will read past the end of the array on the last iteration through. I think you'll need to put a `(let ((,val (-> ,arr ,i)))` inside of the `while` to get `val` with the right type, and not read past the end.

water111 (changes requested):
looks good - just one comment on the `doarray` macro

goal_src/goal-lib.gc
ManDude: This is the original code:
```lisp
(let* ((gp-0 (-> *static-load-boundary-list* length))
      (s5-0 0)
      (a0-1 (-> *static-load-boundary-list* s5-0))
      )
  (while (< s5-0 gp-0)
   (load-boundary-from-template (the-as (array object) a0-1))
   (+! s5-0 1)
   (set! a0-1 (-> *static-load-boundary-list* s5-0))
   )
  )```
`val` will be out of bounds after the last iteration, but that value won't ever be used because the `while` condition will fail anyway.

water111: ah ok, in that case let's do the same thing they did. Agreed that it won't cause problems. Unless you have an array that ends at the end of memory, but they never do this.

PR #986, opened 2021-11-25
[decompiler] detect `cpad-hold?` and `cpad-pressed?`

ManDude:


decompiler/IR2/FormExpressionAnalysis.cpp
```diff
@@ -3808,6 +3808,64 @@ FormElement* try_make_nonzero_logtest(Form* in, FormPool& pool) {
...
+              ->at(0));
+      auto buttons_form = jank->elts().at(1);
+      std::vector<Form*> v = {
+          pool.form<SimpleAtomElement>(SimpleAtom::make_int_constant(mr.maps.ints.at(0)))};
+      int i = 2;
+      buttons_form->apply_form([&](Form* form) {
+        if (i-- <= 0) {
+          v.push_back(form);
+        }
+      });
```
water111, line 3859: I think this whole thing can be:
```cpp
auto buttons_form = dynamic_cast<GenericElement*>(in->at(0));
v.insert(v.end(), buttons_form.elts().begin(), buttons_form.elts_end());
```
the `apply_form` functions iterate over the entire tree, and we probably don't need that here - we just want to grab the enum names for `pad-buttons`

water111 (changes requested):
nice, this looks much nicer. There might be an easier to grab the buttons from the `pad-buttons` form that will make sure that the parent `Form*`/`FormElement*`s are all set correctly.

decompiler/IR2/FormExpressionAnalysis.cpp
ManDude: It couldn't be that simple because that would return the form elements and not the forms themselves needed to make a new `GenericElement`, but this looks cleaner anyway.

Issue #995, opened 2021-12-05
[debugger] be less aggressive with backtracing


ManDude:
it can technically crawl through the entire 32-bit address space before it gives up. This sometimes happens when connecting, and you have to restart everything.

ManDude:
Also probably just do it differently. Maybe search for the value `function` for a bit, or something.

Issue #996, opened 2021-12-05
[game] game hangs when changing levels sometimes


ManDude:
I still haven't figured out what keeps causing this. It doesn't seem consistent enough and I have only gotten it with a release build.

```
GC vil1-sky-00
GC vil1-sky-01
GC vil1-sky-02
GC vil1-sky-03
GC vil1-sky-04
GC vil1-sky-05
GC vil1-sky-06
GC vil1-sky-07
```

ManDude:
might be some corruption caused by `STR` files

PR #997, opened 2021-12-05
[decomp] cleanup `default-menu`, fix dangerous IOP bug, autoconvert ints in pairs to/from bintegers

ManDude:


ManDude:
fixes #995 and #996

PR #1003, opened 2021-12-11
[decomp] `joint` and related

ManDude:
The rewritten asm funcs will need to be thoroughly double-checked...


ManDude:
doesn't actually make anything work in-game, but fixes a few types and bugs in the decompiler. This stuff is extremely complex.

PR #1016, opened 2021-12-14
decomp: finish `vol` | `helix-water` | `minecart`

xTVaser:


goal_src/engine/geometry/vol.gc
```diff
@@ -5,3 +5,272 @@
...
+    )
+  #t
+  )
+
+(defmethod dummy-9 vol-control ((obj vol-control))
+  (with-pp
+    (let ((a0-1 obj))
+      (when (and (and *display-vol-marks* (logtest? (-> a0-1 flags) 1)) (logtest? (-> obj flags) 2))
+        (when (zero? (-> obj debug-point))
+          (let ((s5-0 pp))
```
water111, line 223: they change the process pointer (not handled correctly in the decompiler), but it looks like it has no effect, so we can ignore it.

goal_src/engine/geometry/vol.gc
```diff
@@ -5,3 +5,272 @@
...
+            (let ((s6-1 (-> obj process)))
+              )
+            (let ((s4-0 0))
+              (dotimes (v1-8 (-> obj pos-vol-count))
+                (+! s4-0 (-> obj pos-vol v1-8 num-planes))
+                )
+              (dotimes (v1-11 (-> obj neg-vol-count))
+                (+! s4-0 (-> obj neg-vol v1-11 num-planes))
+                )
+              (set! (-> obj debug-point) (new 'debug 'vector-array (* 10 s4-0)))
```
water111, line 233: probably this used to be `new 'process`, and that's why they changed the process pointer.  But it doesn't matter - allocating on the debug heap doesn't check the current process in any way...

goal_src/levels/sunken/helix-water.gc
```diff
@@ -5,3 +5,782 @@
...
+  :flag-assert         #x14004000b0
+  (:states
+    helix-slide-door-close
+    helix-slide-door-idle-closed
+    helix-slide-door-idle-open
+    )
+  )
+
+;; Manually added!
+(define-perm *helix-slide-door* helix-slide-door #f)
```
water111, line 25: I think we should make these `(define-extern *helix-slide-door* helix-slide-door)`. The `define-perm` version will set the value to `#f`, which didn't happen in the original game.

goal_src/levels/sunken/helix-water.gc
```diff
@@ -5,3 +5,782 @@
...
+    helix-button-activate
+    helix-button-idle-down
+    helix-button-idle-up
+    helix-button-quick-activate
+    helix-button-startup
+    )
+  )
+
+;; Manually added!
+(define-perm *helix-button* helix-button #f)
```
water111, line 57: same here

goal_src/levels/sunken/helix-water.gc
```diff
@@ -5,3 +5,782 @@
...
+    (dummy-21 (_type_) object 21)
+    )
+  (:states
+    helix-water-activated
+    helix-water-idle
+    )
+  )
+
+;; Manually added!
+(define-perm *helix-water* helix-water #f)
```
water111, line 99: and here

water111 (changes requested):
just one tiny change

PR #1023, opened 2021-12-16
decomp: finish `rolling-obs` | `hud` | `title-obs` | `maincave-obs` | `lavatube-energy`

xTVaser:


goal_src/levels/lavatube/lavatube-energy.gc
```diff
@@ -5,3 +5,1946 @@
...
+    (if (nonzero? (-> self sound))
+        (stop! (-> self sound))
+        )
+    (none)
+    )
+  :trans
+  (behavior ()
+    (update! (-> self rotation-speed) 0.0)
+    (energyhub-trans)
+    (let ((t9-2 energyhub-set-lava-height))
```
water111, line 1600: I think this function takes a float argument

goal_src/levels/rolling/rolling-obs.gc
```diff
@@ -5,3 +5,1779 @@
...
+  :trans
+  (the-as (function none :behavior pusher) rider-trans)
+  :code
+  (behavior ()
+    (while #t
+      (let ((f0-0 -1.0))
+        (when (and *target* *camera*)
+          (let ((gp-0 (new 'stack-no-clear 'vector)))
+            (vector-! gp-0 (target-pos 0) (camera-pos))
+            (set! f0-0 (ray-capsule-intersect (-> self cyl) (camera-pos) gp-0))
```
water111, line 124: I think this checks to see if the camera is inside the pusher, and then retracts the pusher out of the way of the camera. Kind of cool

water111 (changes requested):

goal_src/levels/rolling/rolling-obs.gc
xTVaser: I wonder if you were able to flip the camera around to face jak -- if they'd never get in your way.

goal_src/levels/lavatube/lavatube-energy.gc
xTVaser: ya you're right, but its unused -- weird.

PR #1026, opened 2021-12-19
[graphics] TIE extractor

water111:
This is just the extractor and some tie decomp. I want to clean up and merge the extractor/decomp before starting on the PC port renderer.


water111:
Known issues:
- some unhandled blend modes
- wind is not implemented at all
- some colors are wrong I suspect interpolated points have it wrong?
- something is wrong with the window frame on the big hut
- no culling implemented (performance is okay but not great)
- no connection to engine
- files are large enough that the loads are a few frames (probably not for this PR to fix though)

PR #1032, opened 2021-12-27
[runtime] `pckernel` implementation

ManDude:


ManDude:
Done. Works, although many of the settings curreently don't do anything. Right now it can save and load the settings file. Game resolution, aspect ratio, and fullscreen status are applied properly on load.

ManDude:
Also:
- allows using goal macros inside pairs
- fixes game aspect ratio. Previously it was rendering as if it were 448px tall, but our framebuffer is made for 512px. The excess region is rendered in the original game, but scissored off.
- adds a `(:stop)` command to the debugger to detach.
- windows debugger doesnt kill the target on detach. detach also works now.
- fix debug-only files not being dumped as debug-only by the decompiler.
- adds a toggle in decompiler settings to dump .obj files for levels or not.
- fix write mode for `file-stream`.
- adds rolling, swamp, sunken b, jungle b, snow, ogre and lavatube levels.
- added hack to fix debug text in widescreen.
- adds stuff to the debug menu.
- adds a `bit-field` goal macro to arbitrarily extract bitfields from values.
- makes `string->float` actually work.

game/graphics/opengl_renderer/shaders/tfrag3.vert
```diff
@@ -65,7 +65,7 @@ void main() {
     gl_Position = transformed;
 
     // time of day lookup
-    fragment_color = texture(tex_T1, time_of_day_index / 8192.f);
+    fragment_color = texelFetch(tex_T1, time_of_day_index, 0);
```
water111, line 68: ah, this `texelFetch` was what I couldn't find when I tried to do this

goal_src/engine/game/video.gc
```diff
@@ -5,20 +5,29 @@
...
      (set! (-> *video-parms* relative-y-scale) 1.0)
-     (set! *ticks-per-frame* 9765)
+     (set! *ticks-per-frame* 9765) ;; 585900 total?
      (set! (-> *math-camera* isometric vector 1 y) 0.5)
-     (set! (-> *math-camera* y-pix) 112.0)
-     (set! (-> *math-camera* y-clip) 448.0)
+     (#cond
+       (PC_PORT
+        ;; discard scissoring area
+        (set! (-> *math-camera* y-pix) 128.0)
```
water111, line 23: this _might_ cause issues with the progress menu, but probably not worth worrying about until we get it working again.

goal_src/engine/pc/pckernel.gc
```diff
@@ -0,0 +1,579 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+
+(define *pc-settings* (new 'global 'pc-settings))
+
+
+
+)
+
+
```
water111, line 578: awesome, really cool to see new OpenGOAL code being written!

goalc/compiler/compilation/Static.cpp
```diff
@@ -557,16 +557,26 @@ Val* Compiler::compile_bitfield_definition(const goos::Object& form,
...
                                             form.as_pair()->car.as_symbol()->name == "the" ||
                                             form.as_pair()->car.as_symbol()->name == "lambda")) {
       return compile_static(form, env);
     }
-    auto car = compile_static_no_eval_for_pairs(form.as_pair()->car, env, seg);
-    auto cdr = compile_static_no_eval_for_pairs(form.as_pair()->cdr, env, seg);
+    // try as a macro
+    if (can_macro && form.as_pair()->car.is_symbol()) {
+      goos::Object macro_obj;
+      if (try_getting_macro_from_goos(form.as_pair()->car, &macro_obj)) {
```
water111, line 572: I think this is fine, we just have to watch out that no macro names conflict with things we actually want in static pairs.

ManDude: you're right, `send-event` comes to mind, I don't think there's any examples of a pair starting with `send-event` thankfully. I think one change we could make later is have it so `'(send-event 1 2 3)` is a pair but `',(send-event ...)` or `'(foo bar ,(send-event ...))` is recognized as a macro. We'll see.

PR #1041, opened 2021-12-30
improve performance on intel graphics

water111:
And split out the sky blend into a separate file, so we can more easily have add a CPU version later, if needed.


water111:
On the integrated graphics in my laptop (intel ?) this decreases frame time from 33ms to 6ms for the "jungle sprite test". It also seems to improve things for dass (has AMD GPU). There is a very small improvement on NVIDIA for me.

List of changes:
- fix a bug with listener sending large files, like `default-menu.gc` causing a 16-bit integer to overflow
- confirm that texture `q` is 1 for all tie/tfrag and avoid the divide in the fragment shader.
- do blending of the sky on the CPU by default (optionally can go back to old GPU method).  The GPU based blending used a bunch of weird opengl stuff that might be slow.
- properly use vertex array object for direct renderer
- use a single attribute buffer for direct renderer instead of separate position, texture, rgba
- use floats for direct renderer vertex positions (uint32 has issues on amd)
- use a different part of the attribute buffer for different draws to avoid having to synchronize on each draw. Some drivers might be able to figure out how to avoid this, but intel on linux doesn't. Done in direct, tie, and tfrag.
- increase vertex buffer sizes in a few places
- use separate textures/buffers per tree in tfrag/tie to avoid synchronizing on each tree switch due to overlap.

PR #1047, opened 2022-01-03
WIP: `shrubbery`

xTVaser:
# unknowns
- `(method 8 drawable-tree-instance-shrub)`
  - has some really weird method error ive never seen before

# login methods

- `shrubbery`'s `login` calls `adgif-shader-login-no-remap` on each texture in the shrub, the number of textures is defined in the shrubs header
- `generic-shrub-fragment`'s `login` calls `adgif-shader-login-no-remap` on each texture, defined by (`cnt-qwc` / `5`)

- `drawable-tree-instance-shrub`'s login method is just a call to `prototype-inline-array-shrub`'s if it exists
  - `prototype-inline-array-shrub` calls `login` on each `geometry` object in the `prototype-bucket`
- `prototype-generic-shrub`'s `login` calls `login` for each drawable object -- this is essentially a typedef of `drawable-group`

- `prototype-shrubbery`'s `login` calls `login` for each `shrubbery`

- `billboard`'s `login` calls `adgif-shader-login` for the `adgif-shader` stored within

# DMA data generation

- `draw-inline-array-instance-shrub`
  - skipped
- `draw-prototype-inline-array-shrub`
  - iterates through each item in the `(inline-array prototype-bucket-shrub)`
  - dma-buffer's are coming out of `display`'s `display-frame` -- offet 584 (vu1-buf?)
    - builds up DMA data from `offset 4720` of instance-shrub-work (`count-tmpl`)
      - iterates through each uint128 (in pairs) (made a `vector4w`, a guess)
      - stores it into the dma chain, and uses the first 64bits of it along with a constant int64 (`0xffffffff | 0x80000000`) L141
      - bitshift operations ensue (and > shifts > or)
      - then it's stored back into the dma chain at the same position
      - then stores a pointer to this object into `next-clear-2` and `next-clear-3` respectively in the current `prototype-bucket`
  - dma-buffer loads a value into `s4` used later as a raw int it seems, seemingly the same value stored into `s3` as well TODO
    - this dma-buffer is then passed to `generic-init-buf`
    - stores some constants - `dma-packet`
      - `#x10000002 OR 2` (int64)
      - `0` (int32)
      - `#x50000002 OR 2` (int32)
    - constructs a gif tag with the following constants
      - L143
      - L142
      - `#x5026b OR 619` set on the gs test_1 register (71)
  - next dma buffer, stores 3 qwords - made it a `matrix3`
    - these values are related to `chainb` in `shrub-work`
    - does a similar operation to that mentioned above, grabs qword 96 as an int64, perform an `and shift (with near-next from shrub-work) and or` and store back into `shrub-work`
    - this modified value goes first into the dma-chain (uint128)
    - and then the 2 next (112 and 128)
    - after moving past this matrix, the offset is stored in `scrub-work::near-last(176)` at offset `4`
      - TODO - what types are stored in near-next/last/etc - `shrub-near-packet` perhaps, the size is too small though.
        - perhaps it refers to `chainb` in the shrub-work itself
  - more constants - likely a `dma-packet`
    - `#x20000000` (int64)
    - 0 (int32)
    - 0 (int32)
  - DO THE FOLLOWING TWICE (i assume because the game builds two frames at a time?)
    - next we pass the dma-buffer to `shrub-init-frame`
    - then to `shrub-upload-model`
    - add a dma-packet to it
    - and another dma-packet
    - the `dma-buffer` is then passed to `dma-bucket-insert-tag`
    - `s4` from above finally comes into play, pull a value out of the dma-buffer and subtract it -- looks like a memory usage calculation
      - repeat the trend, store current offset into `s4` again from the dma-buffer
  - 2x dma-packets
  - enumerate `arg1` defined by `arg0` as the counter
  - one dma-packet
  - calculate the `s4` memory usage again
- `shrub-upload-test`
  - unused, its in the name, i believe this was just a temporary test
- `shrub-upload-model`
  - sets some `texture-giftag`s, maybe interesting?
- `shrub-init-frame`
  - adds a dma-packet
  - adds a gs-gif-tag
  - the arg is passed into GS' test_1 register (71)
- `shrub-do-init-frame`
  - adds vu-function `shrub-vu1-block` to dma-buffer
  - creates a dma-packet
  - adds a MAYBE dma-packet? not sure
  - not sure what this final thing added to the dma-buffer is
    - it just seems like a bunch of constants, perhaps colors? not sure
  - sets `*shrub-state*` to `2`
- `shrub-upload-view-data`
  - adds some sort of `dma-packet`
  - then calls `shrub-init-view-data` with the same thing it just added to the `dma-buffer`
    - however, this should be a `shrub-view-data` (see below) but it doesn't decompile very nicely...not sure 
- `shrub-init-view-data`
  - I'm convinced this takes a `shrub-view-data` as it's argument, it matches too perfectly


decompiler/level_extractor/extract_common.h
```diff
@@ -0,0 +1,66 @@
...
+
+namespace decompiler {
+
+/// <summary>
+/// Get the index of the first draw node in an array. Works for node or tfrag.
+/// </summary>
+/// <param name="array"></param>
+/// <returns></returns>
+u16 get_first_idx(const level_tools::DrawableInlineArray* array) {
+  auto as_tie_instances = dynamic_cast<const level_tools::DrawableInlineArrayInstanceTie*>(array);
```
xTVaser, line 16: i now notice why these were not extracted, reminder for myself to revert

PR #1048, opened 2022-01-03
[files] merge fake duplicate files

ManDude:


ManDude:
Code files will be done some other time.

water111:
This is great, so much simpler!

> Code files will be done some other time.
I think some of the code files do have actual differences. L1 seems to be an older version of some types.  That said, maybe we should make the non-L1 version have a normal name and the others have the `-L1` suffix.

PR #1051, opened 2022-01-03
docs: Refresh the homepage and add some basic gallery support

xTVaser:
Took some time to refresh the docs website a bit

image
Livens up the homepage quite a bit, displays more information in a cleaner way

Also adds some basic gallery support so we can start archiving the progress outside of discord, it's kinda rough:
- we'll want images to fullscreen when clicked
- support videos
- sort by date properly
- we'll probably want to compress gallery images
- pagination so not everything is loaded at once
- ... but this can all be iterated on

Drop images into `/docs/gh-pages-proj/assets/gallery/[jak1|jak2|jak3|jakx|misc]` and the build system will pick them up, replacing `-`s with spaces to use as a caption from the filename.

image


water111:
fancy, looks great!

PR #1055, opened 2022-01-07
[runtime] Fix progress menu aspect ratios

ManDude:


ManDude:
Fixing the crappy progress menu is turning out to be way too complicated for now, we can fix it more later. It looks a *little* better now, at least.
image

Also uses a better scissoring fix, the aspect of everything should be correct now.

PR #1069, opened 2022-01-09
[game] revise build order and add remaining levels + add `part-tester` to menu

ManDude:


decompiler/config/all-types.gc
```diff
@@ -22014,11 +22012,12 @@
 
 ;; - Unknowns
 
-(define-extern *viewer* viewer) ;; unknown type
+(define-extern *viewer* viewer)
 (define-extern viewer-ja-name string)
 (define-extern viewer-geo-name string)
 (define-extern *viewer-sg* skeleton-group)
 (define-extern viewer-string string)
+(define-extern viewew-process (state viewer))
```
xTVaser, line 22020: why isn't this able to be defined in the type?

decompiler/config/all-types.gc
```diff
@@ -22037,8 +22036,6 @@
   :size-assert         #x7c
   :heap-base           #x100
   :flag-assert         #xe0100007c
-  (:states
-    part-tester-idle)
```
xTVaser, line 22041: same here

ManDude: it's not useful, the state is defined before use

PR #1075, opened 2022-01-11
[runtime] GPU sprite renderer

ManDude:
Replaces the old `SpriteRenderer` using the `DirectRenderer` with a new version that does work on the GPU instead of using the CPU and building giftags. Performance is better/more consistent, but blend mode changes are still the largest bottleneck. There's probably more ways to improve this renderer anyway.


water111 (approved):
Sorry, I didn't realize you finished this already!

PR #1076, opened 2022-01-15
[collision] implement most of background collision, spawn target!

water111:
This adds collide-cache, collide-shape, and some random other functions needed to get `*target*` to spawn.

There are a bunch of known issues:
- Edge grab is not decompiled (it's a separate `collide-edge-grab`)
- Foreground collision is not super well tested. Foreground sphere collision seems to work at least sometimes (you can pick up orbs/eco, but the blue eco "magnet" only kinda works), but foreground mesh is completely untested.
- Anything relying on bones or mesh primitives that are bound to a skeleton doesn't work. There's a workaround for `*target*` because punch puts a collision sphere at jak's hand, and we don't know where his hand is yet (it ends up at NaN right now). The blue eco effects all spawn from Jak's foot as a result, and all foreground mesh collision won't work.
- The joint animation stuff doesn't work correctly so there are many warnings about too many animations playing.
- ~~Some yellow eco projectiles and blue ecos aren't visible (for example, blue eco in LPC tube game room).~~ (fog limits applied to `w` used to undo perspective divide, caused sprite positions to multiply by zero)
- Getting a death or powercell animation will crash things.
- Interaction of HUD and pause menu seems wrong (you can have hud and pause menu opened at the same time).
- Changing the level using the checkpoint feature can result in a corrupted sky texture (this is a C++ renderer issue I know about related to texture conversion caching, will be fixed for real when we switch to entirely preconverted textures).


xTVaser (approved):
no obvious issues for me on windows, was able to do a lot of things as "jak"

PR #1103, opened 2022-01-21
test game actors and fix up/disable broken ones

ManDude:


goal_src/engine/target/target-handler.gc
```diff
@@ -298,19 +296,13 @@
              )
          )
         ((= v1-0 'change-state)
-         enter-state
-         (-> arg3 param 1)
-         (-> arg3 param 2)
-         (-> arg3 param 3)
-         (-> arg3 param 4)
-         (go (the-as (state target) (-> arg3 param 0)))
+         (go (the-as (state object object object object target) (-> arg3 param 0)) (-> arg3 param 1) (-> arg3 param 2) (-> arg3 param 3) (-> arg3 param 4))
```
water111, line 299: good catch! I am surprised this didn't cause more problems.

water111 (approved):

PR #1110, opened 2022-01-26
[bones] Part 1: Joint

water111:
This PR adds the decompression functions to `joint` and has process-drawable call them.  This has the terrible jump table functions and also a few that don't use normal calling convention. I patched mips2c functions to call each other directly, and turned the jump table functions into a switch statement. The setup jump table function was rewritten in GOAL to put the right constants in memory. 

At this point, it's hard to know if it's working or not. As far as I can tell, the positions that come out are reasonable and the intro cutscene moves the camera in the right way. There are issues with animations because the animation synchronizes to the audio, which is not working (it probably returns frame -1 the whole time)  The orientations probably come out quaternions somewhere and need to be converted to matrices by bones, and we don't have bones yet.


water111:
##  Bone use
- `real-main-draw-hook` calls `foreground-engine-execute`. This calls `bones-init`, then (through the engine system) calls `add-process-drawable`, which calls the `dma-add-func`, which is `dma-add-process-drawable` for most, which calls `draw-bones`. Then `bones-wrapup`.
- After that, there's a call to `bones-mtx-calc-execute`.
- The `math-engine-execute` uses the results of bones.

PR #1116, opened 2022-01-28
[Overlord] WIP decomp and libsd impl

Hat-Kid:
This is a WIP decomp of some Overlord functions related to sound/VAG stuff and basic implementation of libsd mostly from ps2sdk's freesd.

Still unsure what a lot of stuff does, but I hope it can help with making a tiny bit of progress towards sound.


game/overlord/iso.cpp
```diff
@@ -793,6 +823,120 @@ u32 ProcessVAGData(IsoMessage* _cmd, IsoBufferHeader* buffer_header) {
...
 
+u32 GetPlayPos();
+void UpdatePlayPos();
+void StopVAG(VagCommand* cmd);
+void PauseVAG(VagCommand* cmd);
+void UnpauseVAG(VagCommand* cmd);
+void CalculateVAGVolumes(u32 a1, u32 a2, u32 a3, int* a4, int* a5);
+void SetVAGVol();
+
+u32 GetPlayPos() {
```
water111, line 834: This function is one that we'll probably be interested in soon. When playing cutscenes, the animation player tries to stay in sync with the audio.  This function gets called once per frame on the IOP, and the result gets sent back to the EE.  Unfortunately it uses the `sceSd` library for some stuff, so it's not super simple.

That said, I'd guess that this returns the position in the stream in audio samples.

game/overlord/iso.cpp
```diff
@@ -793,6 +823,120 @@ u32 ProcessVAGData(IsoMessage* _cmd, IsoBufferHeader* buffer_header) {
...
+  gFakeVAGClockPaused = 0;
+}
+
+// TODO - not sure what any of these are yet
+void CalculateVAGVolumes(u32 a1, u32 a2, u32 a3, int* a4, int* a5) {
+  if (a2) {
+    u32 falloff_vol = 0;  // CalculateFalloffVolume(a3, (a1 * gDialogVolume) >> 10, 1, 10, 50);
+    short angle = 0;      // CalculateAngle(a3);
+    *a4 = (angle * falloff_vol) >> 10;
+    *a5 = (angle * falloff_vol) >> 10;
```
water111, line 913: my guess is this is calculating the volume for the left/right audio channels, based on the position of the camera/cutscene audio sorce.

game/overlord/iso.cpp
```diff
@@ -793,6 +823,120 @@ u32 ProcessVAGData(IsoMessage* _cmd, IsoBufferHeader* buffer_header) {
...
+    if (*a4 >= 0x4000) {
+      *a4 = 0x3FFF;
+    }
+    if (*a5 >= 0x4000) {
+      *a5 = 0x3FFF;
+    }
+  } else {
+    u32 v1 = (a1 * gDialogVolume) >> 6;
```
water111, line 921: and this case is if you have no position data, just put left and right at the same volume.

water111 (commented):
Looks good so far!

game/overlord/iso.cpp
```diff
@@ -774,14 +806,147 @@ u32 bswap(u32 in) {
...
+    gSampleRate = _cmd->ready_for_data;
+    gLastVAGHalf = 0;
+    _cmd->status += 48;
+    if (buffer_header->data_size >= _cmd->status) {
+      _cmd->callback_function = (iso_callback_func)(_cmd->status - 16);
+    }
+    if (!DMA_SendToSPUAndSync(buffer_header->data, buffer_header->data_size, gStreamSRAM)) {
+      return 0xFFFFFFFF;
+    }
+    uint16_t v8[36];
```
Ziemas, line 902: This would be an array of sceSdBatch
```c
struct sceSdBatch {
    u16 func;
    u16 entry;
    u32 value;
};
```

game/overlord/iso.cpp
Ziemas: Indeed, the libsd call here is sceSdGetAddr(SD_VA_NAX | gVoice) which will get the current location of the voice in SPU memory.

Hat-Kid:
Closing this since Ziemas' sound PR was merged.

PR #1124, opened 2022-01-31
[graphics] MERC

water111:
Lots of weird bugs still, but it is recognizable as Jak and Daxter.

Don't look too closely at `MercProgram.cpp`... it's gross.


water111:
# MERC Bugs

## Textures failing test
It's particularly obvious on crates:
image
but also happens on various platforms/precursor doors.

## Jak is Transparent
image
Happens reliably at the village2 checkpoint. Just walk forward slightly.  Jak's hair is sometimes transparent too, but this might be okay because we're rendering his hair with the wrong renderer.


## Merc and tfrag dirt don't seem to line up
Jak's feet are on the ground correctly, but the "dirt" is above the ground:
image

This could be a tfrag issue instead of a merc issue, I'm not sure.

## Texture "flashing"
Hard to capture a screenshot, but the fisherman's boat tends to flash violently.

## Near objects (clipping?) sometimes positioned wrong.
image

The normal merc renderer can clip using a mode called "merc prime".  However, it can't always clip, evidently if an edge is too long, as seen on-screen I think. (side note: this is why all the skelgroups have a `longest-edge` thing).
The check longest edge check appears to work (has reasonable numbers, and eventually decides things are too close).  But, if you get right near the limit, merc will have some weird clipping failures.
An easy way to trigger this is the first cutscene when you talk to keira, daxter walks toward the camera.

Another good spot is to walk a bit forward
image
image



## Sky Garbage
image

Sometimes the sky is filled with garbage. This is merc triangles that are very far in the distance. This might be the same as the problem above but the object ends up partially behind the camera.

water111:
With those bug fixes, the only remaining issue is the clipping stuff. It looks like any time mscal 17 or 32 is used ("merc prime") the output is garbage.  The decision to use normal/prime appears to be correct, and the decision to go from prime -> generic is also reasonable looking.  So I don't think we're accidentally using prime when we should be using generic.

Interestingly, if you use prime in cases where you don't need it, it is bad in a similar way.  I'd expect this to work (as the guard band check that picks merc prime over normal is conservative)

water111:
Merc is now working correctly!  I've also set it up to try rendering things that would normally be rendered by other renderers, so there are still some graphical glitches caused by this.  In particular, stuff close to the camera can have clipping disasters and stuff rendered normally with generic may have bad blending/alpha settings.

PR #1129, opened 2022-02-04
random fixes + support `clang-cl` on visual studio

ManDude:


xTVaser:
We might want to document what SHA we are cloning into the repo for future reference -- previous the submodules kept track of this but since we are manually vendoring the files, that's no longer the case.  Some sort of top-level `vendor.txt` file or something would be more than enough.

ManDude:
At first I just wanted to fix up the code for a few things... but then I got a bit sidetracked and decided to revamp our dependencies (hence the enormously inflated line counts) and add `clang-cl` support for our Visual Studio projects.

- change default build type to Release instead of Debug. might help newbies.
- add `clang-cl` support to Visual Studio 2022 builds!! It works for me now but it's totally possible something is missing. You can still build with `msvc` if it doesn't work.
- unify compile flags. Now all of the projects share the same flags (like `-march=haswell` and `/arch:AVX`)
- get rid of our 11zip dependency. We were using it for the memory dump tool.
- the memory dump tool needs the raw `eeMemory.bin` file now and does not support the `.p2s` savestates. To make up for this I created a `memory-dump-p2s.py` python script. You just drag and drop the savestate file and output directory (or type them out if you want) and it extracts the savestate for you. Probably could make it automatically run the dump tool using some scripts.
- revamp the `game-text` font code a bit. I know like nothing of python. This is just to make it much easier for when we eventually add more games and versions.
- replace our `googletest`, `glfw` and `zydis` submodules with clones instead. This really inflates the line count but also allows us to edit them properly and keep them if the originals get deleted for some reason.
- fix `curve` types. They were a bit wrong and old.
- delete unused shaders.
- change borderless on windows to cut off the bottom line instead of the top line.
- add a black outline to the collide cache debug tris. It looks better now.
- hide pc options that do nothing for now. avoids newbie confusion.
- add option to forcefully draw EVERY actor in a level. You still need to be inside bsp for the game to start spawning them though. It also makes the birth/pause distance really large, so all actors will run their logic too. The CPU impact is totally negligible. You need to have "native vis" off for the option to take effect.
- extra info in entity debugger for actors.
- fix some task code. This fixes the levitator assistant not showing up. There are still more issues however due to erroneous float cast detection ( #1131 )
- fix fuel cells in progress screen in non-native widescreen.
- increase actor heap size. needed for "all actors" option.

make sure you completely clean your cmake cache, because of all the changes.

vendor.txt
```diff
@@ -0,0 +1,3 @@
+third-party/glfw:       df8d7bc892937a8b0f7c604c92a9f64f383cf48c
+third-party/googletest: 14aa11db02d9851d957f93ef9fddb110c1aafdc6
+third-party/zydis:      562a7c1e5f2c5017ea882ad0aa3360c465a3eb8f
```
water111: thanks for adding this

water111 (approved):

PR #1132, opened 2022-02-04
streamline newbie decomp/run process, update readme and gallery with new/updated information and pictures

ManDude:
This should hopefully address some questions we've been having.

This might break the project for users that haven't shifted to the more subfolder-y structure yet, but you can just move your files to fix it.


decompiler/config/jak1_ntsc_black_label/inputs.jsonc
```diff
@@ -262,5 +262,32 @@
   //"audio_dir_file_name": "jak1/VAG",
   "audio_dir_file_name": "",
 
-  "streamed_audio_file_names": ["VAGWAD.ENG", "VAGWAD.JAP"]
+  "streamed_audio_file_names": ["VAGWAD.ENG", "VAGWAD.JAP"],
+  
+  "levels_to_extract":[
```
water111, line 267: good idea to move this.  maybe we can eventually have it be automatic based on DGO name.

decompiler/main.cpp
```diff
@@ -40,8 +40,8 @@ int main(int argc, char** argv) {
     return 1;
   }
 
-  std::string in_folder = argv[2];
-  std::string out_folder = argv[3];
+  std::string in_folder = file_util::combine_path(argv[2], config.game_name);
```
water111, line 43: looks like I'll finally have to move my `iso_data`

water111 (approved):
Thanks for cleaning this up!

PR #1137, opened 2022-02-06
Implement Discord Rich Presence

Hat-Kid:
This PR adds Discord Rich Presence support to OpenGOAL using the [`discord-rpc`](https://github.com/discord/discord-rpc) library. It shows the current level you are in, as well as your power cell, orb and scout fly count. There are also some special cases to show when you are in the intro, the title screen, watching a cutscene or fighting the final boss.

In the future, it would be nice for it to also show things like whether you are driving the zoomer or riding on Flut Flut and if you are currently fighting `plant-boss` or `ogreboss`, but I haven't figured that out yet.

~~It hasn't been tested on Linux yet, but it should work on Windows unless I messed something up in CMake.~~
Tested on both Linux and Windows.

This will obviously need to be reconfigured heavily for Jak 2 and 3, but it should be a good starting point for now.


goal_src/engine/pc/pckernel.gc
```diff
@@ -135,6 +144,16 @@
   (update-from-os obj)
   (update-to-os obj)
 
+  (let ((info (new 'stack 'discord-info)))
+    (set! (-> info fuel) (&-> *game-info* fuel))
+    (set! (-> info money-total) (&-> *game-info* money-total))
+    (set! (-> info buzzer-total) (&-> *game-info* buzzer-total))
+    (set! (-> info status) "Playing Jak and Daxter: TPL")
+    (set! (-> info level) (symbol->string (-> (-> *target* current-level) name)))
```
water111, line 152: I think this could crash if there's no `*target*` spawned.  In this case, `*target*` will be `#f`  You could do something like:
```
(if *target*
    (symbol->string (-> *target* current-level name))
    "Unknown Level"
    )
```

if you want to also get the level that the camera is in, you could do:
```
(-> (level-get-target-inside *level*) name)
```
which, despite the name, the `level-get-target-inside` method will use the camera position if `*target*` isn't spawned.

Also, you can use `->` with multiple dereferences, so `(-> (-> a b) c)` is the same as `(-> a b c)`.

If you want to test with no target spawned, you can just run `(stop #f)` and it will kill the `*target*` process.

water111 (commented):
This looks awesome!  I just had one suggestion for getting the current level.

It might be a good idea to put the version or git commit hash of the discord library somewhere, just so we can remember which version we're using.

Hat-Kid:
> It might be a good idea to put the version or git commit hash of the discord library somewhere, just so we can remember which version we're using.

The version of `discord-rpc` I'm using is the [latest tagged release](https://github.com/discord/discord-rpc/releases/tag/v3.4.0) on the repo (couldn't get the latest commit to build). The library is deprecated, but is still widely used and considered stable.

water111 (approved):
Looks good to me, assuming it builds.

It looks like there's just a merge conflict to fix in `default-menu.gc`, then we can merge it.

Hat-Kid:
I'm not sure why it still fails to link the library on CI, but I just tried it again on Windows and it builds fine.

water111:
Hm... I'm not sure I can help for windows, but I can take a look at linux.

third-party/discord-rpc/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+if(WIN32)
+  find_library(DISCORD_LIB NAMES discord-rpc.lib PATHS lib)
+else()
+  find_library(DISCORD_LIB NAMES libdiscord-rpc.so PATHS lib)
```
water111, line 4: If I change `lib` to `./lib`, it works for me. The logic for these paths is... complicated and I have no idea why this matters. https://cmake.org/cmake/help/latest/command/find_library.html

Issue #1133, opened 2022-02-05
[game] runtime always crashes when taking damage from klaww boulder cannon


doctashay:

ManDude:
youre right, perfectly reproducible. I'll take a look tomorrow.

ManDude:
Caused by a wrong stack type in `ogreboss-missile-impact`. Fixing that fixed the crashes.

PR #1141, opened 2022-02-08
proper support for hardcoded "time" types

ManDude:
WIP not actually working yet


ManDude:
With this, you can set the type of a field to `time-frame` and the decompiler will try its best to appropriately convert any immediate integers of that type into `(seconds ...)` macros. The `time-frame` type is completely interchangeable with `int`/`int64`, getting these "wrong" will NOT result in a compiler error.

I've also gone ahead and fixed the detection for the `cpad` macros because it wasn't working if the pad it was checking wasn't an immediate integer.

This also fixes some decompilation errors with the endless fall and stick cameras (#1146), and also `ogreboss` (#1133), and finishes `lavatube-energy` (#1150). Also fixes #1148.

I've also reduced the stack size for the collectables since they were set before the event block stack fixes. I'm pretty sure it shouldn't crash, but feel free to raise the size if it does.

common/type_system/TypeSystem.cpp
```diff
@@ -1375,25 +1376,39 @@ bool TypeSystem::typecheck_and_throw(const TypeSpec& expected,
...
 
   if (expected == "seconds") {
-    if (actual == "seconds") {
-      return true;
-    }
-    expected = "int";
+    expected = "time-frame";
+  }
+
+  if (actual == "seconds") {
```
water111, line 1393: I wonder if we should just get rid of the `seconds` type eventually.

goal_src/engine/game/game-info-h.gc
```diff
@@ -187,9 +187,9 @@
 (define-extern *game-info* game-info)
 (when (or (not *game-info*) (zero? *game-info*))
   (let ((temp (new 'static 'game-info :mode 'debug :current-continue #f)))
-    (set! (-> temp fuel-cell-time) (new 'global 'boxed-array int64 116))
-    (set! (-> temp enter-level-time) (new 'global 'boxed-array int64 32))
-    (set! (-> temp in-level-time) (new 'global 'boxed-array int64 32))
+    (set! (-> temp fuel-cell-time) (new 'global 'boxed-array time-frame 116))
```
water111, line 190: It probably doesn't matter, but I think this is wrong - I think the compiler will make the `content-type` field of this array be `time-frame` because there's a runtime `time-frame` type.

PR #1149, opened 2022-02-10
[merc] MERC2 (in progress)

water111:
MERC2 is the (very early) PC version of MERC that will hopefully be much faster and more flexible (custom character models???) than normal merc.

The plan is:
- move most of the transformation/skinning math to the GPU. The "bone" based rendering system is somewhat unconventional in that the translation/rotation is already applied to the bones, and the VU code "just' does skinning/perspective.
- separate out the "unpack from DMA" and the other stages. This way we can cache the unpacking, or pre-unpack (like tfrag/tie). It's a bit harder for merc because ripple/texscroll/BLERC modify merc inputs in some way.  But the majority of things don't, so it's possibly still a big win.

### DIFF
--- a/decompiler/config/all-types.gc
+++ b/decompiler/config/all-types.gc
@@ -8272,7 +8272,7 @@
   )
 
 (deftype merc-fragment-control (structure)
-  ((unsigned-four-count uint8            :offset-assert 0)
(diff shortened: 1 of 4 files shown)


water111:
work is now being done on w/merc-v2 instead

PR #1161, opened 2022-02-13
cleanup and bug fix

water111:
sadly something is broken in here...


water111:
Fixed bugs:
- fixed rare camera line-of-sight tracking collision bug
- fixed lighting on jak not smoothly changing as you walk from one area to another (most easily seen entering the lower floor of green sage's hut)

PR #1174, opened 2022-02-16
[game] subtitles support (tools + goal + text file).

ManDude:


ManDude:
This adds the highly requested subtitles feature to the game! It mainly adds the feature, but I've also included actual subtitles for citadel and swamp cutscenes.

It works by creating a process (`subtitle`) that just idles around and sits on its own pool. It waits until it notices that a cutscene is playing, then syncs to it and displays the appropriate subtitles. Currently only works for streamed animation cutscenes, so the various hints in the tutorial level for example aren't gonna be detected. Eventually it should also work for game hints (e.g. voicebox or daxter) and other ambient audio (npc grumbles, etc.). Just like in future Jak games, square toggles them on/off.

The actual subtitle data is stored in `game/assets/subtitle/jak1/game_subtitle_en.txt`. I decided to make a folder for assets that will come included with the game, so our own assets like the subtitles for example. The format for the subtitles file is very similar to the `game_text.txt` file, with the difference being the way different languages are handled. Right now, they're not at all. Eventually you'll be able to parse multiple files and the information is all stored in a database that is used to compile the object files. Same as the game text file, the text strings are encoded for the game's large font specifically, which means you cannot use lowercase characters, but can use any of the available diacritics, controller buttons, kana and kanji. The small font supports all those except for controller buttons and kanji.

The format of the text file is like so:
```lisp
(language-id 0 1 2 3 4 5 6)
("redsage-resolution"
  (60 #f "RED SAGE" "(CHUCKLES) YOU'VE FINALLY COME TO RESCUE ME.")
  (190 #f "RED SAGE" "DO YOU KNOW HOW LONG I'VE BEEN IN HERE? WHAT TOOK YOU SO LONG, YOU KNOW? (CHUCKLES) WHAT ARE YOUR NAMES?")
  (429 #f "DAXTER" "I'M DAXTER! HE'S JAK, HE'S WITH ME.")
  (540 #t "RED SAGE" "GOOD JOB, DAXTER. YOU'RE A REAL HERO.")
  (670 #f "" "")
  (700 #f "RED SAGE" "YOU'VE GOT TO STOP GOL FROM LAUNCHING THE ROBOT. I'LL USE MY ECO POWER TO HELP OPEN THE SHIELD DOOR.")
  )
```
The `language-id` form defines which languages this file is meant for. I put... all 7 languages for now, who cares. The IDs match the values of the `pc-subtitle-lang` enum!! This MUST be the first form in the file.

The next form is made up of the cutscene name (`"redsage-resolution"`). Every entry after the name is a "line" of subtitles, and it is of this format:
`(frame offscreen? speaker text)`

- `frame` - the frame number at which to start displaying this line. This matches the actual frame of animation of the cutscene, - same as the commands - so it is actually in 30 units per second!
- `offscreen?` - a symbol, `#t` (true) or `#f` (false), determining whether the speaker of this line is offscreen. This should probably be left true for all hints and ambients.
- `speaker` - a string, the name of the speaker. There are three in-game settings related to the speaker name: on, off, auto. Off means speaker name is never displayed. On means it is always displayed. Auto means it only displays if the character is offscreen (determined using the setting above!). If you leave this blank, the subtitle system assumes there's "no speaker" and will act as if it was disabled for that line.
- `text` - the actual thing being said.

In the end, you get this!
screenshot_subtest

There's a lot of dialogue and we don't even have audio, so there's no rush in getting all subtitles out right now. I will add support for more kinds of subtitles, and settings for them, in the future. But not now.

ManDude:
I added a few features to the make system to flesh out the subtitles feature a bit more. I made it so that you can also supply a single argument to a `defstep` with `:arg` which can be literally anything, so the tools can use it for extra information that they need (for example, text encoding for text files). It also now supports multiple input files now, needed to compile multiple languages at once without making a mess. I added a Spanish file with 1 line so that you can get an idea of how it works, and it does work in-game too:
image
image

game/graphics/opengl_renderer/DirectRenderer.cpp
```diff
@@ -197,8 +197,8 @@ void DirectRenderer::flush_pending(SharedRenderState* render_state, ScopedProfil
     vertex_offset = 0;
   }
   glBindBuffer(GL_ARRAY_BUFFER, m_ogl.vertex_buffer);
-  glBufferSubData(GL_ARRAY_BUFFER, vertex_offset * sizeof(Vertex),
-                  m_prim_buffer.vert_count * sizeof(Vertex), m_prim_buffer.vertices.data());
+  glBufferData(GL_ARRAY_BUFFER, m_prim_buffer.vert_count * sizeof(Vertex),
```
water111, line 200: can we get rid of `vertex_offset` entirely now?

goal_src/engine/load/loader.gc
```diff
@@ -1026,6 +1023,8 @@
                   (set! (-> self skel status) (logand -5 (-> self skel status)))
                   )
               (execute-commands-up-to *load-state* (ja-aframe-num 0))
+              ;; TEMP
+              (format *stdcon* "~%~%current spool frame is ~D~%" (the int (ja-aframe-num 0)))
```
water111, line 1027: remove?

goal_src/engine/ui/text.gc
```diff
@@ -220,8 +219,8 @@
          (s3-1 s2-1 (-> *temp-string* data) load-status heap 0)
          )
         )
-      (format 0 "loaded text ~A~%" *common-text*)
-      (format 0 "~A~%" (lookup-text! *common-text* (game-text-id continue-without-saving) #f))
+      ;(format 0 "loaded text ~A~%" *common-text*)
+      ;(format 0 "~A~%" (lookup-text! *common-text* (game-text-id continue-without-saving) #f))
```
water111, line 223: feel free to just delete these lines... this is ancient.

goalc/data_compiler/game_subtitle.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <queue>
+#include "game_subtitle.h"
+#include "DataObjectGenerator.h"
+#include "common/goos/Reader.h"
+#include "common/util/FileUtil.h"
+#include "third-party/fmt/core.h"
+
+namespace {
+template <typename T>
+void for_each_in_list(const goos::Object& list, const T& f) {
```
water111, line 20: some of these might be in `parse_helpers.cpp` already

goalc/make/MakeSystem.cpp
```diff
@@ -329,4 +362,4 @@ bool MakeSystem::make(const std::string& target, bool force, bool verbose) {
   fmt::print("\nSuccessfully built all {} targets in {:.3f}s\n", deps.size(),
              make_timer.getSeconds());
   return true;
-}
\ No newline at end of file
+}
```
water111, line 365: I think as written, changing the `:arg` or adding a `:input` won't trigger a rebuild.  Maybe you should check the timestamp on the project file itself?

goalc/make/Tools.cpp
```diff
@@ -15,7 +15,7 @@
 CompilerTool::CompilerTool(Compiler* compiler) : Tool("goalc"), m_compiler(compiler) {}
 
 bool CompilerTool::needs_run(const ToolInput& task) {
-  if (!m_compiler->knows_object_file(std::filesystem::path(task.input).stem().u8string())) {
+  if (!m_compiler->knows_object_file(std::filesystem::path(task.input.at(0)).stem().u8string())) {
```
water111, line 18: it might be a good idea to have an error if you provide multiple inputs to a tool that just looks at the first one.

goal_src/game.gp
```diff
@@ -324,6 +335,15 @@
          "out/iso/6COMMON.TXT")
   )
 
+(defstep
+  :in '("game/assets/subtitle/jak1/game_subtitle_en.txt"
```
water111, line 339: an alternate way to set this up is to have a single `subtitle` file that then references all the languages files (kind of like how DGO is set up now).

ManDude: That works too, and I think I can come up with a way to only use 1 input file and also fix the `:arg` issue.

water111 (approved):

PR #1175, opened 2022-02-16
[graphics] reduce the size of fr3 files

water111:
The current size of the fr3 files is 368 MB, which is pretty big.

With these changes it goes down to about 112 MB, which is still bigger than I'd like. I think we can go a lot lower if we have less duplicate data between lods (tfrag vertices for instance, many are duplicate other than color indices)


water111:
there was a weird crash in the code coverage process. I'm going to assume it's not our fault?

Issue #1173, opened 2022-02-16
[graphics] msaa toggle


ManDude:
I totally meant to do this but it slipped my mind. There should be a way to toggle it on/off (I believe `glEnable(GL_MULTISAMPLING)` or something exists)

Jakjm:
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glEnable.xhtml
Yep looks like it!

Jakjm:
I've tried to address this with  #1176, but I don't think it's quite working - I can still feel a bit of lag when collecting a powercell, whereas commenting out `glfwWindowHint(GLFW_SAMPLES, 4);` from opengl.cpp seems to make it much smoother (but then obviously we wouldn't be able to toggle the antialiasing on or off). So I think my changes might need an adjustment.

PR #1176, opened 2022-02-16
Add toggle for MSAA 4X to menu, rename Frame Rate menu to Graphics Options

Jakjm:
Added checkbox to turn MSAA on/off. Tries to address https://github.com/open-goal/jak-project/issues/1173. However, the lag seems to still be there after unchecking the toggle - commenting out `glfwWindowHint(GLFW_SAMPLES, 4)` from line 130 of opengl.cpp seems to have a better effect, but then users can't have antialiasing even if they want it, which is less than ideal.

This is my first PR here so please let me know if I've messed anything up, I've tried to follow the style of the other files closely.

### DIFF
--- a/game/graphics/pipelines/opengl.cpp
+++ b/game/graphics/pipelines/opengl.cpp
@@ -37,6 +37,9 @@ struct GraphicsData {
   std::condition_variable sync_cv;
   bool vsync_enabled = true;
 
+  //Msaa toggle...
+  bool msaa_enabled = true;
+
   // dma chain transfer
   std::mutex dma_mutex;
   std::condition_variable dma_cv;
@@ -434,6 +437,15 @@ static void gl_render_display(GfxDisplay* display) {
   if (req_vsync != g_gfx_data->vsync_enabled) {
     g_gfx_data->vsync_enabled = req_vsync;
     glfwSwapInterval(req_vsync);
+  }
+
+  //Turn on/off msaa, if requested.
+  bool req_msaa = g_gfx_data->debug_gui.get_msaa_flag();
+  if(req_msaa != g_gfx_data->msaa_enabled){
+    g_gfx_data->msaa_enabled = req_msaa;
+    //Turn on multisampling if requested by user, turn off otherwise.
+    if(req_msaa) glEnable(GL_MULTISAMPLE);
+    else glDisable(GL_MULTISAMPLE);
   }
 
   // actual vsync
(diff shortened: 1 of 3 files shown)


game/graphics/pipelines/opengl.cpp
```diff
@@ -436,6 +439,15 @@ static void gl_render_display(GfxDisplay* display) {
     glfwSwapInterval(req_vsync);
   }
 
+  //Turn on/off msaa, if requested.
+  bool req_msaa = g_gfx_data->debug_gui.get_msaa_flag();
+  if(req_msaa != g_gfx_data->msaa_enabled){
+    g_gfx_data->msaa_enabled = req_msaa;
+    //Turn on multisampling if requested by user, turn off otherwise.
+    if(req_msaa) glEnable(GL_MULTISAMPLE);
```
water111, line 447: maybe we should call `glEnable(GL_MULTISAMPLE)` somewhere during startup just in case - I don't know what the default value will be and nothing seems to set it now. It seems like the sort of thing that might be different on different graphics drivers.

water111 (commented):
Looks good! A few things:
- We use `clang-format` to format all of our c++ files.  If the files aren't formatted, it will cause one of the tests to fail.  If you use an IDE like VSCode it probably has a built-in option to format files with clang-format. If not, you can run `clang-format -i` on the file and it will get formatted.  The clang-format settings are in `jak-project/.clang-format` but the clang-format tool (and the IDE) will probably find it automatically.
- I think when we do `glfwWindowHint(GLFW_SAMPLES, 4);`, it will generate a larger framebuffer that MSAA will read from.  Then, if you later do `glDisable(GL_MULTISAMPLE);` it will turn off MSAA, but we'll still be rendering to a large framebuffer, which is slower. So this is probably why you still have lag.  I don't know the solution to this off the top of my head, but I'll look into it.

game/graphics/pipelines/opengl.cpp
Jakjm: It looks like GL_MULTISAMPLED is enabled by default, so there's probably no need to enable it on startup.

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glEnable.xhtml
image

Thanks for the feedback! I'll get the formatting done.

Jakjm:
image
https://www.glfw.org/docs/3.3/window_guide.html#GLFW_SAMPLES
I'm going to try just setting the hint to 0 as suggested here, and let's see if that works!

ManDude:
closing in favor of #1641

PR #1177, opened 2022-02-16
decomp: `ocean-near` and `ocean-vu0`

Hat-Kid:
`ocean-near` is not fully decompiled yet (two functions remain) and I'm not 100% confident on everything in it, but I figured I'd put up what I have so far.


water111 (approved):
Looks good!

PR #1184, opened 2022-02-19
decomp: `generic-merc` | `generic-vu0` | `generic-vu1` | `shadow-vu1`

Hat-Kid:
`generic-reset-buffers` is unused and I couldn't figure out the last bit, so I uncommented it in the final output.


water111 (approved):
Looks good, thanks!

PR #1191, opened 2022-02-21
Allow symbol table expansion.

### LINKED ISSUE #1172: [kernel] expand symbol table
Jak 1 already uses ~7950 symbols of the ~8100 usable in the symbol table. While I doubt we'll run into the issue of running out of symbols (but it's totally possible), anyone wanting to add anything to the game likely will. So we should probably think about making it possible to expand the memory if necessary. Only issues I can think of is with `symbol->string`, which will fail for any code where we didn't convert it properly into that form and it still says `+ 0xff38`. I also don't know if the symbol hashing supports more symbols.

ManDude:
This allows us to expand the symbol table to any multiple of 2 if we need to, just change a couple of numbers and done!

Closes #1172


common/goal_constants.h
```diff
@@ -9,8 +9,30 @@ constexpr int BASIC_OFFSET = 4;
...
+
+constexpr s32 SYM_TABLE_MEM_SIZE = GOAL_MAX_SYMBOLS * 8 * 2;
+
+constexpr int bits_for_sym() {
+  int b = -1;
+  for (int i = 0; i < 32; ++i) {
+    if ((GOAL_MAX_SYMBOLS & (1 << i)) != 0) {
+      if (b != -1) {
+        // already got a set bit... not a multiple of 2!
+        throw 0;
```
water111, line 29: the version of `gcc` I have on my desktop complains about this `throw`. Could you make this something like:
```
static_assert(b == -1, "must be a power of 2");
```

water111 (commented):
Looks good, just one comment on the symbol bits function to make sure it compiles on old buggy versions of gcc.

PR #1192, opened 2022-02-22
[graphics] generic merc

water111:
This is the start of generic merc, it sorta works but there's some cases where the blending or lighting looks wrong, and there's a ton of not fully fixed up stuff in GenericProgram.  There's also some cleanup to do, and to add in draw bones hud.

But, the merc -> generic code is working (sadly it does not help us understand the merc format, it is even more confusing than merc), and the generic code is working (both EE and VU1).

image


water111:
It's slowly starting to work. The issues found so far:
- sometimes the lighting is clearly wrong. The wrongness seems to be per-fragment, but this is just a guess.  In some places it does seem to be right.
- there are crashes. They seem related to the merc death effect. It crashes very reliably when killing the first rat in boggy. If you don't kill anything, it seems to be stable.
- some LPC stuff is a little transparent.  I'm a little suspicious that this is a symptom of the messed up lights so not going to focus on this for now.
- ripple effect is not implemented
- some clipping cases are not implemented in GenericProgram (and are hacked around)
- the performance is very bad

water111:
For the lighting, I'm pretty sure that the `vu-light`s going in are correct. They are identical to the lights sent to merc, and merc does not have this problem. Sometimes the lighting seems to work. Other times it looks too bright.

I've been looking at this box as a good test case:

image


image

It appears that the "light" in generic is actually done on the EE, as part of the `generic-light-proc` function. The colors are unpacked from uint8's, lighting is calculated, and they're converted back.

The input color to this function is (128, 128, 128, 128), which makes sense. The values that come out aren't NaN or crazy.  The calculation is done in a VU0 microprogram (generic vu0 0), but this is such a simple one that I don't think it's wrong.

I think all this could be explained by scrambled normals.

water111:
The normals seem reasonable. It's possible they are in the wrong order though.  One thing that's a little suspicious is that the `vu-light`s (which are some weird scrambled up format) have directions:
```
#<vector      -0.0669      -0.1830       0.0000       0.0000 @ #x174040>
#<vector       0.2500       0.6830       0.0000       0.0000 @ #x174050>
#<vector       0.9659       0.7071       0.0000       0.0000 @ #x174060>
```
I do believe that the w is unused (they stash other stuff there), but the y being 0 is suspicious.

But all this works with merc, so maybe it's right?

water111:
I made a way to see the normals, and they seem wrong in the same way as the lighting

image

It's hard to tell in that screenshot, but the same is true for jak's hair.

It's somewhat surprising to me the environment mapping works if the normals are garbage, so I'm also suspicious of:
- we aren't reading the normals correctly in the lighting program.
- the normals were right, then corrupted (not sure if env map is before or after lighting) (but if they are being corrupted, they are being corrupted with things that have magnitude 1, which is lucky)

water111:
The issue with normals was with the unpack to VU1 memory:
Fixed pretty rainbow normals
image

finalboss normals look nice and smooth:
image

hud bones works too:
image


remaining issues:
- blending: platforms transparent
- blending: daxter's face in some cutscenes
- missing eyes on dogs on beach (and some others)
- ripple
- clipping (can probably put this off....)
- performance

water111:
The next bug is SPS when ripple runs.  I spent a bunch of time looking at the ripple code, and it all seems reasonable. I can't confirm that it's perfect, but I don't think it should cause SPS/NaNs.

I believe that merc/generic are looking at the `ripple-area` as a bunch of matrices (likely using the exact same code as the skeleton animation stuff).  But all that data is full of NaNs.

In `bones.gc`, the ripple-area is populated by doing some math on the output of `bones-mtx-calc`... which of course is producing garbage here.  I've been suspicious of this code for a while now - I think the plant boss weird polygon is coming from here.

Looking at the mips2c code for that, we can see it just multiplies bone matrices by joint prebind matrices.  I dumped some information in the middle, for Jak, and saw
```
bone init: -0.8818922 0.3346152 -0.33210906 0 @ 0x174500
bone init: 0.43028992 0.85914725 -0.27697632 0 @ 0x174500
bone init: 0.1926501 -0.38716754 -0.9016567 0 @ 0x174500
bone init: -640431.44 149623.11 769890.75 1 @ 0x174500
joint init: 1 -0 0 -0 @ 0x174100
joint init: -0 0.99254614 -0.12186934 0 @ 0x174100
joint init: 0 0.12186934 0.99254614 -0 @ 0x174100
joint init: -0 -1754.3708 185.23412 1 @ 0x174100
```
which is very reasonable - the upper 3x3 of `bone` is a rotation matrix, and the last row is XYZ coordinates somewhere in sandover. Makes sense for a bone.
The "joint" is likely the transformation from one of the bone to the other, when in the bind pose.  The upper 3x3 is also a transformation matrix (this time, only a rotation around the x axis), and a small offset (reasonable for a bone in jak's model)


when doing the `ripple` effect, the game looks at the first matrix output by `bones-mtx-calc`, which is NaNs.  I looked at the input, and the joint looks very wrong:
```
UPLOAD 0 : I 0x17a7b50 -> 0x100 16 <- spr to for the joint
UPLOAD 1 : X 0x1ba450 -> 0x500 24 <- spr to for the bones
bone init: 1 0 0 0 @ 0x174500
bone init: 0 1 0 0 @ 0x174500
bone init: 0 0 1 0 @ 0x174500
bone init: -431715.6 44853.637 299986.75 1 @ 0x174500
joint init: 1.6634402e+22 5.4976163e+31 8.387005e+26 4.444133e-39 @ 0x174100
joint init: 2.208642e-38 0 4.6007084e-38 4e-45 @ 0x174100
joint init: 4.6007218e-38 0 0 0 @ 0x174100
joint init: 4.600641e-38 4.6006635e-38 4.600686e-38 0 @ 0x174100
```
(but the bone is right)

I confirmed that the bone is right:
```
bone: 0
[001ba450] matrix
	[      1.0000] [      0.0000] [      0.0000] [      0.0000]
	[      0.0000] [      1.0000] [      0.0000] [      0.0000]
	[      0.0000] [      0.0000] [      1.0000] [      0.0000]
	[-431715.5937] [  44853.6367] [ 299986.7500] [      1.0000]
```
but the joint is wrong
```
joint: 0
[017a7b94] joint
	name: "align"
	number: 0
	parent: #f
	bind-pose: #<matrix @ #x17a7ba0>

[017a7ba0] matrix
	[      1.0000] [      0.0000] [      0.0000] [      0.0000]
	[      0.0000] [      1.0000] [      0.0000] [      0.0000]
	[      0.0000] [      0.0000] [      1.0000] [      0.0000]
	[      0.0000] [      0.0000] [      0.0000] [      1.0000]
```

water111:
as hex and float. Some of that joint "matrix" is just a bunch of scratchpad pointers...
```
joint init: 0x6461702d 0x742d7964 0x6c2d706f 0x30646f @ 0x174100
joint init: 0xf07ff4 0x0 0x17a7c84 0x3 @ 0x174100
joint init: 0x17a7cb4 0x0 0x0 0x0 @ 0x174100
joint init: 0x17a7b94 0x17a7be4 0x17a7c34 0x0 @ 0x174100


joint init: 1.6634402e+22 5.4976163e+31 8.387005e+26 4.444133e-39 @ 0x174100
joint init: 2.208642e-38 0 4.6007084e-38 4e-45 @ 0x174100
joint init: 4.6007218e-38 0 0 0 @ 0x174100
joint init: 4.600641e-38 4.6006635e-38 4.600686e-38 0 @ 0x174100
```

water111:
Known issues:
- SPS on frog and plant boss. I don't think this is a generic issue
- some eyes are missing. I don't think this is a generic issue
- The lava in lava tube is too dark. Generic and merc both have this problem and have completely different implementations for lighting, so I'm suspicious of some other problem (either with lights, or Direct).
- Not all clipping cases are implemented and there's some hacks to bypass these. It's way better than it used to be, but if you get close to some things, you'll either see some missing triangles, or some garbage. I'm putting this off because I'm sick of dealing with VU code and rewriting merc/generic to do stuff on the GPU will solve this issue.
- Daxter is messed up. I have no idea what's wrong here
- The performance isn't amazing. For things with environment mapping, it's roughly 2x slower than merc.

PR #1203, opened 2022-02-26
[game] HUD/sprite fixes for widescreen + a few other things

### LINKED ISSUE #1087: [hud] fix for widescreen

### LINKED ISSUE #1193: [bones] bad lod hack
Aside from the plant boss and kermit always having buggy bones, you can trigger SPS on a lot more things by messing with the LODs. This is easy to do if you force LODs to lowest quality when things spawn, then later force it to high quality.
image
image
image

This goes away once the actors animate a single time, which for doors is a little tricky since you need to make them open/close.

### LINKED ISSUE #1202: [game] automatic 16x9 letterboxing for movies

ManDude:
the HUD now looks more normal in widescreen.

image
image

Also includes some mood palette debug features.

Fixes #1087, fixes #1202, fixes #1193


ManDude:
also adds a toggleable particle effect on the orbs for better visibility.

goal_src/goal-lib.gc
```diff
@@ -546,6 +546,14 @@
   `(neq? ,thing 0)
   )
 
+(defmacro xor (a b)
+  `(not (and ,a ,b))
```
water111, line 550: I don't think this is right if a and b are both false.

ManDude: oops, not used. It turns out it's the exact same as `(neq? a b)` anyway.

PR #1212, opened 2022-03-02
[graphics] Rewrite of texture system

water111:


water111:
This ended up having a lot more changes to the loading process than I thought it would.  We now start loading PC levels ahead of time, instead of waiting for the first tfrag/tie draw.

water111:
Summary of changes:
- decompiler now checks that duplicate textures have the same data
- decompiler includes all level textures in FR3's and adds a flag for if they are part of the levels texture (not "borrowed" from other levels that happened to be loaded)
- decompiler supports texture replacement
- decompiler outputs GAME.fr3 file for common textures
- texture names in assets/texture replacement no longer have the size/mip
- got rid of graphics dumps
- engine now tells graphics systems which levels are loaded/loading
- graphics begins loading based on game level loading, not tfrag/tie draws (sometimes we need texture from loaded levels that are not drawing tfrag/tie)
- all texture are provided through the loader
- fixed some logic bugs in the vertex upload load states. We're getting closer to loading without dropping frames.
- the texture converter and cache is removed!

PR #1217, opened 2022-03-03
Add PC Port settings to legit start menu

xTVaser:
## Potential Regressions
- I removed the original logic to hide certain language options (and maybe graphics options, hard to remember) for PAL / JAPAN releases.  There may be bugs here, but I don't have the ISO to test them yet -- something we can patch up later imo as we are focused on ntsc black label.

Reminder to look at the text heap issue -- I doubled it to 64kb and it seems to work fine though.

These settings are not saved yet, ill follow up with that.


Taskfile.yml
```diff
@@ -6,17 +6,29 @@ includes:
 tasks:
   # GENERAL
   extract-jak1:
+    preconditions:
+      - sh: test -f {{.DECOMP_BIN_RELEASE_DIR}}/decompiler{{.EXE_FILE_EXTENSION}}
+        msg: "Couldn't locate decompiler executable -- Have you compiled in release mode?"
```
water111, line 11: good idea!

decompiler/data/game_text.cpp
```diff
@@ -134,25 +134,26 @@ GameTextResult process_game_text(ObjectFileData& data, GameTextVersion version)
 }
 
 std::string write_game_text(
+    const Config& cfg,
     const std::unordered_map<int, std::unordered_map<int, std::string>>& data) {
   // first sort languages:
-  std::vector<int> langauges;
+  std::vector<int> languages;
```
water111, line 140: wow, I spelled languages wrong 🤦‍♂️

decompiler/data/game_text.cpp
```diff
@@ -162,6 +163,28 @@ std::string write_game_text(
...
+    result += fmt::format("(#x{}\n  ", key);
+    for (auto const& str : val) {
+      result += fmt::format("\"{}\"\n  ", str);
+    }
+    for (int i = 0; i < languages.size() - val.size(); i++) {
+      result += fmt::format("\"{}\"\n  ", "TODO");
+    }
+    result += ")\n\n";
+  }
+
```
water111, line 187: this approach is fine for now, but in the future it might be nicer to have separate files for our custom text and the decompiled text.  We could modify the text tool to take multiple inputs and combine them.

goal_src/engine/ui/progress-h.gc
```diff
@@ -44,6 +44,8 @@
   (no-disc 32)
   (bad-disc 33)
   (quit 34)
+  ;; custom
+  (language-options 35)
```
water111, line 48: I think adding this is going to break the type consistency check.  I'd really rather not add in some feature to the compiler to ignore certain type error, so how about this idea?

We can keep `progress-h.gc` and other files in here unmodified. Then, we create a new folder like `goal_src/pc/` and put in our modified `progress-h.gc` there.  The type consistency check will look at the unmodified version. Then, you can edit the `game.gp` file to use `goal_src/pc/progress-h.gc` instead.   The type consistency check doesn't use the build system, so I think it will still pass.

goal_src/engine/ui/progress-h.gc
```diff
@@ -128,6 +130,16 @@
...
+;; new custom type to extend the very limited progress-menu system
+;; the user can only interact with one element at a time, so a single global is sufficient
+(deftype progress-menu-list-tracker (basic)
+  ((direction symbol :offset-assert 4) ; 'left | 'right
+   (transition? symbol :offset-assert 8) ; '??
+   (x-offset int32 :offset-assert 12)
+   (selected-index int32 :offset-assert 16)))
+
+;; why a new global? because i dont want to modify the type below....but that would get me what i want as well!
+(define *progress-menu-list-tracker* (the-as progress-menu-list-tracker #f))
```
water111, line 142: where do you actually allocate this?
My preference would be to put `(define *progress-menu-list-traker* (new 'static 'progress-menu-list-tracker))` here.

goal_src/engine/ui/progress/progress.gc
```diff
@@ -75,7 +75,7 @@
   "Set the options for all of the menus."
 
   ;; start off by making them all invalid
-  (dotimes (i 35)
+  (dotimes (i 36)
```
water111, line 78: could you make this `(length *options-remap*)` instead, in case we add more in the future?

goal_src/goal-lib.gc
```diff
@@ -756,6 +768,34 @@
...
+  `(dec (length ,coll)))
+
+(defmacro arr-idx-of (coll val def)
+  "Returns the index of an item in an array, returns <def> if is nothing is found.  Assumes the collection is non-empty"
+  `(let ((curr-val #f)
+         (curr-idx -1))
+    (until (or (> curr-idx (length ,coll))
+               (= curr-val ,val))
+      (set! curr-idx (inc curr-idx))
+      (set! curr-val (-> ,coll curr-idx)))
```
water111, line 794: I think this can read 2 past the end - the `or` check will let an iteration run with `curr-idx` equal to the length, then it increments past that.  This is super awkward to write without an early return, but OpenGOAL does support named blocks to return from, like this:
```
(block find-element
  (dotimes (i (lenth arr))
     (if (= val (-> arr i))
      (return-from find-element i)))
  def)
```
you could be the first to use this feature!

water111 (commented):
Looks great!  It's always fun to see new stuff being written in OpenGOAL.
I still need to check the linker thing, but I didn't see too much that was suspicious in this PR so I probably won't find anything.

I just had a few suggestions for dealing with the type consistency check

water111:
I think the pal/japan stuff is probably okay.  It might be a good idea to allow switching the language to japanese while we're at it.  As far as I know it works, but is just hidden in the progress menu.

xTVaser:
> I think the pal/japan stuff is probably okay. It might be a good idea to allow switching the language to japanese while we're at it. As far as I know it works, but is just hidden in the progress menu.

yeah this is what i currently changed it to.  I don't think i handle the UK english properly yet, because i think the original game just used that variant of english if PAL (i didnt check)

goal_src/goal-lib.gc
xTVaser: ah i originally wanted to do an early-return version but i couldn't figure it out. This looks much better.

goal_src/engine/ui/progress-h.gc
xTVaser: yeah, I like this idea a lot.  It also makes it obvious what we've heavily modified and what we havn't.  I'll do `pc/engine/ui/progress...` instead to preserve the file-tree.  Then its as simple as adding/removing `pc/` from the file paths to regression test

goal_src/engine/ui/progress/progress.gc
xTVaser: can't getting the length on the options-remap returns `0`, perhaps this is a compiler bug on enums?
image

Issue #1111, opened 2022-01-27
[memcard] save files don't work reliably


ManDude:

water111:
fixed in #1222

PR #1222, opened 2022-03-06
[game] rewrite memcard and game-save code

ManDude:


ManDude:
This fixes #1223 and #1111

ManDude:
Saving for me takes ~2.5-3.5ms while loading takes ~1-2ms. Most of the time in the save code comes from just the first part, where it updates the "memory card" information to know what files are or aren't present + the preview data. I've optimized this now so hopefully it's fast enough even on the slowest devices.

water111 (approved):
Looks great! I wouldn't be too sad if we just deleted most of the old crap in `kmemcard.cpp`. Maybe I will do this at some point.

PR #1225, opened 2022-03-08
[graphics] revert bad change to sprite shader

water111:
image


water111:
after reading more carefully, my original change was right.  Reverting the changes breaks sprites in LPC. So this PR is bad.

PR #1230, opened 2022-03-09
add ocean-vu0 mips2c and decomp ocean-texture

water111:
This adds the `ocean-vu0` functions that are used to generate the ocean vertices. I believe these vertices are patterned by the other ocean renderers.  I have no way to really confirm that the vertices that come out are prefect, but they are reasonable values, and it seems like every other vector is a normal (which makes sense, we'll need these for the environment mapping).  They change slowly, like you'd expect.

It also has the decompilation of the `ocean-texture.gc` GOAL code, which is a terrible mess of GS setup stuff.  It has a VU1 program (pretty simple) and renderers to a texture so it'll require some work on the C++ side before it does anything interesting.


water111:
This also adds the ocean-far renderer (it's really simple), and a more optimized version of ocean-texture (about 2x faster than the current unoptimized, now down to 0.2 ms, which is probably good enough for now)

Sadly ocean-mid and ocean-near are both pretty complicated. I am hoping that ocean mid and near do the same thing, but only near handles clipping, and we can do the usual trick of not using the near renderer.

PR #1232, opened 2022-03-12
[graphics] Ocean Mid (in progress)

water111:
Issues:
ocean-mid is actually two renderers - one for far away stuff that's not environment mapped and one for closer stuff that is. The transition between these two is missing, so there's an obvious line:
image

Some ocean "tiles" are flipped:
image

Sometimes we draw tiles that we probably shouldn't (this one should be left for ocean near)
image
and also some bad ocean shows up as you enter boggy (possibly the same problem)

The more complicated clipping case doesn't quite work (ride the lpc elevator down, or look straight down and move toward the ocean).


water111:
the issue with the "rotated" tile might actually be a DirectRenderer bug with the mesh:
image

water111:
a few issues fixed:
- incorrect unpack of masks, causing ocean sub-tiles to draw when it shouldn't
- ??, the issue with the scrambled beginning of strips is fixed
- fixed bugs in fcand/fcor to actually do clipping correctly. If we force ocean-near off, it uses ocean-mid everywhere and clipping looks perfect now!
- fixed VU register dependency (div src and mul dest in the same instruction), fixed SPS on clipping

Remaining issues:
- still no envmap fadeout
- ocean transition draws nothing

water111:
ocean transition/seams now seems to work. There is a slight popping as the ocean subdivides more, but I think this is expected because they are doing environment mapping using textures on a very low triangle count mesh and the inaccuracy of how the environment map texture is interpolated will actually be somewhat visible, and different, depending on the triangle size.

If you just draw the environment map, it does look like it's working properly.

There's still the issue with the sudden transition for the environment mapped ocean that I can't figure out. As far as I can tell, there's nothing that tries to change rgbas at all (they are copied directly from a table where there's no fade) and it is a sudden transition, but somehow we're doing the environment map wrong or something?

water111:
I think that's all the issues fixed. The fadeout was accomplished with some tricky mipmapping stuff.

PR #1233, opened 2022-03-14
game: auto-save pc-settings to user's home directory as well as memcard files

xTVaser:
This will automatically load the user's pc-settings if it is available from `$HOME/OpenGOAL/jak1/settings/pc-settings.gc`.  If it's not there or it cannot be loaded, the defaults will be saved and used instead.

This also saves memcard files to `$HOME/OpenGOAL/jak1/saves/*`

<details>
<summary>Fixed this issue, leaving the write up here</summary>

There's one remaining issue here that im aware of:

If i set the game's aspect-ratio during boot as part of the pc-setting's load routine, i get a crash.  If i run the same code later when the game is fully running, it's fine and works as expected.  The current code sets it as part of the `setting-control` defaults, but these two files are tightly coupled when trying to load the `pc-settings` at boot _as well_, one has to come before the other:
- pc-settings needs `*setting-control*` to be initialized so it can fix the hud layout (atleast for now)
- *setting-control* needs pc-settings to be initialized to know which aspect-ratio setting to boot with
  - this is why settings can't come last

This is how the original game changes the aspect ratio in the progress menu - https://github.com/open-goal/jak-project/blob/b2131d43e39ef87a566a1bfb57f21fa0598bf558/goal_src/engine/ui/progress/progress.gc#L1643-L1645

For clarification, im talking about running just this line at boot time, you can even add them to the bottom of `settings.gc` to reproduce:

```lisp
(set! (-> *setting-control* default aspect-ratio) 'aspect16x9)
```

I can think of hack workarounds to skip things during the setup of the `pc-settings` global, but it feels like the better solution is to figure out why its crashing only at boot time.

The current stacktrace isnt very good since this is all triggered from top-level code

```
Read symbol table (411136 bytes, 2189 reads, 2188 symbols, 8.42 ms)      
rax: 0x000000000014f14c rcx: 0x3726de336f6e0000 rdx: 0x0000000000000009 rbx: 0x0000000000f63914
rsp: 0x0000018f1a254450 rbp: 0x0000018f1a254670 rsi: 0x000000003fe38e39 rdi: 0x0000000000f63914
 r8: 0x0000000000000000  r9: 0x000000000014f12c r10: 0x0000000000f63914 r11: 0x0001000000020001
r12: 0x0000000000174474 r13: 0x0000018f1db97ea4 r14: 0x0000000000147d24 r15: 0x0000018f1a0e0000
rip: 0x0000018f1b03a13d
In unknown code
  [0x18f1b03a11d] INVALID (0x1f)
  [0x18f1b03a11d] cld
  [0x18f1b03a11e] mov r9d, [r15+r9*1+0x78]
  [0x18f1b03a123] mov rdi, rbx
  [0x18f1b03a126] add r9, r15
  [0x18f1b03a129] call r9
  [0x18f1b03a12c] lea r9, [r14+0x7408]
  [0x18f1b03a134] mov r8d, [r15+r14*1+0x5DD8]
- [0x18f1b03a13c] mov [r15+r8*1+0x1F4], r9d
  [0x18f1b03a144] mov r9d, [r15+r14*1+0x5B60]
  [0x18f1b03a14c] mov esi, [r15+r9*1+0x0C]
  [0x18f1b03a151] mov r9d, [r15+rbx*1-0x04]
  [0x18f1b03a156] mov r9d, [r15+r9*1+0x70]
  [0x18f1b03a15b] mov rdi, rbx
  [0x18f1b03a15e] add r9, r15
  [0x18f1b03a161] call r9
  [0x18f1b03a164] jmp 0x0000018F1B03A212
  [0x18f1b03a169] lea r9, [r14+0x7430]
  [0x18f1b03a171] cmp rsi, r9
  [0x18f1b03a174] jnz 0x0000018F1B03A1BC
  [0x18f1b03a17a] INVALID (0xbe)
  [0x18f1b03a17a] INVALID (0x15)

Backtrace:
   rsp: 0x18f1a254450 (#x174450) rip: 0x18f1b03a13d (#xf5a13d)
Unknown Function at rip
...
gs> (:sym-name #x5dd8)
symbol name for symbol 5DD8h is *setting-control*
gs> (:sym-name #x5B60)
symbol name for symbol 5B60h is *pc-graphics-16x9-valid-resolutions*
gs> (:sym-name #x7408)
symbol name for symbol 7408h is aspect16x9
```
</details>


ManDude:
> The current stacktrace isnt very good since this is all triggered from top-level code

In the stacktrace, we can see R8 is zero. R8 is loaded with the value of `*setting-control*`, so `*setting-control*` here is zero (e.g. not loaded).

xTVaser:
So that explains the problem, but how does it make sense that `*setting-control*` is zero when `pc-kernel.gc` builds after `settings.gc` now?

https://github.com/xTVaser/jak-project/blob/d98cf1ff2c5fe465c93d7157be5c8e2e644269b1/goal_src/game.gp#L1879

And inside `settings.gc` it initializes it if it's zero

https://github.com/xTVaser/jak-project/blob/d98cf1ff2c5fe465c93d7157be5c8e2e644269b1/goal_src/engine/game/settings.gc#L410-L411

I am fairly certain this is the entrypoint of the crash, but ill have to triple check.  Is there something else im misunderstanding?

https://github.com/xTVaser/jak-project/blob/d98cf1ff2c5fe465c93d7157be5c8e2e644269b1/goal_src/pc/pckernel.gc#L836

ManDude:
The build order is not the same as the load order. The DGO description files have the list of files in the order they are loaded.

xTVaser:
ah! kk ill look into that, didnt know about those.  Thanks

common/util/FileUtil.cpp
```diff
@@ -66,6 +79,14 @@ std::string get_project_path() {
 }
 
 std::string get_file_path(const std::vector<std::string>& input) {
+  // TODO - clean this behaviour up, it causes unexpected behaviour when working with files
+  // the project path should be explicitly provided by whatever if needed
+  // TEMP HACK
+  // - if the provided path is absolute, don't add the project path
+  if (input.size() == 1 && std::filesystem::path(input.at(0)).is_absolute()) {
```
water111, line 86: should we make this a different function, instead of having it do different things with absolute and relative paths?

I'm not opposed to this (it makes sense to let absolute paths be absolute paths), but I'm just wondering.

common/util/FileUtil.cpp
```diff
@@ -84,11 +105,17 @@ std::string get_file_path(const std::vector<std::string>& input) {
...
 bool create_dir_if_needed(const std::string& path) {
-  if (!std::filesystem::is_directory(path)) {
-    std::filesystem::create_directories(path);
-    return true;
+  if (std::filesystem::is_directory(path)) {
+    return std::filesystem::create_directories(path);
+  } else {
+    // check if the "file" has an extension
+    auto temp = std::filesystem::path(path);
+    if (temp.has_extension()) {
```
water111, line 113: will this work if you have a folder with a `.` in this name? should we have a separate function for "make folders so I can save this file?"

game/kernel/kmachine.cpp
```diff
@@ -816,6 +817,19 @@ void update_discord_rpc(u32 discord_info) {
   }
 }
 
+u64 filepath_exists(u32 filepath) {
+  auto filepath_str = std::string(Ptr<String>(filepath).c()->data());
+  if (file_util::file_exists(filepath_str)) {
```
water111, line 822: maybe we can get rid of this function and just do `std::filesystem::exists(filepath_str)`?

water111 (commented):
Looks good to me, just a few questions (nothing that needs changing if you don't want). It's probably a good idea to have @ManDude look over some of the pckernel/memory card saving stuff, but it all looks good to me.

common/util/FileUtil.cpp
xTVaser: So the main reason I had to make this change is this method is used by the file-stream stuff:
https://github.com/open-goal/jak-project/blob/da00b1ff133488afffe651d5c5964d206e70e38f/game/sce/sif_ee.cpp#L100-L102

I think long-term yeah we probably want a more explicit function like `get_project_file_path` and that function linked above should not be assuming everything is within the project directory.  I think we can do a lot of cleanup / simplification, but im not sure how much we want to do immediately.

common/util/FileUtil.cpp
xTVaser: hmmm, ill have to check, maybe potentially reverting to a separate function for now yeah.  I just wanted to get away from the pattern of:
1. make directory with a path omitting the file name
2. then append the filename to the path and work with it

and instead just allow the function to intelligently make the directory properly even if it terminates with a file name.  But as you highlight....the edge-cases for this can be annoying / potentially untenable.

game/kernel/kmemcard.cpp
```diff
@@ -100,12 +100,12 @@ void mc_print(const std::string& str, Args&&... args) {
...
-    "/BASCUS-97124AYBABTU!/bank0.bin", "/BASCUS-97124AYBABTU!/bank1.bin",
-    "/BASCUS-97124AYBABTU!/bank2.bin", "/BASCUS-97124AYBABTU!/bank3.bin",
-    "/BASCUS-97124AYBABTU!/bank4.bin", "/BASCUS-97124AYBABTU!/bank5.bin",
-    "/BASCUS-97124AYBABTU!/bank6.bin", "/BASCUS-97124AYBABTU!/bank7.bin"};
+    "BASCUS-97124AYBABTU!",           "BASCUS-97124AYBABTU!/icon.sys",
+    "BASCUS-97124AYBABTU!/icon.ico",  "BASCUS-97124AYBABTU!/BASCUS-97124AYBABTU!",
+    "BASCUS-97124AYBABTU!/bank0.bin", "BASCUS-97124AYBABTU!/bank1.bin",
+    "BASCUS-97124AYBABTU!/bank2.bin", "BASCUS-97124AYBABTU!/bank3.bin",
+    "BASCUS-97124AYBABTU!/bank4.bin", "BASCUS-97124AYBABTU!/bank5.bin",
+    "BASCUS-97124AYBABTU!/bank6.bin", "BASCUS-97124AYBABTU!/bank7.bin"};
```
ManDude, line 108: Does the leading backslash cause issues? Don't really think it's a big deal to change these strings, but it would be nice if our filepath stuff could handle redundant slashes properly.

goal_src/build/game_dgos.gc
```diff
@@ -768,6 +767,7 @@
   ("game-info.o" "game-info")
   ("game-save.o" "game-save")
   ("settings.o" "settings")
+  ("pckernel.o" "pckernel") ;; added
```
ManDude, line 770: should find a way to re-use this file one day...

goal_src/pc/pckernel-h.gc
```diff
@@ -345,6 +374,8 @@
 (define *pc-settings* (the-as pc-settings #f))
 (format 0 "PC kernel version: ~D.~D~%" PC_KERNEL_VERSION_MAJOR PC_KERNEL_VERSION_MINOR)
 
+(define *pc-temp-string-1* (new 'global 'string 2048 (the-as string #f)))
+
```
ManDude, line 378: Can't use the other `*pc-temp-string*`?

goal_src/pc/pckernel.gc
```diff
@@ -131,6 +165,15 @@
     (('5120x1440) (set-size! obj 5120 1440)))
   (none))
 
+(defmethod commit-to-file pc-settings ((obj pc-settings))
+  "commits the current settings to the file"
+  ;; auto load settings if available
+  (clear *pc-temp-string-1*)
+  (format *pc-temp-string-1* "~S/pc-settings.gc" *pc-settings-folder*)
+  (write-to-file obj *pc-temp-string-1*)
+  (clear *pc-temp-string-1*)
```
ManDude, line 174: Look at the `string-format` macro in `gstring.gc`, it simplifies doing this. You can make an alternate version of it that uses that string instead of the usual `*temp-string*`.

goal_src/pc/pckernel.gc
```diff
@@ -760,15 +805,28 @@
...
+    (format *pc-temp-string-1* "~S/pc-settings.gc" *pc-settings-folder*)
+
+    (if (pc-filepath-exists? *pc-temp-string-1*)
+      (begin
+        (format 0 "[PC] PC Settings found at '~S'...loading!~%" *pc-temp-string-1*)
+        (if (read-from-file obj *pc-temp-string-1*)
+          (return obj)
+          (begin
+            (format 0 "[PC] PC Settings found at '~S' but could not be loaded, re-creating with defaults!~%" *pc-temp-string-1*)
+            (reset obj))))
```
ManDude, line 822: hmm yes that function should probably be altered to not overwrite the settings in case an error occurred, I'll have to fix that later.

goal_src/pc/engine/ui/progress/progress-draw.gc
```diff
@@ -1774,19 +1774,21 @@
...
+                      (('pc-aspect-4x3 'orig-aspect-4x3)
+                        (draw-options-list obj font-ctx s2-1 *pc-graphics-4x3-valid-resolutions-names*))
+                      (('pc-aspect-5x4)
+                        (draw-options-list obj font-ctx s2-1 *pc-graphics-5x4-valid-resolutions-names*))
+                      (('pc-aspect-16x9 'orig-aspect-16x9)
+                        (draw-options-list obj font-ctx s2-1 *pc-graphics-16x9-valid-resolutions-names*))
+                      (('pc-aspect-21x9)
+                        (draw-options-list obj font-ctx s2-1 *pc-graphics-21x9-valid-resolutions-names*))
+                      (('pc-aspect-32x9)
+                        (draw-options-list obj font-ctx s2-1 *pc-graphics-32x9-valid-resolutions-names*))))
```
ManDude, line 1791: This menu stuff sure is messy, but if it works then it's good for now. We'll have to deal with it until we rewrite more of it.

goal_src/game.gp
```diff
@@ -1876,7 +1876,7 @@
 
 ;; Custom or Modified Code
 (goal-src "pc/pckernel-h.gc" "dma-disasm")
-(goal-src "pc/pckernel.gc" "gsound")
+(goal-src "pc/pckernel.gc" "settings")
```
ManDude, line 1879: Cool, I wanted to put the file as early as possible so that we don't run into issues where the pc stuff is used before it's properly loaded.

goal_src/pc/pckernel.gc
```diff
@@ -760,15 +805,28 @@
...
+    ;; auto load settings if available
+    ;; if saved settings are corrupted, reset them
+    (clear *pc-temp-string-1*)
+    (format *pc-temp-string-1* "~S/pc-settings.gc" *pc-settings-folder*)
+
+    (if (pc-filepath-exists? *pc-temp-string-1*)
+      (begin
+        (format 0 "[PC] PC Settings found at '~S'...loading!~%" *pc-temp-string-1*)
+        (if (read-from-file obj *pc-temp-string-1*)
+          (return obj)
```
ManDude, line 819: Are you sure you want to return immediately after reading the file?

goal_src/pc/pckernel.gc
```diff
@@ -760,15 +805,28 @@
...
+          (begin
+            (format 0 "[PC] PC Settings found at '~S' but could not be loaded, re-creating with defaults!~%" *pc-temp-string-1*)
+            (reset obj))))
+      (format 0 "[PC] PC Settings not found at '~S'...creating and initializing with defaults!~%" *pc-temp-string-1*))
+
+    ;; persist reset / new settings if we made it this far
+    (pc-mkdir-path *pc-temp-string-1*)
+    (commit-to-file obj)
+    (clear *pc-temp-string-1*)
```
ManDude, line 828: And this - is it just writing the settings after they've been reset? Probably not super useful (reading the default file and not reading a file at all is the exact same thing), and personally I'd rather it not write things at start-up in case things go wrong.

game/kernel/kmemcard.cpp
xTVaser: When being combined with `std::filesystem` it's a problem because you end up with a path with two slashes.  String concatenation of course avoids this.  We could add extra code and steps to handle duplicate slashes but, we should probably just be constructing paths correctly and not have to go through another abstraction layer.

goal_src/pc/pckernel-h.gc
xTVaser: No, I use it to construct a string that is passed to a function that then uses `*pc-temp-string*` (i make the filepath that is passed to the read/write function).  I could be wrong but, one of the problems with depending on these global variables to do things :(

goal_src/pc/pckernel.gc
xTVaser: I think resetting the settings back to defaults when they are not valid is sensible -- we aren't wiping save-file data or something important.  It's better imo than crashing.

goal_src/pc/pckernel.gc
xTVaser: From what I can tell it only really saves on the calls to `clear`?  I'd rather just keeps things explicit, it's not obvious that macro uses a particular global and the pattern seem unsustainable

ManDude: You would have to just write `(write-to-file obj (pc-string-format "~S/pc-settings.gc" *pc-settings-folder*))` instead of those 4 lines.

xTVaser: I understand that it saves lines, that isn't what I'm opinionated against.

goal_src/pc/pckernel-h.gc
ManDude: ok, it's fine then

goal_src/pc/pckernel.gc
ManDude: Well, the global only exists to be used a temporary string in the first place, so I don't think there's an issue.

xTVaser: Someone could get into the same issue I did and it wouldn't be obvious -- they call a function using the output of `pc-string-format` and within the function it also uses `pc-string-format`.  It's not obvious what's going on unless you expand the macro.  I just prefer to be more explicit.

water111 (approved):
thanks!

PR #1239, opened 2022-03-17
Overlord sound player

Ziemas:
Marked as draft because it results in a pretty huge amount of printf spam.


Ziemas:
Now also decompiled most of ssound.c.

Perhaps that should have been separate, but I'm not very organised.

Ziemas:
Now also decompiled most of ssound.c.

Perhaps that should have been separate, but I'm not very organised.

game/overlord/srpc.cpp
```diff
@@ -47,7 +72,267 @@ u32 Thread_Loader() {
...
+      snd_SetSoundVolPan(unk->sound_handle, 0x7FFFFFFF, 0);
+      snd_SetMIDIRegister(unk->sound_handle, 0, gFlava);
+    }
+
+    int n_messages = size / SRPC_MESSAGE_SIZE;
+    SoundRpcCommand* cmd = (SoundRpcCommand*)(data);
+    while (n_messages > 0) {
+      switch (cmd->command) {
+        case SoundCommand::PLAY: {
+          // spool- soundsn are vag sounds?
```
water111, line 109: yeah, that makes sense. It's a bit hacked together, but there are a few "sound effects" (started with `sound-play-by-name`) that seem to be VAG sounds and not SBK sounds, so I guess this is how it knows the difference.

game/overlord/fake_iso.cpp
```diff
@@ -298,16 +303,60 @@ void FS_PollDrive() {}
...
+    offset = 1 * 2048;
+  }
+
+  strcpy(namebuf, name);
+  namebuf[8] = 0;
+  strcat(namebuf, ".sbk");
+
+  auto file = FS_Find(namebuf);
+  if (!file) {
+    file = FS_Find("empty1.sbk");
```
water111, line 337: I think this is where the tests are failing, because we don't have empty1.sbk file available in the tests (we don't want to upload game files to github).  I haven't tried it yet, but it seems like it might be okay if you just did a `if (!file) return` here instead.

We don't run the game loop in tests, so there shouldn't try to play any sounds. But it does initialize the game code, which loads the common/empty1/empty2 banks.

water111 (approved):
All the decompiled code looks great, I think we can merge this as soon as the tests are passing. If you get stuck on the initialization stuff let me know.

Ziemas:
Would you want me to clean up the `PC_DEBUG_SOUND_ENABLE` stuff in gsound.gc now that the game functions with it enabled?

water111 (approved):

PR #1242, opened 2022-03-20
Input scheme quick fixes

charliebruce:
This adjusts the controller mapping for both gamepad and keyboard to make it more playable. Specifically:

* ~~On gamepad, the right-hand X axis is inverted, now feels correct when used with an Xbox controller. (I'm not sure if this will cause a regression if you're using a PS3 controller, or if the behaviour was incorrect on PS3 controller too - can anyone confirm?).~~
* On keyboard, analog stick movement is now implemented on WASD, and camera controls are on IJKL.
* On keyboard, buttons are now mapped in a scheme that feels somewhat playable (though controller is still preferable):
  * X: Space
  * Square: F
  * Circle: E
  * Triangle: R
  * R1: Q

### DIFF
--- a/game/system/newpad.cpp
+++ b/game/system/newpad.cpp
@@ -165,7 +165,7 @@ int AnalogValue(MappingInfo& /*mapping*/, Analog analog, int pad = 0) {
 
   // Map from input to output range
   return int((input - input_low) * (output_high - output_low) / (input_high - input_low) +
-    output_low);
+             output_low);
 }
 
 // map a button on a pad to a key
--- a/game/system/newpad.cpp
+++ b/game/system/newpad.cpp
@@ -132,7 +132,7 @@ int AnalogValue(MappingInfo& /*mapping*/, Analog analog, int pad = 0) {
     return 127;
   }
 
-  if (g_gamepads.gamepad_idx == -1) {
+  if (g_gamepads.gamepad_idx == -1) { // Gamepad not present - use keyboard
 
     // Movement controls mapped to WASD keys
     if (g_buffered_key_status[GLFW_KEY_W] && analog == Analog::Left_Y)
@@ -154,7 +154,7 @@ int AnalogValue(MappingInfo& /*mapping*/, Analog analog, int pad = 0) {
     if (g_buffered_key_status[GLFW_KEY_L] && analog == Analog::Right_X)
       input += 1.0f;
 
-  } else {
+  } else { // Gamepad present
     input = g_gamepad_analogs[(int)analog];
   }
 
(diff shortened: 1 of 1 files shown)


xTVaser:
Hey there, thanks for the PR.

> On gamepad, the right-hand X axis is inverted, now feels correct when used with an Xbox controller. (I'm not sure if this will cause a regression if you're using a PS3 controller, or if the behaviour was incorrect on PS3 controller too - can anyone confirm?).

This immediately jumps out at me, the original games use inverted X-axis camera, I'll have to double check we didn't make a mistake but we'll want to default to the original game and maybe add a flag for switching it.  I also use an xbox controller (360) and it's felt correct for me.

charliebruce:
> This immediately jumps out at me, the original games use inverted X-axis camera, I'll have to double check we didn't make a mistake but we'll want to default to the original game and maybe add a flag for switching it. I also use an xbox controller (360) and it's felt correct for me.

You're correct, just confirmed with PCSX2 - I must have been playing something else that's thrown my muscle memory off, I could have sworn that it felt incorrect originally.

Is there an easy way to add a settings entry that would allow an end user to change this? If not I'll just remove the inverted control for now.

ManDude:
We'll make an in-game toggle for camera movement control so you can play with normal or inverted movement if you'd like.

charliebruce:
Apparently if you hold R3, and press Dpad-Up when in the Graphics settings menu it will toggle the camera X axis, so the game already has some logic for handling this. Might be easiest to fix there rather than adding the hack in the gamepad logic, I'll remove that from my PR.

ManDude:
I am pretty sure that's only in the PS4 version of the game.

charliebruce:
> I am pretty sure that's only in the PS4 version of the game.

Correct, doesn't do anything in PCSX2. I've just pushed a commit which removes the inverting logic, so this now matches the unmodified PS2 behaviour. 

What's the best way to add X-axis inverting - are you thinking a patch that adds an entry to the settings menu (like the Discord RPC option)?

ManDude:
Yeah, the pause menu will have a ton of alterations for new options, it's about the only part of the game we can comfortable change a lot without negatively affecting the experience.

xTVaser:
Unless you are feeling very brave (the progress-menu code that you are referring to sucks!) you could just add an imgui option for now that would flip the behaviour.  Later on it could be more properly wired up.

water111:
looks good, thanks for cleaning this up!

The formatting check is failing, if you run `clang-format` on these files, it should pass.

Also, if you're interested, we have a discord where we talk about development here: https://discord.gg/VZbXMHXzWv

xcom169:
For some reason WASD keys are not working for me, if I ever used a DS4 gamepad. Is it possible?

PR #1247, opened 2022-03-24
[graphics] shadow volumes

water111:
In-progress shadow rendering.

Correctly capped shadow volumes appear, with correct front/back face determination.
image

but there are still a few issues:
- when the camera enters the bounding box of the shadow volume (which probably requires some tricky clipping), the shadow volume isn't drawn properly
- had to add a `magic_bogus_float_adjuster` function to correct NaNs. It does seem to do the right thing, but the vertices it has trouble are suspiciously always located in the same slot, so it feel like something isn't being scaled right somewhere. (could also be a bug in the original game that was never noticed because the PS2 deals with these large values)

The next steps are to try to hook this up to a stencil buffer and actually draw shadows. I think we can "reuse" the actual darkening draw from the original game, as it's just a full screen sprite with some alpha test settings.

To avoid the clipping junk, it might be possible to just rewrite the VU1 stuff in OpenGL.  I think we'd lose the original "am I in the shadow volume?" check (maybe?) but we could modify the jak 1 shadows to support this (jak 2's do already).


water111:
some faces seem to be missing:
image

decompiler/VuDisasm/VuDisassembler.cpp
```diff
@@ -904,6 +904,8 @@ std::string VuDisassembler::to_cpp(const VuInstruction& instr, bool mips2c_forma
       return fmt::format("fcand(vu.vi01, 0x{:x}, cf);\n", instr.src.at(0).value());
     case VuInstrK::FCOR:
       return fmt::format("fcor(vu.vi01, 0x{:x}, cf);\n", instr.src.at(0).value());
+    case VuInstrK::FSAND:
+      return fmt::format("fsand(vu.vi01, 0x{:x}, sf);\n", instr.src.at(0).value());
```
water111, line 908: todo: this is wrong

game/graphics/opengl_renderer/DirectRenderer.cpp
```diff
@@ -682,6 +682,19 @@ void DirectRenderer::handle_ad(const u8* data,
...
       break;
+    case GsRegisterAddress::RGBAQ:
+      // shadow scissor does this?
+      {
+        m_prim_building.rgba_reg[0] = data[0];
+        m_prim_building.rgba_reg[1] = data[1];
+        m_prim_building.rgba_reg[2] = data[2];
+        m_prim_building.rgba_reg[3] = data[3];
+        memcpy(&m_prim_building.Q, data + 4, 4);
+        fmt::print("rgba: {} {} {} {}: {}\n", m_prim_building.rgba_reg[0],
```
water111, line 693: rm

water111:
unsurprisingly, the glitches seem to be related to the values adjusted by `magic_bogus_float_adjuster`

water111:
the "adjustment" issue has been solved. They put in some vertices at `-FLT_MAX` (the IEEE compliant one) and expect them to survive clipping and transformation without becoming NaN (and the renderer rejects them).

The next problem is these weird lines that show up on Samos stick and the cow horns (I'm assuming these are the same issue):
image
It doesn't look too bad in this picture, but if you go in certain places you'll see really weird shadows.  The shadow stuff is confusing- if the the shadow volume isn't correct, you get weird shadow stuff leaking all over the place.

If we draw the volume, there's a obviously incorrect vertex, and we can see some green faces (these are supposed to be facing in and always invisible)
image

I played around with the VU1 code to force all vertices to be drawn with clipping (safe, but slow), and all drawn without clipping (unsafe if you get too close).  In both cases, this vertex looks like this, so I think the problem is before clipping.

PR #1262, opened 2022-03-29
ci: Add an automated release process

xTVaser:
It's a bit tedious to test these things so hopefully i didn't miss anything on the conversion from my fork back to this repo.  But this process has worked well over on pcsx2 (around 500 releases at this point, we won't be releasing nearly that much).  The test i finally got to work - https://github.com/xTVaser/jak-project/releases/tag/v0.1.3

Releases are created manually via workflow dispatch, release notes are currently not automatically generated, but should be pretty easy to do cleanly since we squash merge everything:
image

A draft release is created until all the necessary assets are uploaded:
image

And then it's live (ignore the naming, from a previous run, they should all be `open-goal-<windows/linux>-<decompiler/gk/goalc>.<bin/exe>`
image

```mermaid
sequenceDiagram
  Project Repo->>Project Repo: New tag created via workflow dispatch
  Project Repo->>Project Repo: Create a draft release
  Project Repo->>Project Repo: Run Builds
  loop Every build we intend to release
    Project Repo->>Project Repo: Upload Asset to Draft Release
    opt If All assets have been uploaded
      Project Repo->>Project Repo: Publish the release
    end
  end
  Launcher Repo->>Project Repo: Pull down latest/pinned release for it's process
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,13 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release")
 endif()
 
+option(BUILD_FOR_RELEASE "Build for release purposes (static everything)" OFF)
+if(BUILD_FOR_RELEASE)
+    set(BUILD_SHARED_LIBS OFF)
+else()
+    set(BUILD_SHARED_LIBS ON)
+endif()
+
 # Set default compile flags
 # optimization level can be set here. You can overwrite these in a per-project basis if you want.
 if(MSVC AND (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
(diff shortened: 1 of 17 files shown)


xTVaser:
Missing the dll's/so's -- can't merge until I fix that or if we statically link everything

CMakeLists.txt
```diff
@@ -9,6 +9,13 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release")
 endif()
 
+option(BUILD_FOR_RELEASE "Build for release purposes (static everything)" OFF)
```
water111, line 12: awesome, it's nice that we can control this with a single flag!

water111 (approved):
looks great! I like this workflow, and I wonder why github doesn't make it easier to set up.  We can look into the stripping thing later. Or just remove debug symbol compiler flags when RELEASE_BUILD is set, if we're stuck.

I think this has the side effect of making non-release builds dynamically link everything, but I don't see why that would be problem.

PR #1264, opened 2022-03-30
[game] 150fps support (and 100fps support)

Brent-Hickey:
In order to keep physics and game logic steady at multiple different frame-rates, the game relies on a constant of 300 ticks per second, which was chosen because it is a cleanly divisible by both 50fps and 60fps
```
(defglobalconstant TICKS_PER_SECOND 300) ;; 5 t/frame @ 60fps, 6 t/frame @ 50fps
```
Since 300 is also cleanly divisible by 150fps, we can get away with adjusting the ratios in the same way that the game does to switch between 50fps and 60fps

Most things are scaled by dividing by 2.5 or multiplying by .4 because 150 is 2.5 times faster than 60

I made a launch-particles macro so that the particle launch rate can be dynamically adjusted based on the frame rate

This now automatically chooses the '150fps or '100fps video mode based on the display refresh rate on boot and sets the frame limiter accordingly. There are only a few things wrong in '150fps mode:
- The seagulls don't fly correctly because there's some frame based timing involved. You can still get the powercell
- There are probably other minor frame-based things that are similarly off that I haven't noticed
- Things that rely on counting frames (like input frames) have shorter windows
- Jak's spool-based idle animations needed to be slowed down, but no other spools so far. I hard-coded the names of the ambient stance spools, but there might be others that I haven't noticed yet that need slowed down
- The very first black screen during the intro kinda flashed. Didn't look into this cause it only lasts a few seconds

Played through from intro to village3 without issue and can load all levels


goal_src/engine/load/loader.gc
```diff
@@ -991,7 +991,20 @@
...
+                                        (string= (-> arg0 name) "eichar-ambient-3")
+                                        (string= (-> arg0 name) "eichar-ambient-4")
+                                        )
+                                    )
+                               0.4
+                               1.0
+                               )
+                           0.05859375 (-> s2-4 speed)
+                           )
+                   )
```
ManDude, line 1007: It might be better to check if the spool is playing sound or not? But tbh I'm not sure how to properly check for that. While you're at it though, decompile `loader` and copy the new output for `ja-play-spooled-anim` and `ja-abort-spooled-anim` since it has the new `ja` macros that massively simplify things. I seem to have forgotten to do that before lol.

goal_src/engine/sparticle/sparticle.gc
```diff
@@ -678,6 +681,15 @@
...
   (cond
+    ((= arg0 2)
+     (set-vector!
+      *sp-frame-time*
+      0.000000000000000000000000000000000000011754947
+      2.0
+      0.4
+      0.4
+      )
+     )
```
ManDude, line 692: No clue why Naughty Dog is using cases here, but you can turn this whole function into just a formula:
```lisp
(set-vector! *sp-frame-time* (the-as float (logior #x800000 arg0)) (the float arg0) (* 0.2 arg0) (* 0.2 arg0))
```

goal_src/engine/load/loader.gc
Brent-Hickey: Unfortunately the idles play sound with clapping/yawning

ManDude: What if you checked for something like `(= -1 (current-str-pos spool-sound))`?

Brent-Hickey: that counts cutscenes as well it seems

Brent-Hickey: checking `(= 1 (-> arg0 parts))` works well, but I think that it might also slow down some things like tips? Not bad though

ManDude: It might break some of the death cutscenes.

Brent-Hickey: yep it does

goal_src/engine/sparticle/sparticle.gc
```diff
@@ -677,45 +680,7 @@
...
-     *sp-frame-time*
-     0.00000000000000000000000000000000000001175496
-     12.0
-     2.4
-     2.4
-     )
-    )
-   )
-  0
+  (set-vector! *sp-frame-time* (the-as float (logior #x800000 arg0)) (the float arg0) (* 0.2 arg0) (* 0.2 arg0))
```
water111, line 683: Might be worth a comment saying that this was rewritten.

I don't think it matters, but I think this will have slightly different behavior if the game is running at 15 fps.

goal_src/engine/ambient/weather-part.gc
```diff
@@ -464,22 +450,8 @@
...
-        )
-      (sp-launch-particles-var
-        *sp-particle-system-3d*
-        (-> *part-id-table* 40)
-        gp-0
-        (the-as sparticle-launch-state #f)
-        (the-as sparticle-launch-control #f)
-        1.0
-        )
+      (launch-particles :rate 1.0 (-> *part-id-table* 39) gp-0)
```
water111, line 453: Why the `:rate 1.0` here, and on a few of the others?

goal_src/engine/gfx/sky/sky-tng.gc
```diff
@@ -168,15 +168,24 @@
...
-   (+! (-> v1-0 off-t-1) 42)
-   (set! (-> v1-0 time) arg0)
+    (cond ((eq? (-> *setting-control* current video-mode) '150fps)
+           (+! (-> v1-0 off-s-0) 6.4)
+           (+! (-> v1-0 off-t-0) 12.8)
+           (+! (-> v1-0 off-s-1) -8.4)
+           (+! (-> v1-0 off-t-1) 16.8)
+           )
+          (else
+           (+! (-> v1-0 off-s-0) 16)
```
water111, line 178: Interesting... I guess the sky moved slower in 50 fps?

goal_src/engine/load/loader.gc
```diff
@@ -967,41 +950,42 @@
...
-            (set! (-> a0-42 frame-num) 0.0)
-            (joint-control-channel-group! a0-42 s2-4 num-func-seek!)
-            )
+          (ja-no-eval :group! s2-4 :num! (seek!) :frame-num 0.0)
           (when (zero? spool-part)
             (str-play-async (-> arg0 name) spool-sound)
             (set! (-> *art-control* active-stream) (-> arg0 name))
             )
-          (let* ((f30-0 (* 0.05859375 (-> s2-4 speed)))
+          (let* ((f30-0 (* (if (and (= (-> *setting-control* current video-mode) '150fps)
```
water111, line 968: might be worth a comment here

water111 (commented):
This is awesome, thanks for doing such a thorough job with this!

The `'150fps` and `BootVideoMode` stuff all looks good to me, I just had a few questions/comments.

One thing I'm wondering about is the particle speed stuff.  If there is an easy way that we can adjust the particle timing inside the sparticle code instead of at all the `sp-launch-particles-var` calls, I think that would be better.  But the particle system is kind of a nightmare, so it might not be possible.

Also, it might be helpful to add a note to the top of the files where you've made more complicated 150fps changes, just so we don't accidentally get rid of them when we're updating the decomp.

water111:
Looks like the test failure is:
```
[ 39%] [goalc   ]       goal_src/engine/load/loader.gc                                        
-- Compilation Error! --
Unrecognized symbol ja-group as head of form.
Form:
(ja-group)
Location:
goal_src/engine/load/loader.gc:903
        (when (!= (ja-group) arg1)
                   ^
Code:
(ja-group)
```
might need to move the definition of that macro earlier?

goal_src/engine/ambient/weather-part.gc
Brent-Hickey: Setting the rate is overriding it so it doesn't get set to 0.4. I don't understand why, but setting the rate to 0.4 on these cause it to crash. It seems like this is the case for any function that is like `check-drop-level` but also a couple of other places. I can't notice any difference visually/functionally and haven't really tried to understand what these are doing

This might be a reason that doing it inside the particle functions isn't possible since it can't just set everything to 0.4

goal_src/engine/gfx/sky/sky-tng.gc
Brent-Hickey: It must. These values took a while to find lol, i was just changing random sky stuff until it changed speeds

ManDude:
Oops, I guess move the macros to `process-drawable-h` then.

game/graphics/pipelines/opengl.cpp
```diff
@@ -191,6 +191,10 @@ static std::shared_ptr<GfxDisplay> gl_make_display(int width,
       g_gfx_data->debug_gui.m_vsync = false;
       g_gfx_data->vsync_enabled = false;
       glfwSwapInterval(false);
+      if (primary_monitor_video_mode->refreshRate > 120) {
```
Calinou, line 194: Should this also be done when the refresh rate is *exactly* 120 Hz (which is common on high-end TVs)? I'd wager playing at 150 FPS on a 120 Hz display is an overall better experience than playing at 60 FPS on a 60 Hz display.

That said, there could also be an option to override this regardless of your refresh rate. Even playing at 150 FPS on a 60 Hz display may be a better experience if you're looking to minimize input lag.

To account for floating-point imprecision, the check should probably be > 119 (if `refreshRate` is a float, that is).

Brent-Hickey: The 120 is arbitrary so I'm open to changing this. We can support 100fps the same way I did 150 because its a divisor of 300. Because of that, I think it'd be reasonable to set 150FPS mode for anything > 100, and then we'd set 100FPS mode for anything > 60

The reason I was restrictive is just that it's new and being automatically enabled. If you aren't able to run at a consistent 150fps the game breaks.

Also you can change video modes via the REPL so we could add it as a menu option as well

PR #1266, opened 2022-03-31
Display the currently built commit sha when debugging the game

xTVaser:
Just so when people are sharing screenshots and describing issues it is obvious if they have forgotten to rebuild, etc.
image

It's not perfectly aligned as I'm unsure how to calculate the font height properly while factoring in the resolution.


water111 (approved):
looks good!
maybe for the final release we can move it to imgui, I hate seeing the jak 1 font :slightly_smiling_face:

PR #1276, opened 2022-04-03
Add extractor tool

water111:


.github/workflows/linux-workflow.yaml
```diff
@@ -100,22 +100,16 @@ jobs:
...
-          cp ./build/game/gk ./ci-artifacts/linux-gk.bin
-          cp ./build/goalc/goalc ./ci-artifacts/linux-goalc.bin
-          chmod +x ./ci-artifacts/*.bin
-          strip --strip-debug ./ci-artifacts/linux-decompiler.bin
-          strip --strip-debug ./ci-artifacts/linux-gk.bin
-          strip --strip-debug ./ci-artifacts/linux-goalc.bin
-          ls -l ./ci-artifacts/
+          mkdir -p ./ci-artifacts/out
+          ./scripts/shell/extract_build.sh ./ci-artifacts/out ./
+          zip -r ./ci-artifacts/opengoal.zip ./ci-artifacts/out
```
Margen67, line 105: You should use `7z` instead since it's already installed.

water111: good idea, I may switch to `tar.gz` because linux people are most likely to have that, and it's already installed

Margen67: If it's meant for Linux you should use `tar`

PR #1281, opened 2022-04-04
[game] pc port progress menu

ManDude:
Changes the in-game progress menu to have different, and more PC-relevant options. The original code is mostly unchanged, instead it just gets overwritten by a later file when loading the game.

Also adds `:do-not-decompile` option to bitfield struct fields. Changes some typedefs with new enums and bitfields and fixes some method types.


ManDude:
The higher priority menus are done now. Demonstration:
20220409234607665_sqEjkrSEmE
20220409234618073_2L5S4xuagz
20220409234633456_9hQl58pM9S
20220409234650357_gNqU3NEusl
20220409234700351_fV7spJMhCj
20220409234705539_WBVjFJx1sV
20220409234709149_TkWYKs12CT
20220409234721949_EsCnYv4zY5
20220409234733341_hkrOzue6OG
20220409234742730_dXMkQTI1Zi
20220409212959592_NDkPTzgvr3
20220409234558155_CwoBEXraue

goal_src/engine/ui/progress/progress-draw.gc
```diff
@@ -1510,432 +1510,207 @@
...
-    (sv-896 string)
-    (sv-912 string)
-    )
   (let ((s3-0 (-> *options-remap* (-> obj display-state))))
     (when s3-0
       (let ((s2-1 (- arg0 (/ (* arg1 (length s3-0)) 2)))
             (s1-0 0)
+            (unkx 27)
+            (unk2 0)
+            (font (new 'stack 'font-context *font-default-matrix* 0 0 0.0 (font-color default) (font-flags shadow kerning)))
```
water111, line 1519: what was the reason for getting rid of the separate `pc` version of this file?

goal_src/pc/pckernel.gc
```diff
@@ -270,13 +168,9 @@
...
     (set! (-> info money-total) (&-> *game-info* money-total))
     (set! (-> info buzzer-total) (&-> *game-info* buzzer-total))
-    (set! (-> info status) "Playing Jak and Daxter: TPL")
-    (set! (-> info level) (if *target*
-      (symbol->string (-> *target* current-level name)) ;; use target's level if it exists
-      (symbol->string (-> (level-get-target-inside *level*) name)) ;; use camera's level otherwise
-       )
-      )
-    (set! (-> info cutscene?) (movie?))
+    (set! (-> info status) "Playing Jak and Daxter: The Precursor Legacy™")
```
water111, line 171: I'm kinda surprised that the TM doesn't cause issues here

goal_src/pc/progress-pc.gc
```diff
@@ -0,0 +1,1512 @@
+;;-*-Lisp-*-
+(in-package goal)
+
+#|
+
+  Code for the progress menu in the PC port. The original code is still loaded, this just has some overriden functions.
```
water111, line 6: ah, I see. Ignore the previous question. I can't find it to delete it.

goalc/compiler/Compiler.h
```diff
@@ -88,6 +90,7 @@ class Compiler {
   SymbolInfoMap m_symbol_info;
   std::unique_ptr<ReplWrapper> m_repl;
   MakeSystem m_make;
+  GameTextDB m_text_db;
```
water111, line 93: do we need to keep this here? Or could we construct a temporary one of these inside of `TextTool::run`?

water111: or even in `compile_game_text`?

water111 (commented):
looks good, thanks for cleaning up the flags.
I had just a few questions/comments.  

It's nothing that needs to be solved in this PR, but the settings saving is still pretty confusing to me. When I start the game, it always starts in PC widescreen (I think because I have an old version of the settings?). If I switch it to PS2 aspect ratio, everything is squashed, and the setting isn't saved (the progress menu no longer saves its settings?). Saving from the debug menu works, and after toggling the aspect ratio in the debug menu, everything is unsquashed.

common/type_system/Type.cpp
```diff
@@ -936,12 +936,16 @@ std::string BasicType::diff_impl(const Type& other_) const {
...
+    : m_type(std::move(type)),
+      m_name(std::move(name)),
+      m_offset(offset),
+      m_size(size),
+      m_skip_in_static_decomp(skip_in_decomp) {}
 
 bool BitField::operator==(const BitField& other) const {
   return m_type == other.m_type && m_name == other.m_name && m_offset == other.m_offset &&
-         other.m_size == m_size;
+         m_size == other.m_size;
```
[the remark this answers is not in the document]
ManDude: No, it's specifically *not* checked for in normal structures.

ManDude:
I actually totally forgot about saving, so the progress menu won't save anything. I'm not sure what would be a good way to fix this. Saving every frame sounds janky. But at least the button in the debug menu should work fine. I'll just bump the settings file version so that they're all cleared for everyone to avoid confusion. As for the aspect ratio, I think it just needs a few minor fixes.

common/type_system/Type.cpp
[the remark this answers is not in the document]
water111: ok, makese sense. I think if you define a type twice, and only have `do-not-decompile` in the second definition it will be silently ignored, but that's not the end of the world.

water111:
ok, that explanation makes sense. Maybe we can save every time you exit the menu. I played around with it some more, and everything that's hooked up seems to work exactly the same as using the debug menu, so it all seems good to me

xTVaser:
Is it no longer possible to save when the settings are changed / modified?  Saving when the menu exiting is going to cause potential confusion as well (what if i change settings and then force-exit the game as an extreme example).

ManDude:
That would just mean I have to copy-paste the same 1 line in a bunch of places which I really don't want to do to be honest. If a user changes a setting then the game crashes 2 seconds later before they could close the menu, I don't think it's unreasonable to have them have to do that again (not to mention, safer). Doing it on leaving the settings menu or closing the progress menu seems fine to me.

xTVaser:
So I take it it's no longer possible to just add that 1 line in a _single_ spot when the user confirms the selection? 
- https://github.com/open-goal/jak-project/blob/77c0c97fa0a95a342b6ab231c9a153415a6dccd0/goal_src/engine/ui/progress/progress.gc#L1550

goalc/compiler/Compiler.h
ManDude: The idea is that you could run for example multiple `asm-text-file` like this

```lisp
(asm-text-file text 10 "text_base.txt")
(asm-text-file text 10 "text_extra1.txt")
(asm-text-file text 10 "text_extra2.txt")
```

so the database would need to be saved somewhere like in the compiler. Right now the build system only uses 1 command to build all text so you can probably run it inside the tool for the same effect, but idk.

ManDude: I think it's possible to change it though if the "text project" system is reworked a bit.

game/assets/game_text.txt
```diff
@@ -0,0 +1,11 @@
+;; "project file" for text make tool.
```
xTVaser, line 1: a suggestion - can we avoid naming files with LISP syntax code/data in them `*.txt`?  I find this personally confusing and means they cannot be syntax highlighted ontop of that.  Perhaps this is a good use-case for the `.gd` extension instead? (GOAL Data?)

ManDude: I agree. I actually named one of them "*.gc" when messing with stuff but ended up deleting it. I'll rename them now.

common/type_system/Type.h
```diff
@@ -368,6 +370,13 @@ class EnumType : public ValueType {
   const std::unordered_map<std::string, s64>& entries() const { return m_entries; }
   bool is_bitfield() const { return m_is_bitfield; }
   std::string diff_impl(const Type& other) const override;
+  s64 max() const {
```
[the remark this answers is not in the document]
ManDude: This ended up unused so I'll just cut it out, but also I would find it strange if someone was making an enum and putting values like 5,000,000,000 in it...

goalc/compiler/Compiler.h
water111: I think it would be better to change it for a few reasons:
- it makes it easier to track dependencies
- it reduces the amount of state that the compiler has
- it moves closer to a "run the compiler once per output file" model, which I think is what we want to do for jak 2.

I think it's better to end up reading the `text_base.txt` file multiple times.

common/type_system/Type.cpp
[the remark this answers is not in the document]
ManDude: Yep, seems intentional https://github.com/open-goal/jak-project/blame/77c0c97fa0a95a342b6ab231c9a153415a6dccd0/common/type_system/Type.cpp#L117

goalc/compiler/Compiler.h
ManDude: OK I changed it. Also `asm-text-file` already worked a bit differently anyway, you pass it a list of inputs and not just a single input, like `:files ("text_base.txt" "text_extra1.txt")` so I agree it makes more sense to get rid of this stuff in the compiler structure.

water111 (approved):
:+1:

PR #1282, opened 2022-04-04
[goalc] macro expansion in integer constants

water111:


water111:
I think this is mostly good to go, but I suspect something is wrong because code is coming out slightly larger. The whole engine is ~3kB larger:
```
4907664 old
4910240 new
```

interestingly the kernel is the exact same size, so it's likely some feature like bitfields that broke.

PR #1288, opened 2022-04-08
Enable framelimiter by default

doctashay:
Flips a bool to keep the framelimiter on at boot, this helps people with refresh rates above 60hz as this is not a problem unless  you are on a 120hz/144hz monitor. 

I'm sure these settings will be reworked eventually into an .ini file or something but this is probably a useful addition in the meantime.


xTVaser:
I think this can be closed due to the linked PR above, but also that experimental 150fps has been merged recently?

Feel free to re-open if im wrong.

PR #1302, opened 2022-04-14
extractor: split up extraction process and allow overriding `data` dir path

xTVaser:
These are all the changes i made along the way to get the frontend working - https://github.com/open-goal/launcher/pull/16


third-party/CLI11.hpp
```diff
@@ -0,0 +1,9190 @@
...
+#include <type_traits>
+#include <map>
+#include <vector>
+#include <exception>
+#include <iterator>
+#include <algorithm>
+#include <cstdint>
+#include <memory>
+#include <limits>
+#include <fstream>
```
water111, line 58: wow, that's a lot of includes. let's try to only include this file in `main.cpp`'s, it's definitely going to take a long time to compile.

water111 (approved):

PR #1308, opened 2022-04-15
some fixes

### LINKED ISSUE #1271: [decompiler] fix "object files not named properly" error for real

### LINKED ISSUE #1307: [decompiler] run profiler and investigate slow speed
The decompiler is much slower on Windows, and the level extraction is like 10x slower than on Linux. This should be investigated a bit further because they really shouldn't be any different.

ManDude:
- add a `RelWithDebInfo` build for windows-clang
- change cmake flags for windows
- force windows sdk version to 7.1?
- fix some goal macros
- fix level tool being extremely slow (fixes #1307)
- fix #1271


goal_src/goal-lib.gc
```diff
@@ -438,16 +438,23 @@
...
+  `(let ((it ,condition))
+      (if it
+          ,true
+          ,false
           )
       )
   )
 
+(defmacro awhen (condition &rest body)
+  "Anaphoric when, similar to Common Lisp"
```
water111, line 453: thanks dictionary, this makes it more clear
image

decompiler/ObjectFile/ObjectFileDB.cpp
```diff
@@ -393,11 +393,14 @@ std::string ObjectFileDB::generate_obj_listing(const std::unordered_set<std::str
       result += "[\"" + pad_string(name + "\", ", 50) + "\"" +
                 pad_string(x.name_in_dgo + "\", ", 50) + std::to_string(x.obj_version) + ", " +
                 dgos + ", \"\"],\n";
-      unique_count++;
       if (all_unique_names.find(name) != all_unique_names.end() &&
           merged_objs.find(name) == merged_objs.end()) {
         lg::error("Object file {} appears multiple times with the same name.", name);
       }
+      if (merged_objs.find(name) == merged_objs.end() ||
```
water111, line 400: thanks for fixing this

common/util/Serializer.h
```diff
@@ -35,7 +36,7 @@ class Serializer {
    * later be accessed with get_save_result.
    */
   Serializer() : m_writing(true) {
-    const size_t initial_size = 32;
+    constexpr size_t initial_size = 32 * 1024 * 1024;
```
water111, line 39: I think I have a better fix for this. There's a silly bug so it resizes on every addition :man_facepalming:

water111: but this is fine for now

water111 (approved):

goal_src/goal-lib.gc
ManDude: Lol, this is what I found first http://community.schemewiki.org/?anaphoric-if

ManDude: (I don't think "anaphoric when" is a lisp thing, only "anaphoric if", so I'm kind of cheating)

PR #1314, opened 2022-04-17
cpad1

towai:
you can mess around with the load boundaries using an actual controller now! that's all this does. (ok, actually you couldn't use the analogs at all before, so no moving the vertices around - the keybinds i added can be changed, it's just what seemed to me to work best and since it's a debug function only it probably doesn't matter too much. i realize not everyone has a numpad though. ideally analog binds for both pads should be user configurable in the end)
i changed the nature of a few existing variables/functions, but nothing that leaves the scope of this one file.

### DIFF
--- a/game/system/newpad.cpp
+++ b/game/system/newpad.cpp
@@ -104,9 +104,6 @@ int IsPressed(MappingInfo& mapping, Button button, int pad = 0) {
   if (g_gamepad_buttons[pad][(int)button]) {
     return 1;
   }
-  // if (pad == 1 && g_gamepad_2_buttons[(int)button]) {
-  //   return 1;
-  // }
   auto key = mapping.pad_mapping[pad][(int)button];
   if (key == -1)
     return 0;
@@ -147,7 +144,7 @@ int AnalogValue(MappingInfo& /*mapping*/, Analog analog, int pad = 0) {
     if (g_buffered_key_status[GLFW_KEY_L] && analog == Analog::Right_X)
       input += 1.0f;
 
-  } else if (pad == 1 && g_gamepads.gamepad_idx[1] == -1) {  // try to hack in reading of second controller's analog
+  } else if (pad == 1 && g_gamepads.gamepad_idx[1] == -1) {
   
     // these bindings are not sane
     if (g_buffered_key_status[GLFW_KEY_KP_5] && analog == Analog::Left_Y)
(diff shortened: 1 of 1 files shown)


water111 (commented):
Looks good, thanks!

We require that all of our c++ code is formatted clang-format, which is why the formatting check is failing.
If you use an editor like vscode, it has a clang-format plugin that you can use. Or, use `clang-format -i` on a file to format it.

towai:
> We require that all of our c++ code is formatted clang-format, which is why the formatting check is failing. If you use an editor like vscode, it has a clang-format plugin that you can use. Or, use `clang-format -i` on a file to format it.

nice! i did not know about that plugin or that tool. i realized that the style of my edits was inconsistent with the other code, so it's totally my bad that i didn't go back over and fix it. i've pushed a new commit that fixes those formatting issues.

ManDude:
I don't know what you use, but if you're using Visual Studio you can Ctrl+A to select the entire file then copy-paste it and it will automatically format everything you pasted (e.g. the entire file). You can also see the details of where the formatting is wrong if you click on the task here that failed.

towai:
> I don't know what you use, but if you're using Visual Studio you can Ctrl+A to select the entire file then copy-paste it and it will automatically format everything you pasted (e.g. the entire file). You can also see the details of where the formatting is wrong if you click on the task here that failed.

lol, i absentmindedly dicked around in the file for absolutely no reason _after_ running `clang-format` (from VS Code's terminal) thinking a newline wouldn't affect anything. i guess everyone has to learn some way that formatting is the very last thing you do before a push if it's on a project where it matters. i had to enable the copy-paste thing in settings (maybe it's on by default, in which case i turned it off at some point because i probably found it annoying) but it works after i set clang to the default formatter. the dedicated 'format document' key combo isn't at all hard to use either, though.

ManDude:
> the dedicated 'format document' key combo isn't at all hard to use either, though.

uhh... which one is this?

towai:
> > the dedicated 'format document' key combo isn't at all hard to use either, though.
> 
> uhh... which one is this?

alt+shift+f for me. I think in the right click context menu it showed up as Ctrl+shift+f for whatever reason though. it might be there because of the extension water mentioned

ManDude:
Hmm alt+f just opens the "File" dropdown for me (and ctrl+shift+f opens the horrible abomination that is the "find in files" menu)

PR #1317, opened 2022-04-18
[decompiler] detect `seek!` and `seekl!` macros

ManDude:


ManDude:
Sadly puts us a couple hundred lines further from the 400k threshold. Darn!

PR #1320, opened 2022-04-18
Smarter `send-event` detection and fix `touching-list` bugs

ManDude:


water111 (approved):
nice fix for the touching list!

PR #1324, opened 2022-04-19
rename a macro and remove a menu option

ManDude:


water111 (approved):
:+1:
reminds me that I should probably make the decompiler recognize these at some point....

PR #1325, opened 2022-04-20
Initial Sound Implementation

Ziemas:
TODO
- [ ] Fix up naming in 989snd stuff to match rest of project
- [ ] SFX's with Flag bit 1 need to stop previous instances
- [x] Make bank unloading more robust, it needs to delete all the voices associated with the bank. (Currently trips asan due to use after free)
- [ ] setvolpan impl for midi

Probably more that i forgot.

maybe
- [ ] RE effect system (pitchbend/pitchmod) (seems unused though?)
- [ ] separate out the audio callback from 989snd (probably belongs in common code)


xTVaser (approved):
Looks good to me on Windows.  Game booted properly and sound/music works.

The only issue I found was this https://github.com/open-goal/jak-project/issues/1362 but it's a problem on master as well.

CI is passing, codacy can be ignored / cleaned up later.

Issue #1327, opened 2022-04-20
nix build from flake.nix could not find X11


heph2:
Trying to build from nix with flakes support.

Stacktrace:

```
error: builder for '/nix/store/i1pfr2vdyw2skip6vdi617fsq1sgxmam-jak-0.5.0-20220420.5369932.drv' failed with exit code 1;
       last 10 log lines:
       >   Could NOT find X11 (missing: X11_X11_INCLUDE_PATH X11_X11_LIB)
       > Call Stack (most recent call first):
       >   /nix/store/mzbylj1zprn18m7h2nsf6vmfafkvz54r-cmake-3.19.3/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:582 (_FPHSA_FAILURE_MESSAGE)
       >   /nix/store/mzbylj1zprn18m7h2nsf6vmfafkvz54r-cmake-3.19.3/share/cmake-3.19/Modules/FindX11.cmake:437 (find_package_handle_standard_args)
       >   third-party/glfw/src/CMakeLists.txt:186 (find_package)
       >
       >
       > -- Configuring incomplete, errors occurred!
       > See also "/build/source/build/CMakeFiles/CMakeOutput.log".
       > See also "/build/source/build/CMakeFiles/CMakeError.log".
       For full logs, run 'nix log /nix/store/i1pfr2vdyw2skip6vdi617fsq1sgxmam-jak-0.5.0-20220420.5369932.drv'.
```

0x715C:
@heph2 I added the needed dependencies for xorg and rapid json [here](https://github.com/0x715C/jak-project/tree/nix). This builds for me, but the flake still needs work on getting the binaries and libraries to $out with correct RPATHS etc.
I don't know enough to make more progress.

PR #1329, opened 2022-04-20
remove unused nix build files

water111:
unless somebody wants to maintain these and add tests.


water111:
will close #1327

xTVaser:
We'll want to cleanup the readme that references this as well https://github.com/open-goal/jak-project#with-nix

But yeah I don't think anyone actively working on the project uses this or maintains it, makes sense to get rid of it.

heph2:
Maybe who made the PR which add them, can say something about this. @bb010g 
An option would be to create a derivation and port them to nixpkgs, but idk

bb010g:
I can update these this weekend.

xTVaser:
Should we just move forward with this removal?  Removing what is presently broken doesn't prevent someone from contributing a fix.

water111:
Sure. I'll remove it this friday unless I hear that somebody has actually used it, or somebody is working on a fix.  If we don't have tests or any contributors using nix, I don't know that it's worth it to maintain.

PR #1330, opened 2022-04-21
Collision mesh extraction

water111:
Still WIP, but can extract instanced and non-instanced collision meshes to obj.  I will eventually add a collision renderer.

I still need to understand how the pat data is packed too.

It's going to be awful to make a packer for these collision meshes.


water111:
image

PR #1335, opened 2022-04-24
goalc: Add TCP server socket in REPL process

### LINKED ISSUE #1289: [goalc] `(reload)` doesnt save repl history

xTVaser:
Also tried to cleanup how we pass args to the compiler / how we inject code automatically into the REPL.

This allows any process to send arbitrary code over the socket into the REPL.  In the future, I'll add a client within our own code, and we'll want the server to actually be more useful by returning the result of the evaluation, and not just only accepting input.

I was able to connect to the game, build it, and run it all without touching the REPL process itself directly.

image
> note that the first line was an error, but things gracefully continued.

Closes #1289


goalc/compiler/Compiler.cpp
```diff
@@ -48,6 +52,18 @@ Compiler::Compiler(const std::string& user_profile, std::unique_ptr<ReplWrapper>
 
   // add GOOS forms that get info from the compiler
   setup_goos_forms();
+
+  m_nrepl_port = nrepl_port;
+  fmt::print("[nREPL]: Server Will Listen for a Connection on Port {}!\n\r", m_nrepl_port);
+  nrepl_thread = std::thread([&]() {
+    asio::io_context io_context;
+    ReplServer s(io_context, this);
```
xTVaser, line 60: the tests are failing for a very strange reason (no asserts / no exceptions, just dies).  And ive narrowed it down to this line being the cause, the server doesn't even have to start running.

I assuming im doing something wrong with the pointer handling and such but,  not experienced enough to know the fix

water111:
Unfortunately github's UI is too laggy for me to leave comments in the code, so I'll just put them all here.

All the changes to the I/O for the compiler look good to me, but as you've noticed, the hardest part is getting all the startup/shutdown/threading stuff to work reliably. When I was first making the listener stuff, I made some tests `test_listener_deci2.cpp` that check various orders of starting up/shutting down the compiler/runtime, and just starting/stopping connections in a loop to catch threading bugs.  It was annoying to get right, but definitely worth it in the end.  It might be worth it to add a similar test for the repl server.

## Compiler.cpp
I think it might be better to move the `ReplServer` to be outside of the `Compiler` class. It would more cleanly separate the server logic from the compiler implementation.  I believe that the public methods of `Compiler` should have everything you need. And if not, we should add more.  When possible, I like to separate these types of things - constructing the server creates a new thread (slow), opens a socket (can run into issues if you have multiple open) and likely uses at least one file descriptor (you're limited to 1024 at a time).  This approach will also make it easier avoid threading bugs. Right now, I think there can be race conditions if you run `(mi)`, then send the compiler something over the network while that runs.  Moving both the server and the repl loop outside the `Compiler` class will make it much easier to avoid this type of bug. You could have something like this:
```
// gets called from the server thread when a message is received
void on_message_received(data) {
  compiler_mutex.lock();
  compiler.compile(data);
  compiler_mutex.unlock();
}

void main() {
  Compiler compiler;
  std::thread server_thread(run_the_server);
  while (!want_exit) {
   std::string  input = read_from_stdin();
   compiler_mutex.lock();
   compiler.compile(input);
   compiler_mutex.unlock(); 
  }

  kill_the_server();
  server_thread.join();
}
```
If you want, you could make a `ServerAndCompiler` wrapper class that handles the details of managing both. Keeping things organized this way also makes cleanup/shutdown logic easier.

For `read_eval_print`, I think it would be nicer to have a `read_from_repl` that does the prompt/`read_from_stdin` and returns a `goos::Object`, and a separate `read_from_string`. Then you could have `eval_and_print(read_from_stdin())` for the normal repl and `eval_and_print(read_from_string("foo"))` for the auto input.

I think you'll need to call `join` on the `nrepl_thread` in the destructor of `Compiler`.. In C++, you have to `join` a thread before the thread object is destroyed. I bet this is why the tests are crashing - I see `terminate called without an active exception` which is usually caused by this. Having to call `join` seems annoying, but it forces you to think about when the thread will be stopped. As its written now, the `Compiler` object can be in the middle of being destroyed, and a message can come in that tries to use the partially destroyed compiler. 

## ReplServer.cpp
In `do_read`, I'm not sure this will work correctly for large inputs. Looking at https://www.boost.org/doc/libs/1_43_0/doc/html/boost_asio/reference/basic_stream_socket/async_read_some.html, it sounds like it may not read the full message. Sending only a partial message to the compiler will probably be a syntax error. It's common to send a fixed size header that contains the length of the full message, then possible do multiple reads.

I'm not a huge fan of the `enable_shared_from_this` and capturing the `self` pointers by value. This pattern keeps the `ReplSession` from being destroyed while the server is running, but the way it interacts with other code is somewhat hard to reason about, and it means you need to make _everything_ accessed through the lambda be a shared pointer. I believe the `ReplSession` can outlive the `ReplServer`. By capturing `self`, it makes sure that the `ReplSession` isn't destroyed before the lambda, but you can't be sure that the `Compiler* m_repl` is still around.  If you had a class that owned both the `Compiler` and the server thread, it would be easier to get this right - this wrapper class would be responsible for shutting down the server thread entirely before destroying the compiler.

## asio
I'm don't love this library. It's a crazy 170,000 lines of code, and I'm not really sure what most of that code does. That's about the same length as the entire Jak engine (not including level code).  I like the way that it's cross platform, but I wonder if there's something smaller we could use instead.  It doesn't seem to have too many of the features we'd find useful, like being able to call `send` or `receive` and have it take care of the details of not receiving the full message in a single packet.  I'll look around for something better. I'm not totally opposed to this library, but I don't like to have massive dependencies like this. 

Sorry I'm being kind of picky about these changes. I think it's worth it to spend timing getting this interface right if we're going to use it for a language server later on.

CMakeLists.txt
```diff
@@ -139,6 +139,9 @@ add_subdirectory(third-party/imgui)
 # build the game code in C++
 add_subdirectory(game)
 
+# build asio library
+include_directories(third-party/asio/asio/include)
+
```
ManDude, line 144: Still need this?

common/cross_sockets/XSocketServer.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <WS2tcpip.h>
+#endif
+
+XSocketServer::XSocketServer(std::function<bool()> shutdown_callback,
+                             int _tcp_port,
+                             int _buffer_size)
+    : want_exit_callback(std::move(shutdown_callback)) {
+  tcp_port = _tcp_port;
+  buffer_size = _buffer_size;
+  buffer = new char[_buffer_size];
```
water111, line 21: the codacy warning here is right - if you're allocating stuff in the constructor, you need to free it in the destructor and do something about copy constructor and assignment operator.

My suggestion is:
- replace buffer with `std::vector<char>`. If you use a vector, you don't have to worry about freeing the memory.  In the constructor, you can just do `buffer.resize(_buffer_size);`
- add this to the `.h`, inside `class XSocketServer`:
```cpp
XSocketServer(const XSocketServer&) = delete;
XSocketServer& operator=(const XSocketServer&) = delete;
```

This prevents somebody from accidentally copying the server object. This would be bad because both copies would hold the same socket, and then when the copies were eventually destroyed, they would both try to close the same socket.  If we add `=delete`, it will make it a compile time error.

water111: this was an error I had in my original `Listener` too.

common/cross_sockets/XSocketServer.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+void XSocketServer::shutdown_server() {
+  // Close the listening and accepted socket socket
+  close_server_socket();
+  close_socket(accepted_socket);
+
+  // If the accept thread is still running (nothing ever connected)
+  // kill it and clean it up
+  if (accept_thread_running) {
```
water111, line 35: maybe this should go before the closes, so the thread doesn't try to `accept` on a socket that we just closed.

common/cross_sockets/XSocketServer.h
```diff
@@ -0,0 +1,50 @@
...
+#include <mutex>
+
+/// @brief A cross platform generic socket server implementation
+class XSocketServer {
+ public:
+  static constexpr int DEF_BUFFER_SIZE = 32 * 1024 * 1024;
+  XSocketServer(std::function<bool()> shutdown_callback,
+                int _tcp_port,
+                int _buffer_size = DEF_BUFFER_SIZE);
+  ~XSocketServer();
```
water111, line 17: `virtual ~XSocketServer();`
it's a good idea to make the destructor virtual if you are using virtual methods.

scripts/shell/offline_test_git_branch.sh
```diff
@@ -48,7 +48,7 @@ echo " ================ Decompiling..."
 ../scripts/shell/decomp.sh
 
 echo " ================ Building project..."
-../scripts/shell/gc.sh -cmd \(make-group\ \"iso\"\)
+../scripts/shell/gc.sh --cmd \(make-group\ \"iso\"\)
```
water111, line 51: I totally forgot this script even exists :)

scripts/nrepl-test.py
```diff
@@ -0,0 +1,16 @@
...
+import time
+clientSocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM);
+clientSocket.connect(("127.0.0.1", 8181))
+print(clientSocket)
+data = clientSocket.recv(1024)
+print(data.decode())
+
+form = "(repl-help)"
+
+num_sent = clientSocket.send(b'\x0B\x00\x00\x00\x0A\x00\x00\x00' + form.encode())
```
water111, line 11: should be `sendall`, I think.  A plain `send` might not send the whole thing.

scripts/nrepl-test.py
```diff
@@ -0,0 +1,16 @@
+import socket
+import time
+clientSocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM);
+clientSocket.connect(("127.0.0.1", 8181))
+print(clientSocket)
+data = clientSocket.recv(1024)
+print(data.decode())
+
+form = "(repl-help)"
+
```
water111, line 10: python actually has a nice built-in library for creating stuff like this: https://docs.python.org/3/library/struct.html
```
import struct
header = struct.pack('<II', len(form), 10)
```
could replace the `b'\x0B\x00\x00\x00\x0A\x00\x00\x00'`.

water111 (commented):
Looks good! I think there are a few more small issues to work out, but it's very close.

I'm seeing a segfault when I close the compiler. I think this is because the `buffer` is being deleted twice, and this will probably just go away if you switch a `vector`.

I think the threading for the compiler/server is still not quite right. If I run `(mi)` in the normal repl, then use the python script to send `(mi)`, it runs both at the same time and crashes.

My opinion is that this logic gets a lot easier if the compiler/server don't own their own mutexes/references to each other. The compiler "reload" feature also doesn't work properly any more, and it would be really hard to implement if the server held a reference to the compiler.

I still think the right approach here is to add another layer that wraps the compiler and handles the synchronization.  I'll give this a try on another branch. We'll need to lock after a string is read from stdin and before it's read, which right now is awkwardly buried inside the `Reader`.

water111:
I did some small reorganization in this branch: https://github.com/open-goal/jak-project/pull/1343
I think it will be easier to handle the (repl or server) -> compiler logic with these changes to how the normal repl works.

common/cross_sockets/XSocket.cpp
```diff
@@ -32,6 +34,26 @@ int open_socket(int af, int type, int protocol) {
...
+  return accept(socket, addr, addrLen);
+}
+#endif
+
+#ifdef _WIN32
+int accept_socket(int socket, sockaddr* addr, int* addrLen) {
+  WSADATA wsaData = {0};
+  int iResult = 0;
+  // Initialize Winsock
+  iResult = WSAStartup(MAKEWORD(2, 2), &wsaData);
```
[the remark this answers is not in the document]
xTVaser: I think so, when we close the socket we call `WSACleanup` which depends on their being an `WSAStartup` before it.  `WSAStartup` is also safe to call multiple times apparently

xTVaser:
Closes #1289

xTVaser:
Very strange....why are only the clang tests hanging:
```
2022-05-02T00:40:04.8551741Z Testing add-int-literals.static.gc
2022-05-02T01:24:09.1333706Z ##[error]The operation was canceled.
```

water111:
```
[2022-05-02 00:40:04:592] [debug] [DECI2] Waiting for listener...
[XSocket:-1] Error writing to socket
Failed to setsockopt(-1, 1, 20, _, _) - Error: Bad file descriptor
[2022-05-02 00:40:04:742] [debug] [DECI2] Connected!
Loaded project goal_src/game.gp with 1308 steps in 327 ms
[ RUN      ] ArithmeticTests.Addition
Testing add-int-literals.static.gc
```
my guess is some sort of threading or cleanup issue. The tests create and clean up a bunch of compiler instances back to back, so it is very sensitive to bugs here. I can take a closer look tomorrow. 

Using `--gtest_repeat` and `--gtest_filter` options to run the tests many times is usually an effective way to trigger these bugs. Also running something like `stress` in the background (idk if there's an equivalent for windows) is good to simulate things under load, where the delays might be larger.

Issue #1332, opened 2022-04-22
[Releases] [0.0.7] Have Extractor.exe place files in out\data\iso_data\jak1 in release builds on windows


Zedb0T:
Currently it places them in out\data\extracted_iso
With the files in this location, if you run goalc and run (mi) you get an error as it cant find the data from the iso.

If it places them in out\data\iso_data\jak1 then goalc.exe can run (mi) and find the files correctly.

Currently a user can just manually move it though as a temporary fix 👍.

water111:
this is a good idea. I think I wanted to do this originally, but I don't remember why I didn't do it.

PR #1339, opened 2022-04-26
[extractor] extract iso to a better spot

water111:
Three changes here:
- extract iso files to `data/iso_data/jak1`. This makes `goalc`'s `(mi)` work on the release for people who want to play around.
- fix extracting for folder. It's useful for debugging because it takes a bit to extract the ISO
- only accept folders if an explicit flag is given. For actual use, we'll only support ISOs, and the launcher won't provide this flag.

Will close https://github.com/open-goal/jak-project/issues/1332


xTVaser (approved):
this looks good and sets us up better for jak2+  this was something on my mind while working on it as well.

PR #1343, opened 2022-04-30
[goalc] one possible approach to multiple threads using the compiler

water111:
the commented out `std::thread` stuff in `main.cpp` could use the REPL server to send stuff to the compiler


xTVaser:
changes incorporated here - https://github.com/open-goal/jak-project/pull/1335/commits/0a74c9ad9d691fa8e3ce4b46dfafd3fd08878e8e

PR #1344, opened 2022-04-30
collision renderer v2

### LINKED ISSUE #1346: "show memcard info" crashes game

ManDude:
Fleshes out the collision renderer to display pat information as well (along with multiple display modes and filters) and implements an interface for it in GOAL.

- fixes #1346 

image
image
image
image


ManDude:
missing:
- skip filters

bugs:
- material filter broken

PR #1350, opened 2022-05-01
[decomp] add a bunch of flag enums

ManDude:


ManDude:
also makes static sound-spec's decompile in fancy ways

decompiler/util/data_decompile.cpp
```diff
@@ -578,6 +590,147 @@ goos::Object sp_launch_grp_launcher_decompile(const std::vector<LinkedWord>& wor
...
+
+  s32 volume = word_as_s32(obj_words.at(11));
+  s32 pitch = word_as_s32(obj_words.at(12));
+  s32 bend = word_as_s32(obj_words.at(13));
+  s16 fo_min = word_as_s32(obj_words.at(14));
+  s16 fo_max = word_as_s32(obj_words.at(14)) >> 16;
+  s8 fo_curve = word_as_s32(obj_words.at(15));
+  s8 priority = word_as_s32(obj_words.at(15)) >> 8;
+  s32 auto_time = word_as_s32(obj_words.at(16));
+  s32 auto_from = word_as_s32(obj_words.at(17));
```
water111, line 651: it might be possible to reuse some of the level unpacking code here, if you end up doing a lot more of these. You can do stuff like `read_plain_data_field<s16>(ref, "fo_min", dts)`.

water111 (approved):
thanks for cleaning these up!

PR #1352, opened 2022-05-03
[decomp] better handling of animation code and art files

### LINKED ISSUE #1065: [decompiler] decompile art groups
The art groups have names of animations, models, joints etc. which are meant to be used instead of the cryptic numbers we have right now.

### LINKED ISSUE #1354: [decompiler] fancier printing for fixed-point -> float conversions
For example, `22` (in time frames) should probably decompile to `(seconds 0.075)` and not `(seconds 0.073333333)`

ManDude:
Detects some macros used for joint animation stuff. WIP, I also want to convert art element indices into their unique names.

fixes #1065 and fixes #1354


ManDude:
Works pretty well! There were 28 (or 23... I forget) failed checks throughout the whole game, I think because of some `let` weirdness, so it's too few failures to actually matter (most of them were in `target` files anyway). The size of a lot of the game object source files went down by like 20% which is massive. It was a lot of effort but it was very much worth it, now all the animation code is actually readable...

Issue #386, opened 2021-04-25
[Game] Ideas & improvements (Jak 1)


ManDude:
I'll be compiling a list of any ideas & improvements we can make to the game in this issue. Really, any idea can go even if it's really unfeasible or complex to achieve. You can add comments with your own ideas & suggestions.

Personally I'm looking for changes that don't affect direct gameplay, OR affect it strictly beneficially (e.g. fixing crashes and things like that). Anything else (audio-visually, under-the-hood technical details, etc.) is fine.

## Accessibility & i18n
- [ ] Subtitles toggle. The latter games have this. Same as those, press square during cutscenes or toggle in options menu.
  - [ ] Subtitles language option. We likely won't have subtitles for all the six languages already in the game anyway.
  - [ ] Subtitles for non-cutscene dialogue. (Yeah I know Daxter has like hours of this type of audio)
- [ ] Custom language support. Allow users to supply their own audio and text files for their language and extra language options in the options menu.

## Audiovisual
- [ ] The obvious - get rid of console-specific features that we won't/can't use:
  - [ ] CRT offset adjust.
  - [ ] Display mode (default to 60Hz).
- [ ] "PS2 version" toggles
  - [ ] PS2 mipmapping toggle.
  - [ ] PS2 LOD toggle.
  - [ ] PS2 fog toggle if you're feeling adventurous.
- [ ] Custom resolutions.
- [ ] More framerates.
- [ ] REAL aspect ratios *maybe*. The visibility data will make this difficult to achieve properly, and I'd rather not actually have to change how that works very much since some object timing could change as a result. Might work as a toggle.
  - [ ] At least support custom "fake" aspect ratios?
- [ ] Render distance settings. Highest would have very far away fog and high LODs, lowest would just be PS2 tier. Hopefully not necessary.
- [ ] Replace low-poly level meshes with high-poly meshes.
- [ ] Fix audio mixing. Jak 1 isn't winning any sound mixing awards and that could definitely be improved a little. Klaww fight would be a pretty decent place to mess with this.
- [ ] "Improved" music quality toggle. Replaces the base music channels with their preconsole mixes. This can sound wildly different at times which is why it'd be best implemented as toggleable.
- [ ] Music fade-in toggle, because I'm picky.

## Technical
- [ ] Enable/disable autosave option.
- [ ] More save slots.
- [ ] Delete save slots.
- [ ] Fix crashes. Nobody wants their game to crash, after all.
  - [ ] Warp gate crashes.
[...]

PR #1355, opened 2022-05-07
Camera inversion/flipped support for main first+third person cameras

dallmeyer:
Picked this up as a way to initially get oriented with OpenGOAL, so don't be shy to tell me if I'm implementing anything in an inappropriate fashion.

- I saw some posts in discord about using "flipped" instead of "inverted" to avoid confusion, so have replaced that across the board
- I split out settings for first-person cameras vs third-person cameras to give more granular control
screenshot
screenshot

Any non-debug-mode cameras that can be controlled by the right stick should now respect these normal/flipped settings (e.g. fish game is only controlled via the left stick, so it's unaffected)

~I needed to do some manual intervention for a couple changes which might cause backwards compatibility issues for others - not sure how much we care about that in this stage?~
- ~Since I renamed some game-text entries I needed to bypass a validation in [TypeSystem.cpp](https://github.com/open-goal/jak-project/blob/7b6d732a7719580f0ad5bfe71ca5964d69091811/common/type_system/TypeSystem.cpp#L59-L74)~ (EDIT: just needed to use `(reload)`)
- ~Since I split and renamed the pc-settings, I needed to manually modify my saved settings in `OpenGOAL\jak1\settings\pc-settings.gc` accordingly~ (EDIT: can't reproduce this, was probably relating to me not utilizing `(reload)`)


goal_src/engine/camera/cam-states.gc
```diff
@@ -2324,11 +2326,16 @@
...
+      (if (and PC_PORT (-> *pc-settings* third-camera-vflip?)) 1.0 -1.0)
+      f0-0
     )
   )
+)
 
+;; main third-person camera
 (defbehavior cam-string-joystick camera-slave ()
   (set! (-> self options) (logand -257 (-> self options)))
   (let ((f28-0 (cam-dist-analog-input (the-as int (-> *cpad-list* cpads 0 righty)) (the-as float 0.05)))
```
dallmeyer, line 2341: `cam-dist-analog-input` is the equivalent of `v3-analog-input` for `cam-string-joystick`, since it has a unique vertical camera control

game/assets/jak1/text/game_text_en.gs
```diff
@@ -8,12 +8,16 @@
...
+(#x1002 "FLIPPED"
+        "FLIPPED")
+(#x1003 "FIRST-PERSON HORIZONTAL CAMERA CONTROL"
+        "FIRST-PERSON HORIZONTAL CAMERA CONTROL")
+(#x1004 "FIRST-PERSON VERTICAL CAMERA CONTROL"
+        "FIRST-PERSON VERTICAL CAMERA CONTROL")
+(#x1005 "THIRD-PERSON HORIZONTAL CAMERA CONTROL"
+        "THIRD-PERSON HORIZONTAL CAMERA CONTROL")
+(#x1006 "THIRD-PERSON VERTICAL CAMERA CONTROL"
+        "THIRD-PERSON VERTICAL CAMERA CONTROL")
```
ManDude, line 20: See if you can find a way to shorten this text. Lines on the progress menu should not break (be 2 lines large) at 4x3.

In the `post` for `progress-normal` (the overriden one) you can find the draw functions and change the parameters to make the text slighty smaller if you need to.

goal_src/engine/ps2/pad.gc
```diff
@@ -172,6 +172,38 @@
...
+)
+
+(defun h3-analog-input ((in int) (offset float) (center-val float) (max-val float) (out-range float))
+  "Same as analog-input but respects Third-Person Horizontal camera control setting."
+  (*
+    (if (and PC_PORT (-> *pc-settings* third-camera-hflip?)) 1.0 -1.0)
+    (analog-input in offset center-val max-val out-range)
+  )
+)
+
```
ManDude, line 206: You can probably give these a more descriptive name like `analog-input-horizontal-third` and make them a macro instead of a whole since it's a small piece of code and the arguments are used only once, and it saves us 4 symbols. It's a very simple replacement macro so you just add a backquote ` at the start and a comma , before each of the arguments.

goal_src/levels/jungle/jungle-mirrors.gc
```diff
@@ -694,15 +695,15 @@
...
+            (let ((f26-0 (h1-analog-input
                            (the-as int (+ (-> *cpad-list* cpads 0 rightx) -256 (-> *cpad-list* cpads 0 leftx)))
                            0.0
                            48.0
                            110.0
                            -1.0
                            )
                          )
-                  (f0-0 (analog-input
+                  (f0-0 (v1-analog-input
```
ManDude, line 706: Oh yeah, also leave a comment saying this was changed for the pc port

water111:
Awesome - great work figuring this out!

I think there is one small issue that's causing a test to fail:
```
-- Compilation Error! --
No method or function named analog-input for type int
Form:
(analog-input in offset center-val max-val out-range)
Location:
goal_src/engine/ps2/pad.gc:179
    (analog-input in offset center-val max-val out-range)
     ^
Code:
(analog-input in offset center-val max-val out-range)
```

If you move your new input functions to below the `(defun analog-input`, the error should go away. The OpenGOAL compiler does a single pass, so you need to either forward declare a function (with `define-extern`) or have the actual definition (with `defun`) before the function is used.  

For the backward compatibility stuff:
- I'm not sure what you mean about the text entry renaming exactly, but what you've done looks correct to me. Adding the text ids has to go in both `all-types.gc` and `text-h.gc`.  The purpose of the `all-types.gc` file is to tell the decompiler all the type information, and there's a test to make sure that type definitions on `all-types.gc` match up exactly with the ones produced when compiling the game.  For new code that we add, this check doesn't do much help, but it's still a good idea to keep everything up to date.
- For the settings file, what happened when you tried to start the game with the wrong settings file? If it's crashing, maybe we should update the `PC_KERNEL_VERSION` so it rejects the old files. Maybe @ManDude has other suggestions.

ManDude (changes requested):
Looks good to me. I'm just being picky about the menu text here. Your picture is in widescreen which gives it even more space, but you should always check this stuff in 4x3 since that's the original aspect ratio.

ManDude:
I have not checked the tests yet since I'm on my phone and just wanted to go through the code right now.

dallmeyer:
> For the backward compatibility stuff:
> 
> * I'm not sure what you mean about the text entry renaming exactly, but what you've done looks correct to me. Adding the text ids has to go in both `all-types.gc` and `text-h.gc`.  The purpose of the `all-types.gc` file is to tell the decompiler all the type information, and there's a test to make sure that type definitions on `all-types.gc` match up exactly with the ones produced when compiling the game.  For new code that we add, this check doesn't do much help, but it's still a good idea to keep everything up to date.

Hm I had previously run `(mi)` with the original text, `inverted` for example, and that built fine. Then when I replaced some test entries, such as `inverted` to `flipped`, it would fail to run `(mi)` unless I tweaked the code to allow redfinition [within TypeSystem.cpp](https://github.com/open-goal/jak-project/blob/7b6d732a7719580f0ad5bfe71ca5964d69091811/common/type_system/TypeSystem.cpp#L59-L74)

> * For the settings file, what happened when you tried to start the game with the wrong settings file? If it's crashing, maybe we should update the `PC_KERNEL_VERSION` so it rejects the old files. Maybe @ManDude has other suggestions.

Yeah it was crashing, I think failing to parse "legacy" symbols or failing to find my new symbols - I can try to reproduce to get a more exact answer.

ManDude:
> Hm I had previously run (mi) with the original text, inverted for example, and that built fine. Then when I replaced some test entries, such as inverted to flipped, it would fail to run (mi) unless I tweaked the code to allow redfinition [within TypeSystem.cpp](https://github.com/open-goal/jak-project/blob/7b6d732a7719580f0ad5bfe71ca5964d69091811/common/type_system/TypeSystem.cpp#L59-L74)

Like with before, this is because the compiler has a state and when you ran `(mi)` for the first time, it was given information about the `pc-settings` type and the `game-text-id` enum. Then, you altered them, then ran `(mi)` again, which meant the compiler parsed the `deftype` for `pc-settings` again, which doesn't match what it had before (obviously!). This is a bit of a flaw but it exists for a very good reason, it lets us catch bugs when altering types. If you're editing code and you're changing types and enums (constants and macros are actually fine to redefine!), you should run `(reload)` which reboots the compiler, cleaning its state.

dallmeyer:
> > Hm I had previously run (mi) with the original text, inverted for example, and that built fine. Then when I replaced some test entries, such as inverted to flipped, it would fail to run (mi) unless I tweaked the code to allow redfinition [within TypeSystem.cpp](https://github.com/open-goal/jak-project/blob/7b6d732a7719580f0ad5bfe71ca5964d69091811/common/type_system/TypeSystem.cpp#L59-L74)
> 
> Like with before, this is because the compiler has a state and when you ran `(mi)` for the first time, it was given information about the `pc-settings` type and the `game-text-id` enum. Then, you altered them, then ran `(mi)` again, which meant the compiler parsed the `deftype` for `pc-settings` again, which doesn't match what it had before (obviously!). This is a bit of a flaw but it exists for a very good reason, it lets us catch bugs when altering types. If you're editing code and you're changing types and enums (constants and macros are actually fine to redefine!), you should run `(reload)` which reboots the compiler, cleaning its state.

Ah got it, yeah it makes sense what it was catching and the motivation behind it, just wasn't sure the "right" way to force my change through - good to know about `(reload)`

dallmeyer:
Do we care if the menu line breaks when you select it to change between normal/flipped? It wasn't intentional but it almost looks better this way imo... though I can shrink the text further if preferred
screenshot

ManDude:
If you want to you can change some code to make it so it doesn't show normal/flipped when the option isn't selected.

water111:
@ManDude does this look good to you?

ManDude:
Oops sorry I missed that.

@dallmeyer see if you can change "CAM" to "CAMERA", reduce the line height (it's the `36` I believe), and make the "FLIPPED"/"DEFAULT" text not appear when option isn't selected, it's [here](https://github.com/open-goal/jak-project/blob/master/goal_src/pc/progress-pc.gc#L1173) I believe.

goal_src/pc/progress-pc.gc
```diff
@@ -1302,12 +1306,12 @@
...
-                  (((game-option-type normal-inverted))
+                  (((game-option-type normal-flipped))
                    ;; etc
                    (set! option-str (string-format "~S: ~S" (lookup-text! *common-text* (-> options index name) #f)
                                                         (if (-> (the-as (pointer uint32) (-> options index value-to-modify)))
                                                             (lookup-text! *common-text* (game-text-id normal) #f)
-                                                            (lookup-text! *common-text* (game-text-id inverted) #f)
+                                                            (lookup-text! *common-text* (game-text-id flipped) #f)
                                                             )))
                    )
```
ManDude, line 1316: Remove this entire case and add it to where all the other ones are (it's right above this, but I can't link it for some reason).

goal_src/pc/progress-pc.gc
```diff
@@ -1424,7 +1428,7 @@
         (((progress-screen camera-options))
          ;; camera options lines are a bit too big
          (hide-progress-icons)
-         (draw-options self 115 36 0.77)
+         (draw-options self 115 36 0.60)
```
ManDude, line 1431: cool. Just reduce the line height now. It's the 36 in here. I don't remember why I set it so high.

game/assets/jak1/text/game_text_en.gs
```diff
@@ -8,12 +8,16 @@
...
+(#x1002 "FLIPPED"
+        "FLIPPED")
+(#x1003 "1ST-PERSON HORIZONTAL CAM"
+        "1ST-PERSON HORIZONTAL CAM")
+(#x1004 "1ST-PERSON VERTICAL CAM"
+        "1ST-PERSON VERTICAL CAM")
+(#x1005 "3RD-PERSON HORIZONTAL CAM"
+        "3RD-PERSON HORIZONTAL CAM")
+(#x1006 "3RD-PERSON VERTICAL CAM"
+        "3RD-PERSON VERTICAL CAM")
```
ManDude, line 20: now see if you can have "CAMERA" instead of "CAM"! If not just reduce the scale a tiny bit?

ManDude (changes requested):

goal_src/pc/progress-pc.gc
ManDude: It will change the unselected text to just be the option's text and doesn't add the FLIPPED/DEFAULT at the end.

dallmeyer:
@ManDude @water111 how's this look?

screenshot

screenshot

ManDude (approved):
Looks perfect now. Thanks!

dallmeyer:
Decided to support the main debug camera `cam-free-floating-input` too. There are a handful of other cameras that I'm not sure how to test so will prob leave alone for this PR:

- cam-layout.gc
  - `defun clmf-input`
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-layout.gc#L807-L808
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-layout.gc#L818-L819
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-layout.gc#L827-L828
- cam-states-dbg.gc
  - `defstate cam-point-watch`
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-states-dbg.gc#L52-L53
  - `cam-free-floating-input` specifically `when *display-load-boundaries*` (something involving 2nd controller?)
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-states-dbg.gc#L309-L310
  - `defstate cam-orbit` (seemed to only be used in some animation test mode)
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-states-dbg.gc#L516
    - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/camera/cam-states-dbg.gc#L557-L558
 - load-boundary.gc
   - `defun-debug edit-load-boundaries`
     - https://github.com/dallmeyer/jak-project/blob/d31f234da9cc8757caa772f7c71f89f67374f430/goal_src/engine/level/load-boundary.gc#L523-L524

ManDude (commented):
This is just camera debugging and some other editors for load boundaries and camera modes, I don't think it's super worth it to bother with them.

water111:
Everything looks good - I'll merge this as soon as the tests run (takes about 10 minutes)

dallmeyer:
@ManDude no rush but at some point would you mind checking off this camera inversion feature on [this list](https://github.com/open-goal/jak-project/issues/386#issue-867068953)?

PR #1359, opened 2022-05-18
`anim-tester-x` + cam flip fixes + `logxor!` decomp

ManDude:
Made a new anim tester that crashes a bit less and is slightly more flexible. It lacks the custom animation playback features the original did, but I can add those a bit later.

Also fixes some camera invert things and makes the decompiler detect `logxor!`.


goal_src/engine/ps2/pad.gc
```diff
@@ -202,35 +202,35 @@
 
 (defmacro analog-input-vertical-first (in offset center-val max-val out-range)
   "Same as analog-input but respects First-Person Vertical camera control setting."
-  `(*
-    (if (and PC_PORT (-> *pc-settings* first-camera-vflip?)) 1.0 -1.0)
-    (analog-input ,in ,offset ,center-val ,max-val ,out-range)
+  `(#if PC_PORT
+        (* (if (-> *pc-settings* first-camera-vflip?) -1.0 1.0) (analog-input ,in ,offset ,center-val ,max-val ,out-range))
+        (analog-input ,in ,offset ,center-val ,max-val ,out-range)
+        )
```
dallmeyer, line 208: Hah good catch, this was "working" for me because the value of `first-camera-vflip?` is also backwards. "Normal" = "On" = true, and "Flipped" = "Off" = false. Looking back I think I did notice this but left it as-is since that menu logic predated my PR.

I think we need to change how these options are saved from the UI too, otherwise this only makes sense when you're looking at the variable names/values directly. I can open a separate PR for fixing both ends of this

dallmeyer: https://github.com/open-goal/jak-project/pull/1360

PR #1360, opened 2022-05-18
Fix camera inversion booleans so variable names make sense

dallmeyer:
The original setup for the normal/flipped options in the menu mapped "Normal" to true and "Flipped" to false, which was backwards of how the settings were named e.g. `third-camera-vflip?` being true would suggest it should be "Flipped"

This PR fixes the mapping so "Normal" is false and "Flipped" is true, and correspondingly updates the actual camera controls.

One byproduct of this is that the ordering in the menu is reversed from Normal/Flipped to Flipped/Normal:
screenshot
If we care to swap this back I think the easiest fix might be to just rename the internal settings themselves? e.g. `third-camera-vnormal?` or something


ManDude:
Shouldn't the text be "NORMAL"/"INVERTED" then? The default controls are inverted.

dallmeyer:
> Shouldn't the text be "NORMAL"/"INVERTED" then? The default controls are inverted.

they are and they aren't depending on first/third person, there was a bit of [discussion in discord](https://discord.com/channels/756287461377703987/756287461830688810/960609891968483359) about this and how "inverted" would be confusing. I went with normal/flipped as in the original jak controls vs the opposite of the original

ManDude:
I still think we should use NORMAL/INVERTED because it's more consistent with modern games. They all use these terms and people know what they mean.

dallmeyer:
Not necessarily opposed to that, but it will make the code a bit more confusing since in some cases we will multiply by -1.0 for inverted, and in other cases not. I can take a stab at that later today to see what it ends up looking like.

ManDude:
Multiplying by -1.0 just means whatever the original value was got "flipped". That's why things are named that way, by turning the `flip?` things on you are simply flipping the analog inputs.

dallmeyer:
Ok opened a PR to use the commonly-understood "inverted" terminology/behavior. [This is the piece I was talking about where multiplying by -1.0 vs 1.0 is inconsistent](https://github.com/open-goal/jak-project/pull/1361/files#diff-acb1343cfcd940a82ab465c53d32dcd9a6e0d197c254ac86dcb926a7f4b0575fR206/) depending on which camera - hopefully the comments I've left make it clear

dallmeyer:
closing in favor of above PR

PR #1368, opened 2022-05-20
logs: replace every `fmt::print` with a `lg` call instead

### LINKED ISSUE #1358: [log] replace `fmt::print` with something that prints to the log file
Maybe write a `lg::print` that does that.

xTVaser:
Favors the `lg` namespace over `fmt` directly, as this will output the logs to a file / has log levels.

I also made assertion errors go to a file, this unfortunately means importing `lg` and hence `fmt` which was attempted to be avoided before.  But I'm not sure how else to do this aspect without re-inventing the file logging.

We have a lot of commented out prints as well that we should probably cleanup at some point / switch them to trace level and default to `info` level.

I noticed the pattern of disabling debug logs behind some boolean, something to consider cleaning up in the future -- if our logs were more structured (knowing where they are coming from) then a lot this boilerplate could be eliminated.

Closes #1358


common/audio/audio_formats.cpp
```diff
@@ -1,6 +1,7 @@
 #include "audio_formats.h"
 #include "common/util/BinaryWriter.h"
 #include "third-party/fmt/core.h"
+#include <common/log/log.h>
```
water111, line 4: I think we should use `"common/log/log.h"` instead of the angle brackets in all these files. The angle brackets are usually used for system headers that need to search on some weird out-of-project path.

common/type_system/Type.cpp
```diff
@@ -193,8 +194,8 @@ std::string Type::get_name() const {
 
 std::string Type::get_runtime_name() const {
   if (!m_allow_in_runtime) {
-    fmt::print("[TypeSystem] Tried to use type {} as a runtime type, which is not allowed.\n",
-               get_name());
+    lg::error("[TypeSystem] Tried to use type {} as a runtime type, which is not allowed.\n",
```
water111, line 197: let's keep this one as a `print`, like the rest of the compiler errors.

common/type_system/defenum.cpp
```diff
@@ -71,7 +72,7 @@ EnumType* parse_defenum(const goos::Object& defenum, TypeSystem* ts) {
       } else if (symbol_string(option_value) == "#f") {
         is_bitfield = false;
       } else {
-        fmt::print("Invalid option {} to :bitfield option.\n", option_value.print());
+        lg::error("Invalid option {} to :bitfield option.\n", option_value.print());
```
water111, line 75: same here, let's keep all the ones in this file as normal prints

common/type_system/deftype.cpp
```diff
@@ -608,14 +609,14 @@ DeftypeResult parse_deftype(const goos::Object& deftype,
       new_type->set_pack(true);
     }
     if (sr.allow_misaligned) {
-      fmt::print(
+      lg::error(
```
water111, line 612: same with this file

common/util/Assert.cpp
```diff
@@ -3,21 +3,23 @@
...
-    fprintf(stderr, "Assertion failed: '%s'\n\tSource: %s:%d\n\tFunction: %s\n", expr, file, line,
-            function);
+    std::string log = fmt::format("Assertion failed: '{}'\n\tSource: {}:{}\n\tFunction: {}\n", expr,
+                                  file, line, function);
+    lg::die(log);
   } else {
-    fprintf(stderr, "Assertion failed: '%s'\n\tMessage: %s\n\tSource: %s:%d\n\tFunction: %s\n",
-            expr, msg, file, line, function);
+    std::string log =
+        fmt::format("Assertion failed: '{}'\n\tMessage: %s\n\tSource: {}:{}\n\tFunction: {}\n",
```
water111, line 19: `%s` should be `{}` here. (this is part of the reason i don't like libfmt that much... instead of being a compiler error it ends up being some confusing "format error" exception thrown that you only see at runtime)

decompiler/VuDisasm/VuDisassembler.cpp
```diff
@@ -411,15 +412,15 @@ VuProgram VuDisassembler::disassemble(void* data, int size_bytes, bool debug_pri
 
     // debug
     if (debug_print) {
-      fmt::print("{}\n", to_string(VuInstructionPair{upper_instr, lower_instr}));
+      lg::debug("{}\n", to_string(VuInstructionPair{upper_instr, lower_instr}));
```
water111, line 415: can we leave this file as normal prints too? The output is stuff that I copy and paste directly into .cpp files.

decompiler/VuDisasm/VuDisassembler.cpp
```diff
@@ -1011,17 +1012,12 @@ std::string VuDisassembler::to_cpp(const VuInstruction& instr, bool mips2c_forma
                          instr.src.at(1).to_string(m_label_names));
     default:
       unk++;
-      fmt::print("unknown 0 is {}\n", to_string(instr));
-
-      return "ASSERT(false);";  //"???";
+      return "ASSERT_MSG(false, fmt::format(\"unknown 0 is {}\\n\", to_string(instr));";  //"???";
```
water111, line 1015: I think we should revert this change

decompiler/VuDisasm/VuDisassembler.cpp
```diff
@@ -1011,17 +1012,12 @@ std::string VuDisassembler::to_cpp(const VuInstruction& instr, bool mips2c_forma
...
   }
 
 unknown:
-
   unk++;
-  fmt::print("unknown 1 is {}\n", to_string(instr));
-
-  return "ASSERT(false);";
+  return "ASSERT_MSG(false, fmt::format(\"unknown 1 is {}\\n\", to_string(instr));";  //"???";
```
water111, line 1020: and this one

decompiler/analysis/variable_naming.cpp
```diff
@@ -1068,21 +1068,21 @@ std::optional<VariableNames> run_variable_renaming(const Function& function,
...
-    fmt::print("Basic SSA\n{}\n------------------------------------\n", ssa.print());
+    lg::debug("Basic SSA\n{}\n------------------------------------\n", ssa.print());
   }
 
   // eliminate PHIs that are not needed, still keeping us in SSA.
   while (ssa.simplify()) {
   }
   if (debug_prints) {
-    fmt::print("Simplified SSA\n{}-------------------------------\n", ssa.print());
+    lg::debug("Simplified SSA\n{}-------------------------------\n", ssa.print());
```
water111, line 1085: I think this file should stay as prints - the alignment out the output will probably get messed up because the print functions here insert their own `\n`'s

game/sound/989snd/midi_handler.cpp
```diff
@@ -322,13 +323,12 @@ bool midi_handler::tick() {
...
   } catch (midi_error& e) {
     m_track_complete = true;
-    fmt::print("MIDI Error: {}\n", e.what());
-
-    fmt::print("Sequence following: ");
+    lg::error("MIDI Error: {}\n", e.what());
+    lg::error("Sequence following: ");
     for (int i = 0; i < 10; i++) {
-      fmt::print("{:x} ", m_seq_ptr[i]);
+      lg::error("{:x} ", m_seq_ptr[i]);
```
water111, line 329: I think this is going to break the formatting, maybe leave this as prints

goalc/compiler/Env.cpp
```diff
@@ -153,7 +154,7 @@ void FileEnv::add_top_level_function(std::unique_ptr<FunctionEnv> fe) {
 void FileEnv::debug_print_tl() {
   if (m_top_level_func) {
     for (auto& code : m_top_level_func->code()) {
-      fmt::print("{}\n", code->print());
+      lg::debug("{}\n", code->print());
```
water111, line 157: I think this will mess up the formatting, maybe leave as prints

goalc/regalloc/Allocator.cpp
```diff
@@ -716,21 +717,21 @@ namespace {
...
-  printf("[RegAlloc] Alive Info\n");
-  printf("-----------------------------------------------------------------\n");
+  lg::debug("[RegAlloc] Alive Info\n");
+  lg::debug("-----------------------------------------------------------------\n");
   // align to where we start putting live stuff
-  printf("      %30s    ", "");
+  lg::debug("      {:30}    ", "");
   for (int i = 0; i < cache->max_var; i++) {
-    printf("%2d ", i);
+    lg::debug("{} ", i);
```
water111, line 731: this is going to mess up the formatting, let's leave these as prints

goalc/compiler/compilation/Debug.cpp
```diff
@@ -201,18 +202,18 @@ void mem_print(T* data, int count, u32 start_addr, PrintMode mode) {
   for (int i = 0; i < count; i++) {
     if ((i % elt_per_line) == 0) {
       // first in line, so we should print the GOAL address
-      fmt::print(" 0x{:08x}: ", start_addr + (i * sizeof(T)));
+      lg::info(" 0x{:08x}: ", start_addr + (i * sizeof(T)));
```
water111, line 205: this one is also going to mess up the formatting

goalc/compiler/compilation/Debug.cpp
```diff
@@ -361,8 +362,8 @@ Val* Compiler::compile_disasm(const goos::Object& form, const goos::Object& rest
                          addr, addr + size, EE_MAIN_MEM_LOW_PROTECT, EE_MAIN_MEM_SIZE);
   }
 
-  fmt::print("{}\n", m_debugger.get_info_about_addr(addr));
-  fmt::print("{}\n", m_debugger.disassemble_x86_with_symbols(size, addr));
+  lg::info("{}\n", m_debugger.get_info_about_addr(addr));
```
water111, line 365: same here

water111 (commented):
That's a lot of prints, thanks for doing this.

I'm not totally convinced we want to use the log stuff for normal compiler output (syntax errors, warnings, status prints...).   The error messages are dense and confusing to start with, and having a bunch of `[2022-05-21 20:02:18:661] [debug]` everywhere will make it harder to read.  I think it would be better if we stuck to normal prints for the `goalc/` folder.

I quickly went through all of them and commented on a number of places where the log prefix and auto newline stuff will mess up the formatting.

ManDude:
Yeah, the timestamp thing is partly why I suggested a `lg::print` instead of changing everything to `lg::debug` and stuff. It's cluttery.

xTVaser:
imo the vast majority of the prints I replaced should be using a lg::debug/info/error/etc so they actually have some context / can be disabled in a somewhat decent way.

But yeah there were some, as mentioned above, that should stay as a normal print (or be refactored more in-depth).

ManDude:
I think we can condense them a little bit. I do not believe the milliseconds are useful, for example.

decompiler/VuDisasm/VuDisassembler.cpp
```diff
@@ -5,6 +5,7 @@
 #include "third-party/fmt/core.h"
 #include "common/util/print_float.h"
 #include "common/util/Assert.h"
+#include "common/log/log.h"
```
water111, line 8: not needed?

water111: (also true in a few other files)

decompiler/analysis/insert_lets.cpp
```diff
@@ -84,10 +84,8 @@ Form* lca_form(Form* a, Form* b, const Env& env) {
     ai--;
     bi--;
   }
-  if (!result) {
-    fmt::print("{} bad form is {}\n\n{}\n", env.func->name(), a->to_string(env), b->to_string(env));
-  }
-  ASSERT(result);
+  ASSERT_MSG(result, fmt::format("{} bad form is {}\n\n{}\n", env.func->name(), a->to_string(env),
```
water111, line 87: The one thing I don't like about this `ASSERT_MSG` pattern is that if you mess up the format string in `fmt::format`, it throws can exception and doesn't end up asserting (and just looks like one of the many other decompiler errors).  I am pretty tempted to modify libfmt so it just asserts in this case.  But we can worry about that later

game/kernel/kmemcard.cpp
```diff
@@ -58,11 +59,11 @@ using namespace ee;
 template <typename... Args>
 void mc_print(const std::string& str, Args&&... args) {
   if (memcard_debug) {
-    fmt::print("[MC] ");
+    lg::info("[MC] ");
```
water111, line 62: will this have issues with newlines? Does this print even get used?

goalc/compiler/Compiler.cpp
```diff
@@ -9,6 +9,7 @@
 #include "goalc/regalloc/Allocator_v2.h"
 #include "third-party/fmt/core.h"
 #include "common/goos/PrettyPrinter.h"
+#include "common/log/log.h"
```
water111, line 12: not needed?

goalc/debugger/Debugger.cpp
```diff
@@ -13,6 +13,7 @@
 #include "goalc/listener/Listener.h"
 #include "third-party/fmt/core.h"
 #include "common/util/Assert.h"
+#include "common/log/log.h"
```
water111, line 16: A bunch of these aren't needed

water111 (commented):
Looks good, just a few more. It might be worth a cleanup to remove the includes where they aren't needed - these little things do add up over time to make the build faster.

decompiler/VuDisasm/VuDisassembler.cpp
xTVaser: I wonder if we can eventually wire this up to our CI process https://github.com/include-what-you-use/include-what-you-use

common/log/log.cpp
```diff
@@ -44,10 +44,14 @@ void log_message(level log_level, LogTime& now, const char* message) {
   auto now_milliseconds = now.tv.tv_usec / 1000;
   strftime(date_time_buffer, 128, "%Y-%m-%d %H:%M:%S", localtime(&now_seconds));
   std::string date_string = fmt::format("[{}:{:03d}]", date_time_buffer, now_milliseconds);
+  strftime(date_time_buffer, 128, "%M:%S", localtime(&now_seconds));
+  std::string time_condensed = fmt::format("[{}]", date_time_buffer);
```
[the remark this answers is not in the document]
ManDude: We could get rid of the minutes and just have `seconds.milliseconds` like PCSX2.

game/kernel/kmemcard.cpp
ManDude: Not anymore since `memcard_debug` is turned off, but it does mess up newlines. It's meant to look lime `[MC] message goes here`

PR #1369, opened 2022-05-21
fix misc bugs preventing the game from starting

water111:
- re-enable `dump_objs` because we need this for the extractor to work
- `(when *debug-segment*` guard for anim tester to prevent crash
- remove duplicate copy of googletest included in cubeb
- link against stdc++ for filesystem stuff
- use `_mm_storel_epi64` instead of `_mm_storeu_si64` because some older versions of gcc/clang are missing it (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87558)


ManDude:
I feel like `(declare-file (debug))` should automatically wrap stuff in a `(when *debug-segment*`

PR #1372, opened 2022-05-23
[decomp] add `import` statements for art groups

water111:
Changes:
- temporarily make `import` a macro that calls `asm-file`. This will be replaced with a real dependency system in the future
- remove `all_imports.gc`
- add `import` statements to files in `goal_src` for their skelgroups
- add feature to put `import` statements in the decompiler output
- create JSON file with `import`s for the decompiler for the art groups


ManDude (approved):
no more slow compiler startup speeds 👍 (again)

PR #1374, opened 2022-05-28
[graphics] merc2 renderer

water111:


water111:
Taking a look at a merc-heavy area (currently 6.2 ms total, 1.5 ms merc)
image

The number of bones isn't great - each involves sending a bone matrix (112 bytes) to the GPU.

water111:
> Is there a way to send a large uniform buffer with a bunch of bones?

Yes, and I'm planning to switch to that. It's not clear that all GPUs will even let us use such a large number of uniforms without using a uniform buffer.

There's also a lot we can do to avoid sending bones that aren't used:
- right now, I send 128 bones every time, but most models don't have that many. The uniform buffers would let us update only part of the buffer if a model doesn't need that many bones.
- some bones are unused, so we can renumber the bones to eliminate these.

water111:
seems promising:
image

From ~1.5 to ~1.1ms for merc here.

PR #1375, opened 2022-05-28
fix float->int cast suppression + minor cleanup of a couple files

ManDude:


water111 (approved):
thanks for fixing this!

PR #1378, opened 2022-05-29
Loader improvements

water111:


water111:
This should reduce (but not totally solve) dropped frames when loading, fixes some bug where we uploaded the same data multiple times, and adjusts the iop kernel dispatch stuff so it doesn't spam running the iop when there's nothing to do.

PR #1385, opened 2022-05-30
[graphics] merc2 fix, texture crash bug fix

water111:
This fixes a bug with merc2 where the alignment of the bones within the uniform buffer was a bit weird. I added some padding to the bones themselves and made sure to follow the exact alignment required by opengl.  In the end, we actually save memory compared to the original layout because we can insert not-multiple-of-bone gaps in the buffer to get alignment. There was an off by one that in rare cases could cause a bone to be skipped as well.

Also fixes crashes when turning off textures in the debug menu.


water111:
also a fix to let the game keep booting even if deci2 stuff fails to init (like due to a networking issues)

PR #1387, opened 2022-05-30
W/custom string translations

Zedb0T:
This adds translation support for the custom strings file. All of these were done via google translate so some translations may not be ideal but it is still more clear than "UNKOWN ID 4216UNKOWN ID 4216:"

I have had both the French and German translations reviewed and improved upon by native speakers and have reached out to have the Spanish verified and improved. While this isn't perfect yet its a good more clear baseline as I reach out to have the other languages verified.

I did not do Japanese as my understanding is that automatic English to Japanese translation is VERY rough and we havn't had any requests for it as of yet. Instead all custom strings read "NEED JP TRANSLATION" So it very clear if a native speaker would be able to contribute.

https://media.discordapp.net/attachments/691768471482007632/980819135187517460/unknown.png

https://cdn.discordapp.com/attachments/691768471482007632/980871350669418546/unknown.png


ManDude:
IMO it's much better to split the files by language rather than have them all in one place. It's messy, and it's very rare any single person cares about all of the languages. Not to mention there will be extra files regardless for when we add custom languages.

Zedb0T:
> 

I could split these, however I thought it was more beneficial to have them all in a single file. This way if there needs to be a change, a bilingual user could reference the English directly above to provide a direct translation instead of cross referencing the string ID. (I guess they could also compare it in game?? Still seems suboptimal to me.)

On the tech side I am unsure of the pros and cons but from an organizational standpoint I do prefer the single file approach.
Let me know if you'd like me to split them.

Zedb0T:
I actually thought having a single file for each language was a messier approach, I just assumed it has some kind of tech benefit or avoids a technical limitation.

ManDude:
> This way if there needs to be a change, a bilingual user could reference the English directly above to provide a direct translation instead of cross referencing the string ID.

You can just open the files side-by-side if you need to do this, or copy-paste and then change.

The single file will get a LOT messier when you add more languages. We can support at least 5 more, and having each ID have 11 strings in them bloats the file up considerably, and if you want to know what language a string is for you need to count the strings manually.

Zedb0T:
Yeah thats fair, Ill split the file.

ManDude:
Another (arguably more important) reason to keep files split: merge conflicts.

xTVaser:
Yeah splitting the files is a good idea -- though this conversation brings up how brutal the current translation setup is (or in other words, how there isn't one).

Editing in text files is serviceable for now (better than nothing), but when handing it to translators it would be much better to have a simple GUI to tie everything together / edit the files.  The same goes for subtitles (I have code in progress for this atleast)

Issue #1401, opened 2022-06-01
[decompiler] let rewrite stats are sometimes gibberish


ManDude:
???

PR #1403, opened 2022-06-01
hook up merc2 to fog intensity, flip some min/max fogs

water111:
now the orbs in boggy look right with increased fog


water111:
also fix https://github.com/open-goal/jak-project/issues/1401

PR #1409, opened 2022-06-02
[game] fix daxter stuck face bug

water111:
fix daxter's face getting stuck when exiting a cutscene or similar.

This happened in the real game, but we handle the messed up state at the end improperly with merc2, which assumes that the blerc/blerc-done flags are used correctly.


water111:
I think the failed test is because there's some code printing some stuff in the middle of the tests...

PR #1416, opened 2022-06-04
[game] music player

### LINKED ISSUE #1414: [progress-pc] scroll arrows are hardcoded
bah!

ManDude:
Adds a "secrets" menu to the title screen, similar to other Jak games. The secrets menu currently only contains this music player. It lets you play any music track in the game, along with any of its flavored variations.

Tracks are unlocked when they are heard in-game for the first time. Some tracks are also unlocked by completing certain tasks to prevent softlocks or because they're unused.

image

image

There is a known bug with OVERLORD that's also in the original game where music tracks won't be loaded while a spool is running, so if you change the track from the title screen's theme (village1) and then unpause or leave the menu, you won't get music in the title screen until you pause the game and unpause. I am not gonna fix this right now though.

This bumps the pc-settings version, so your settings will be reset.

Also fixes #1414 and allows boxed-array static defs to omit both `length` and `allocated-length` at the same time (previously this was an error).


goal_src/engine/game/settings-h.gc
```diff
@@ -60,49 +60,6 @@
     )
   )
 
-(defmethod inspect setting-data ((obj setting-data))
```
water111, line 63: this one actually has some changes - the `movie` has a `ppoitner->process` in it.

goal_src/goal-lib.gc
```diff
@@ -670,6 +670,11 @@
       )
   )
 
+(defmacro float->int (a)
+  "forcefully casts something as a float to int. be careful."
+  `(the int (the float ,a))
```
water111, line 675: maybe should be `(the int (the-as float ,a))` in case you start with a `uint32` or something like that?

goal_src/engine/game/settings-h.gc
ManDude: oh oops, i didn't see that.

goal_src/goal-lib.gc
ManDude: actually, I think I still want this to work properly if you put in an int.

PR #1420, opened 2022-06-05
[decomp] clean up KERNEL.CGO code

water111:
Clean up some old documentation, use `link-flags` in a few more places.


goal_src/kernel/gkernel.gc
```diff
@@ -1134,6 +1137,9 @@
     )
   )
 
+;; this will be set to #t if we're using visibility data.
+;; if we aren't, there will be many cases where we try to spawn too many actors, and we shouldn't
+;; warn when this ahppens.
```
ManDude, line 1142: typo!

PR #1421, opened 2022-06-05
[discord-rpc] fix status not being updated while indoors

Hat-Kid:
When in an indoors area, the small time of day icon and text were not being cleared properly and would be filled with garbage data, which resulted in the game status not being updated until you left the area.


water111:
good catch, thanks for the quick fix!

PR #1449, opened 2022-06-13
[game] fix shadow offset in widescreen

water111:
Fixes a bug in the original font renderer where the font shadow offset changed depending on the aspect ratio. It's particularly noticeable with smaller fonts and very wide screens.


water111:
before
image

after
image

(this is with a stupidly wide aspect ratio, just to make it obvious)

PR #1460, opened 2022-06-16
Update `game` directory description (we do have audio)

ManDude:


connorjclark:
FYI missed a spot https://github.com/open-goal/jak-project#:~:text=There%20is%20currently%20no%20audio.

ManDude:
oops.

Next time I'll just ctrl+f instead of using sourcegraph I guess.

PR #1463, opened 2022-06-17
release: include error metadata file for launcher purposes

xTVaser:
This adds a simple metadata file to the release process that lets the launcher know what error codes mean.  This keeps this repo as the source of truth instead of having to make assumptions in the launcher (and hoping nothing changes).

Also got rid of the `out/` subfolder inside the release assets

### DIFF
--- a/.github/workflows/linux-workflow.yaml
+++ b/.github/workflows/linux-workflow.yaml
@@ -114,9 +114,8 @@ jobs:
         run: |
           mkdir -p ./ci-artifacts/out
           ./.github/scripts/releases/extract_build_linux.sh ./ci-artifacts/out ./
-          pushd ci-artifacts
-          tar czf linux.tar.gz -C ./out .
-          popd
+          cd ci-artifacts/out
+          tar czf ../linux.tar.gz .
 
       - name: Upload Assets and Potential Publish Release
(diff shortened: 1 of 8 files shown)


.github/workflows/linux-workflow.yaml
```diff
@@ -115,7 +115,7 @@ jobs:
           mkdir -p ./ci-artifacts/out
           ./.github/scripts/releases/extract_build_linux.sh ./ci-artifacts/out ./
           pushd ci-artifacts
-          tar czf linux.tar.gz ./out
+          tar czf linux.tar.gz -C ./out .
```
water111, line 118: I forget what happened here... maybe this got changed and then reverted?
There was some issue with `tar` seeing its own output as it was going?

water111 (approved):
:+1:
Feel free to merge when you're ready.

.github/workflows/linux-workflow.yaml
xTVaser: It's a confusing history, I had something somewhat similar (still contained the out folder) https://github.com/open-goal/jak-project/pull/1326/files

And then it was changed slightly here https://github.com/open-goal/jak-project/commit/e8911758f0d07d70b4be2cbfc3ee052e4e263ad9

the `pushd/popd` is basically manually doing what `-C ...` is supposed to do.  I havn't had a problem on my machine but i can switch to that style if it's more consistent?

PR #1465, opened 2022-06-17
a bunch of PC/GFX system fixes

### LINKED ISSUE #1153: [runtime] can't leave borderless properly sometimes
Window stays undecorated for some reason.

### LINKED ISSUE #1456: [runtime] make fullscreen settings actually work properly
This is such a pain to get working.

### LINKED ISSUE #1459: [runtime] refactor display and pipeline code
I'm not sure why, but originally I didn't opt for using abstract classes for this stuff. At this point though I think it'd be way better to rewrite all that code (it's not that much) to use an abstract class for the renderer and displays.

### LINKED ISSUE #1464: [game] split game resolution and window size settings params
a.k.a. do it properly.

ManDude:
Moves a lot of useless boilerplate to virtual methods, reworks a bunch of PC graphics code so that the game doesn't launch with the wrong window size.

- Fixes #1459 
- Fixes #1464 
- Fixes #1456
- Fixes #1153 (hopefully)

### DIFF
--- a/game/graphics/display.cpp
+++ b/game/graphics/display.cpp
@@ -16,10 +16,6 @@
 
 namespace {
 
-bool renderer_is_correct(const GfxRendererModule* renderer, GfxPipeline pipeline) {
-  return renderer->pipeline == pipeline;
-}
-
 void set_main_display(std::shared_ptr<GfxDisplay> display) {
   if (Display::g_displays.size() > 0) {
     Display::g_displays[0] = display;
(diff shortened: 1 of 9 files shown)


water111 (approved):
Thanks for cleaning this up!

I like this approach and I wonder if we can eventually get rid of `GfxRendererModule` in a similar way (either merge with `GfxDisplay` or make another class).

PR #1466, opened 2022-06-17
Update pc-pad-utils.gc

towai:
It got a bit left behind at some point, the file couldn't be injected into the game anymore because debug-draw1 got renamed. Also the second pad's binds went offscreen in the display function, so I altered it somewhat.
Old (or, an approximation - the indentation is different and i don't think it actually has any leading newlines at all, but i just slightly edited the new version back to get the gist across):
image
New:
image
The new one is also less likely to get covered by other debug text that happens to be up (i.e load boundary info)
Also added a (hopefully) helpful message in the repl pointing users who might not be aware of GLFW to the page that contains its keycode constants (yeah, i know the project ~~might~~ probably will at some point switch to something else for input but that's a future problem)


water111 (approved):
thanks for keeping this up to date. Eventually we should have some system to make sure the `pc_debug` stuff can still build... one day...

PR #1472, opened 2022-06-17
Add menu bar toggle

SortAnon:
- Toggle the ImGui menu bar by pressing Alt
- Hides the menu bar by default in fullscreen and borderless modes

### DIFF
--- a/game/graphics/display.cpp
+++ b/game/graphics/display.cpp
@@ -92,6 +92,7 @@ int InitMainDisplay(int width, int height, const char* title, GfxSettings& setti
     lg::error("Failed to make main display.");
     return 1;
   }
+  display->set_imgui_visible(true);
   set_main_display(display);
   return 0;
 }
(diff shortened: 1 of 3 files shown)


ManDude:
Jeez people really hate the damn bar...

xTVaser:
This also triggers when switching windows via Alt+Tab which i could see being annoying while developing (possibly also during normal usage) -- I don't really have a better key suggestion in mind.

NickDriscoll:
Why not just ESC?

Zedb0T:
ESC by default is already pound to 1st controller start

NickDriscoll:
Huh. For me it's ENTER.

Zedb0T:
Ah maybe that got changed when they updated the keybindings

SortAnon:
> This also triggers when switching windows via Alt+Tab
Good point. I've changed it to trigger on release if the window is focused. That works on KDE, and likely Windows as well.

xTVaser (approved):

PR #1473, opened 2022-06-18
actions: add `libpulse-dev` dependency to fix audio on pipewire systems

Hat-Kid:
On the Linux release version, `cubeb` would fail to init on systems that use Pipewire as the audio backend due to a missing build dependency.


xTVaser:
Confirmed to work by you.  Changes are only to the linux build so no need to wait for the windows build to finish.

Merging so it can be included in the release im about to cut

r2rX:
Can also confirm audio works now on Fedora 36. Thanks!

PR #1475, opened 2022-06-18
Binding display readability improvements + add example showing font-color constants

towai:
`pc-pad-show`: Much better. Thanks to Dass for pointing out you can monospace the font with `~0k` (note: it actually has slightly reduced spacing now compared to this screenshot)
Text aligned to soothe the autistic mind (mine)")
Water also said I could put the color constants script I made up in `goal_src/examples`, so I have.
I was gonna also cheekily update the `debug-collide` example, since i noticed that's the only other place `debug-draw1` appears, but that line is actually commented out, so I'll leave it be


towai:
wait a second, i opened the game again and the color constants script won't make now. i get "type handle is unknown" at `(define *color-display-handle* (new 'static 'handle))`. i don't think i changed it from when it was working just fine but i'm trying to figure this out

ManDude:
You need to compile the kernel first.

towai:
> 

🤦‍♀️

PR #1481, opened 2022-06-18
[custom levels] add 3rdparty library for gltf files

water111:


water111:
the stub `build_level` program is included to force gltf to be built, to make sure it works on windows.

PR #1490, opened 2022-06-19
PAL & NTSC-J support + updates

### LINKED ISSUE #1484: figure out how to handle altered types across different versions of the game
There are a few types and functions that were changed from PAL onwards compared to NTSC-U black label that fix a bunch of small to not-so-small bugs. While fixing them isn't a big deal, it DOES break the type consistency check and ref tests a bit. Differences include `target` and `control-info` having extra fields (type size will mismatch) and `race-time-read` taking a different amount of arguments (breaks decompilation of functions that call it and also the function itself).

### LINKED ISSUE #1497: [decompiler] some sort of interface to binary patch specific files

ManDude:
Makes PAL & NTSC-J decompileable and updates our source to have the PAL & NTSC-J fixes.

- Fixes #1497 
- Fixes #1484 

Also fixes the extractor using a hardcoded `jak1` path by changing it to a `_temp` path that's less likely to cause issues, and made it wipe the contents of it whenever new stuff needs to be extracted to avoid amalgamation issues. Also added some text that shows files being extracted so that people don't close the program thinking it's hanging.


ManDude:
Added a patching interface so we can "use" PAL files without hosting their actual contents on the repo.

water111:
Can we split this up into multiple PRs? It's kinda hard to review unrelated changes with 48,000 lines changed across 181 files.

ManDude:
Stuff has been split now.

game/main.cpp
```diff
@@ -81,7 +81,9 @@ int main(int argc, char** argv) {
 
 #ifndef __AVX2__
   if (get_cpu_info().has_avx2) {
-    printf("Note: your CPU supports AVX2, but this build was not compiled with AVX2 support\n");
+    printf(
```
water111, line 84: let's just get rid of this print. Nobody cares about the tiny 2% performance increase from the AVX2 version of time of day interpolation anyway.

test/decompiler/reference/levels/rolling/rolling-obs_REF.gc
```diff
@@ -872,27 +871,40 @@
...
 ;; definition for function race-time-save
-(defun race-time-save ((arg0 race-time) (arg1 int) (arg2 task-control))
-  (dotimes (s3-0 5)
-    (save-reminder arg2 (-> arg0 digit s3-0) (+ arg1 s3-0))
-    )
-  #f
+;; WARN: Type Propagation failed: Failed type prop at op 6 ((set! v1 (l.wu (+ a0 -4)))): Could not get type of load: (set! v1 (l.wu (+ a0 -4))). 
+;; WARN: Type Propagation failed: Type analysis failed
+;; WARN: Function may read a register that is not set: a2
+(defun race-time-save ((a0-0 race-time) (a1-0 task-control))
```
water111, line 877: this one looks kinda broken. If we want all these patches, I wonder if we should just switch to decompiling the PAL version. Having all these combined versions is kinda confusing

decompiler/config/jak1_jp/label_types.jsonc
```diff
@@ -0,0 +1,2152 @@
+// the format is [label_name, type, special]
```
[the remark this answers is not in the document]
ManDude: At the bottom from `BEGIN` onwards is the changes I made compared to the ntsc-u one. Compared to PAL I think it's just `progress` that's slightly different.

test/decompiler/reference/levels/rolling/rolling-obs_REF.gc
ManDude: Yeah I wouldn't mind changing the ref tests to use PAL or something after that.

decompiler/extractor/main.cpp
```diff
@@ -357,7 +371,11 @@ int main(int argc, char** argv) {
     setup_global_decompiler_stuff(std::nullopt);
   }
 
-  std::filesystem::path path_to_iso_files = file_util::get_jak_project_dir() / "iso_data" / "jak1";
+  std::filesystem::path path_to_iso_files = file_util::get_jak_project_dir() / "iso_data" / "_temp";
+  if (flag_extract && data_dir_path != path_to_iso_files) {
+    // in case input is also output, don't just wipe everything (weird)
+    std::filesystem::remove_all(path_to_iso_files);
```
xTVaser, line 377: I think this and the `create_directories` call below should just be moved into the `flag_extract` check below

decompiler/extractor/main.cpp
```diff
@@ -16,7 +16,8 @@ enum class ExtractorErrorCode {
   VALIDATION_ELF_MISSING_FROM_DB = 4002,
   VALIDATION_BAD_ISO_CONTENTS = 4010,
   VALIDATION_INCORRECT_EXTRACTION_COUNT = 4011,
-  VALIDATION_BAD_EXTRACTION = 4020
+  VALIDATION_BAD_EXTRACTION = 4020,
+  VALIDATION_BAD_DECOMP = 4021
```
xTVaser, line 20: I would change this to:
`DECOMPILATION_GENERIC_ERROR` or something.

You also need to define a useful string here https://github.com/open-goal/jak-project/blob/master/.github/scripts/releases/error-code-metadata.json

xTVaser (changes requested):
extractor changes look mostly good, nothing that would break anything but some minor tweaks needed

ManDude:
Build system is hardcoded to look into `jak1`, will have to find a fix for this first.

xTVaser (approved):
looks good extractor wise

water111 (approved):
:+1: feel free to merge when you're ready

ManDude:
ok turns out that isn't the case so it's all good

PR #1495, opened 2022-06-19
Useful texture replacement example.

nameonascreen:
Creates a texture_replacements folder & replaces the default "stamp.png" with the open goal logo.

Useful as it provides users with the texture_replacements folder, & puts the open goal logo in placement of something very random.


ManDude:
This is cool, but it will ALWAYS overwrite the original texture, which I would prefer not to happen when people just wanna play the original game. (I also don't want to mess with any of the pre-existing logos and legal splashes).

nameonascreen:
> This is cool, but it will ALWAYS overwrite the original texture, which I would prefer not to happen when people just wanna play the original game. (I also don't want to mess with any of the pre-existing logos and legal splashes).

IIRC, the stamp wasn't even a thing in the original game... (maybe i'm remembering wrong)
I don't really see a problem with it "ALWAYS overwriting" the original texture, simply because it feels out of place already.

ManDude:
The stamp has always been there :) https://www.youtube.com/watch?v=9lpq3zVXs9A

nameonascreen:
> The stamp has always been there :) https://www.youtube.com/watch?v=9lpq3zVXs9A

just looked for it, you're right lol. (spoke too soon). Still, the stamp doesn't appear anywhere else in the game. Also would be nice in general to just have the texture_replacements folder already included with the data, even if it's just empty

xTVaser:
Perhaps just improve the documentation and include this as an example?

The existing markdown doc https://github.com/open-goal/jak-project/blob/master/docs/texture_replacement.md could be improved and added it to the website https://github.com/open-goal/open-goal.github.io/tree/master/docs/docs

Issue #1492, opened 2022-06-19
150FPS slow motion


nameonascreen:
### Describe the Bug

Prolly more ways to reproduce this, but after roll jumping jak will go into a slow motion state

### How To Reproduce

Set Framelimiter to 150FPS
Roll Jump and continue moving in any direction

### Expected Behavior

after a roll jump jak should be moving at his normal speed

### Operating System

Windows 11

### OpenGOAL Version

Release

### Game Version

NTSC 1.0 (black label)

Brent-Hickey:
Target.gc converts jak's velocity using (-> *display* seconds-per-frame). I've noticed that if I adjust seconds-per-frame to match the 60fps version this problem goes away. I'm checking if it breaks other stuff 

The roll jump walk squish in particular is from this area. unknown-float01 is jak's xz velocity
```(until (ja-done? 0)
                 (suspend)
                 (ja :num! (seek! max (/ (* (fmax 20480.0 (-> self control unknown-float01)) (-> *display* seconds-per-frame))
                                        (/ (-> *TARGET-bank* run-up-cycle-dist) (-> *TARGET-bank* run-cycle-length))
                                        )
                                 )
                     )
                 )```

Issue #1499, opened 2022-06-20
[Graphics] 150FPS speedup Mega Issue


BreakPoints:
### Describe the Bug

After the 150FPS patch I've spotted a few aspects of the game being sped up and while there is already an issue open for one such problem (#1492) I think a mega issue where we can all contribute what we've spotted would be useful.

Aspects currently sped up in >60FPS modes:

- [ ] Character eyes blinking outside of cutscenes
- [ ] Pause menu animations (flashing arrows, power cell animations)

### How To Reproduce

Boot game on a high refresh monitor and observe listed issues.

### Expected Behavior

For listed aspects of the game to maintain the same speed as if it were running at 60FPS.

### Operating System

Windows 10 21H2 (shouldn't be OS specific)

### OpenGOAL Version

Release

### Game Version

NTSC 1.0 (black label)

SirTobyGirl:
When climbing Snowy Moutain the gondala goes up in 3x speed.

BreakPoints:
> When climbing Snowy Moutain the gondala goes up in 3x speed.

Sounds like a feature to me!

SirTobyGirl:
> > When climbing Snowy Moutain the gondala goes up in 3x speed.
> 
> Sounds like a feature to me!

Same effect happens on the boad to misty island, I suggest renaming the boat to jet XD

water111:
https://github.com/open-goal/jak-project/issues/1524 probably both 150fps issues

BreakPoints:
Now that we have a high fps tag, wondering if there's use in keeping this mega issue updated with new findings. The idea was to centralise these issues but most are creating their own issues and to add those here would be duplication.

PTKickass:
The Blue Eco launch pad falling animation plays at the wrong speed (too slowly) when running at 150fps. The camera rotation speed when falling is also way too fast when compared with 60fps. Tested this in Boggy Swamp if that matters.

NolanHewitt:
I think the zoomed still has some 150fps issues, turning still feels slower to me.

BreakPoints:
I've been trying to run two instances simultaneously for direct comparisons because they both take the same controllers input but something's causing them to diverge. Would have been useful to have 60FPS side by side with 150FPS to spot differences with input.

Current divergence happens with both running identical settings so the controller polling must be nondeterministic or they're stepping on each others toes.

ManDude:
It could very well be lag.

BreakPoints:
[...]

PR #1503, opened 2022-06-20
[game] 150fps fixes

Brent-Hickey:
This fixes https://github.com/open-goal/jak-project/issues/1492
And it addresses some of https://github.com/open-goal/jak-project/issues/1499

Unfortunately the 150fps issues seem to all be case-by-case fixes (so its helpful for people to keep tracking them). They either have an animation that needs adjusted, did frame-based math, or relied on the seconds-per-frame as part of math where it probably wasn't supposed to be used. Usually when it related to animations + velocities. 

- Slow walk after roll has been fixed, I don't think it was supposed to look at seconds-per-frame here because the velocity it checks is already scaled
- Jak walking animation I removed the seconds-per-frame and I think it's correct for a similar reason. Are any other jak animations feeling sped up/slowed down?
- Eye blink was counting a certain number of frames so that's fixed
- Had the math wrong for the physics tick (ocean + scrollable textures) making them slightly slower than they should be
- The misty boat is a spooled animation that needed slowed so that's fixed. It also has that table of velocities for controlling it that are probably wrong but i wont worry about that 
- Same thing for gondola ride, should be fixed
- Zoomer has a lot of animations and velocities tied to seconds-per-frame so it doesn't surprise me that something is off here. But I don't know enough to determine which ones are even slow cause the zoomer generally feels good to me so i'll need someone to really tell me what's wrong about the zoomer


Zedb0T:
Pretty sure the cooldown between going into goggles is way faster. Also the aiming when you enter the goggles

Hat-Kid:
Similar to the roll bug, I've observed a slow walk after a normal jump, but not always. That may be fixed already if the roll bug is fixed.

I've also noticed of Jak's idle animations was slowed down considerably.

Brent-Hickey:
Which idle animation? I had to hard code each idle animation by name to be slowed down, it's possible I'm slowing one down that I shouldn't

Zedb0T:
The wind on jak seems sped up

Hat-Kid:
> Yep slowness after normal jumps is also fixed here. Which idle animation? I had to hard code each idle animation by name to be slowed down, it's possible I'm slowing one down that I shouldn't

It was the one with Daxter balancing on Jak's head, but I just noticed that it was only slowed when I locked the framerate to 60, it plays fine in 150 fps.

NolanHewitt:
The Eco Meter (when eco is running low) seems to flash too fast at 150 fps.

water111 (approved):
Looks good! If there end up being a lot of these fixes, you might want to make some macro for `(or (= (-> *setting-control* current video-mode) '150fps) (= (-> *setting-control* current video-mode) '100fps)`. Maybe something like `(high-fps-mode?)`.

PR #1515, opened 2022-06-22
[cleanup] memory bugs, memory usage reduction, delete merc1

water111:
Fixes:
- should fix crash on reboot in debug mode
- fixes reading `char*` buffer of a `std::string` after it goes out of scope (didn't cause issues, just asan complaining)
- removed old merc1 renderer (didn't really work right with merc2-style clipping, slower, and not more accurate)
- reduce insanely large generic renderer buffers, add some stats to see how much we're using (still only like 5%... could go smaller). Memory usage down to about 600 MB on linux.
- change some logging settings in compiler so `info` prints from `build_level` show up (likely to be useful for modding work). nothing other than custom levels uses `lg::` stuff in the compiler.


goalc/main.cpp
```diff
@@ -12,17 +12,14 @@
...
-    lg::set_file_level(lg::level::warn);
-    lg::set_stdout_level(lg::level::warn);
-    lg::set_flush_level(lg::level::warn);
-  }
+  lg::set_file_level(lg::level::info);
+  lg::set_stdout_level(lg::level::info);
+  lg::set_flush_level(lg::level::info);
+  lg::set_file_level(lg::level::warn);
+  lg::set_stdout_level(lg::level::warn);
+  lg::set_flush_level(lg::level::warn);
```
ManDude, line 22: What does calling `set_file_level` etc twice do?

water111: oops, it's just supposed to set it to `info`.

ManDude: nevermind.

PR #1517, opened 2022-06-22
lint: add include sorting config to clang-format

xTVaser:
I got inspired looking at some of my ugly include blocks.  I expected this to cause a lot of problems but everything seems to work ~~perfectly fine (i was mostly worried about winsock2 stuff)~~ indeed, the windows includes were the only problem.


xTVaser:
🤦 https://stackoverflow.com/a/70423406

PR #1522, opened 2022-06-23
[custom levels] fix various collision bugs

water111:


water111:
After this change is made, you may need to do `(make-group "iso" :force #t)` to force it to rebuild levels with the fixed tools.

PR #1525, opened 2022-06-23
full ntsc-j mode + greatest hits support

ManDude:
Fixes windows `gk` batch scripts and lets the game act like it's NTSC-J if the game language is set to Japanese (and the game copy is not black label).

Also removed the STR files from the inputs because we do not use them. Speeds up the decompiler startup a bit.


decompiler/extractor/main.cpp
```diff
@@ -313,6 +325,17 @@ void compile(std::filesystem::path extracted_iso_path) {
...
+  auto buildinfo_path = (extracted_iso_path / "buildinfo.json").string();
+  auto bi = parse_commented_json(file_util::read_text_file(buildinfo_path), buildinfo_path);
+  auto all_flags = bi.at("flags").get<std::vector<std::string>>();
+  int flags = 0;
+  for (const auto& n : all_flags) {
+    if (auto it = sGameIsoFlagNames.find(n); it != sGameIsoFlagNames.end()) {
+      flags |= it->second;
+    }
+  }
+  compiler.make_system().set_constant("*jak1-full-game*", flags & FLAG_JAK1_BLACK_LABEL);
```
water111, line 337: isn't this swapped? If we have black label, we don't have jak1-full-game?

decompiler/config/jak1_ntsc_black_label/inputs.jsonc
```diff
@@ -245,7 +245,7 @@
     "STR/GRSOPREB.STR",
     "STR/GRSOBBB.STR",
     "STR/SA3INTRO.STR"
-  ],
+    */],
```
water111, line 248: thanks

game/overlord/soundcommon.cpp
```diff
@@ -17,6 +17,9 @@ void ReadBankSoundInfo(SoundBank* bank, SoundBank* unk, s32 unk2) {
 }
 
 void PrintBankInfo(SoundBank* bank) {
+  // we dont need this and it spams the console too much
+  return;
```
water111, line 21: 👍

decompiler/extractor/main.cpp
ManDude: oops

decompiler/extractor/main.cpp
```diff
@@ -407,6 +432,25 @@ int main(int argc, char** argv) {
...
+        if (first) {
+          first = false;
+        } else {
+          buildinfo_json += ", ";
+        }
+        buildinfo_json += fmt::format("\"{}\"", n);
+      }
+    }
+    buildinfo_json += "]\n";
+    buildinfo_json += "}\n";
```
water111, line 451: the json library can write json too. You could do something like:
```
nlohmann::json json;
auto& flags = json["flags"];
flags.push_back("foo");
...
json.dump();
```
if you want. (I haven't actually checked if this compiles)

decompiler/extractor/main.cpp
```diff
@@ -407,6 +432,25 @@ int main(int argc, char** argv) {
...
+    // something tells me a ps2 game is unlikely to have a json file in root
+    file_util::write_text_file((path_to_iso_files / "buildinfo.json").string(), buildinfo_json);
```
water111, line 453: Is the use for this JSON file to let you run the extractor without unpacking the ISO, and figure out which game version it is?

water111 (approved):

decompiler/extractor/main.cpp
ManDude: yeah

decompiler/extractor/main.cpp
xTVaser: yeah, let's avoid hand-crafting JSON strings

Hat-Kid: This should produce `[]`:
```
nlohmann::json json = nlohmann::json::array();
```

ManDude: I'm not sure how to add that to an existing json object

decompiler/extractor/main.cpp
```diff
@@ -23,12 +26,18 @@ enum class ExtractorErrorCode {
   DECOMPILATION_GENERIC_ERROR = 4030
 };
 
+enum GameIsoFlags { FLAG_JAK1_BLACK_LABEL = (1 << 0) };
+
```
xTVaser, line 30: Is there documentation for these flags somewhere?  I'm not really understanding what it's being used for

decompiler/extractor/main.cpp
```diff
@@ -407,6 +432,19 @@ int main(int argc, char** argv) {
...
+    auto flags_json = nlohmann::json::array();
+    for (const auto& [n, f] : sGameIsoFlagNames) {
+      if (flags & f) {
+        flags_json.push_back(n);
+      }
+    }
+    buildinfo_json["flags"] = flags_json;
+    // something tells me a ps2 game is unlikely to have a json file in root
+    file_util::write_text_file((path_to_iso_files / "buildinfo.json").string(),
+                               buildinfo_json.dump(2));
```
xTVaser, line 447: looks like we are only writing out the flags to this file?  Is this just to figure out what game was extracted?  If so why not just write out the serial/ELF_HASH (the keys to retrieve the same release info from the db).

decompiler/extractor/main.cpp
ManDude: The flag is used in `compile` below to set the value of `*jak1-full-game*` which the build system (and the game) needs to know since black label is missing the `COMMONJ.SBK` file.

decompiler/extractor/main.cpp
ManDude: I originally was gonna write the whole iso metadata here but ended up just doing flags. I didn't want to just have the elf hash in case the elf is identical but the other files in the disc aren't. I'm not sure this is the case (haven't checked yet) so I went with the safer option.

decompiler/extractor/main.cpp
ManDude: `*jak1-full-game*` is a boolean, not a bit string.

ManDude:
Added greatest hits support since it was really easy.

PR #1526, opened 2022-06-23
game: use the tagged version number if available

xTVaser:
This improves the existing logic to prefer the tag name if it's available, and using the commit sha otherwise like before.  It also adds it to two new locations:
- the dreaded top bar
- the OpenGL window title itself

image

This should hopefully make it much more obvious what version people are running when reporting issues.


ManDude:
while we're at it, add "WORK IN PROGRESS VERSION" to the title as well

PR #1527, opened 2022-06-23
github: improve bug report issue template

xTVaser:
This:
- adds some acknowledgements to the bug report (check for duplicates, have your own legit copy)
- asks if they've reproduced the bug on original hardware or not as this is often a first step we take
- improves the version gathering input, as just "release" is too ambiguous
- formats the issue template's yaml a bit


water111 (approved):
:+1:
It might be a good idea to ask people if they are on 150fps or not as part of the issue template

PR #1536, opened 2022-06-24
Fix one frame of ghost yellow eco in finalboss

### LINKED ISSUE #1535: [Bug] PS2 Issue Yellow eco in finalboss can reappear for 1 frame and "steal" a players blue eco.
### Acknowledgements

### Describe the Bug

This bug is in the original game, however its not that hard to fix. I understand if we dont add this but I personally dont see a reason to not include it - when this happens it is never funny or enjoyable it is only annoying. 

Videos of the bug:
https://clips.twitch.tv/ClumsyNeighborlyJellyfishTBTacoRight-JWJ06kO4w3kCOvgq

Video with debug view showing what happens by HS blue

https://www.youtube.com/watch?v=Lx892oLeoeA
 

### How To Reproduce

Collect yellow eco in finalboss, then wait a while and somewhere in the ecos die process have the finalboss call robotboss-yellow-eco-off.

I have submitted a PR that would fix this.

Not needed, bug is obvious

### Expected Behavior

Eco should not respawn for one frame

### Operating System

Windows

### OpenGOAL Version

Release 0.1.9

### Game Version

NTSC 1.0 (black label)

Yes

Zedb0T:
This would close #1535


Zedb0T:
Hold a day on this. It also does change how the eco respawns when you go into goggles. (for now)

ManDude:
You can mark your PR as a draft until it's ready.

Zedb0T:
Pretty sure it functions as normal with this patched now, but want to run a few more tests.

water111 (approved):
I'm fine with these changes if you've tested them, but I'm wondering if this is something that people would complain about if it's fixed.

Zedb0T:
I've ran through the final boss 15 times with developer mode collision on and they appear to work, then 3 times completely normal just rushing the boss. No one will even notice its fixed, this is more a thing you only notice when it "breaks". From a code/port accuracy/integrity standpoint I guess. But this port already does quite a few things different from the PS2 version

Zedb0T:
My testing is a bit limited as I dont think this change would be too hard to revert should we encounter unintended behavior the change is relatively small in the code and its commented well enough.

goal_src/levels/finalboss/robotboss.gc
```diff
@@ -245,6 +245,13 @@
   )
 
 (defbehavior robotboss-yellow-eco-off robotboss ()
+  ;; These logclears are not present in the orignal game, this is to patch the one frame of ghost yellow eco bug that can 
+  ;; occcur under certain conditions when robot-boss-yellow-eco-off is called.
+  (logclear! (-> self alts 7 extra perm status) (entity-perm-status bit-3))
+  (logclear! (-> self alts 8 extra perm status) (entity-perm-status bit-3))
+  (logclear! (-> self alts 9 extra perm status) (entity-perm-status bit-3))
+  (logclear! (-> self alts 10 extra perm status) (entity-perm-status bit-3))
+
```
ManDude, line 254: Put this in a `(#when PC_PORT` block

Zedb0T: Ah I see nice approach

ManDude:
Fwiw I *think* you can still get the bug if you pick up the eco at the same time as it's meant to despawn.

PR #1559, opened 2022-06-26
[ckernel] split by game version

water111:
This is a refactoring nightmare, but I really want to avoid totally redoing the decomp of ckernel for Jak 2 - there's a lot of identical code and stuff that's only a tiny bit different.  It's harder than I though because of all the tangled global variables.

This is the best pattern I've found so far - split up each file into a common, jak1, and jak2 version and put as much in common.

The stuff in the common/jak1/jak2 folders is done being split (but, in the jak 2 case, some functions are missing where I got lazy) and the stuff remaining in `kernel/*.{cpp, h}` is not yet split.


yodika:
My theory is it'd because of ps2 internal code processing, it can only handle so much, that's why there's alot of identical, like 1s and 0s, possibly to stabilize ps2 from overheating by too much functions being activated at once.
Kinda like a filler for an anime.

PR #1562, opened 2022-06-26
Fix volume calculation when falloff curve is zero

### LINKED ISSUE #1380: [sound] sound effects dont work in finalboss

Francessco121:
Super simple fix to match the original overlord behavior for `CalculateFalloffVolume`.

Fixes #1380.


Ziemas:
oops

water111 (approved):
Wow, great find!

PR #1563, opened 2022-06-26
[build_level] basic support for entity/res-lump

water111:


goal_src/engine/math/vector-h.gc
```diff
@@ -736,6 +736,10 @@
      )
   )
 
+(defmacro print-vector4m (vec &key (dst #t))
+  `(format ,dst "~F ~F ~F ~F~%" (/ (-> ,vec x) 4096.) (/ (-> ,vec y) 4096.) (/ (-> ,vec z) 4096.) (/ (-> ,vec w) 4096.))
```
ManDude, line 740: why not use `~m`?

water111: good idea

PR #1569, opened 2022-06-28
Fix "last-hit" animations not playing for robotboss

### LINKED ISSUE #1257: robotboss parts dont disappear when broken
image

Francessco121:
This was causing destroyed parts of the robot to float in the air not moving for a moment.

Not really sure what passing 0 does here, but every other use of `'art-joint-anim` has it and sure enough it fixed the issue.
Fixes #1257.


ManDude:
It's the time for how long to blend the animation for - not sure why the decompiler would be omitting it though. Probably should check what's going on here.

Francessco121:
Oh ok. Is there any way to verify that it should be zero in this case? It seems to look correct.

ManDude:
You should look around the IR2 output of the functions that have this code. You can also snoop around the `rewrite_let` function and disable the `rewrite_send_event` function so that you can see the unwrapped version of this macro.

water111:
Another great find!

I'm pretty sure what happens is that they left this out the 0 in the original game and it ends up reading some uninitialized memory on the stack, and they got lucky and was 0 or some other value that worked.

I think this is still a good fix, but can you just add a comment that says something like `;; this 0 was missing in the original game, but is a bug that causes undefined behavior`?

We can confirm that they left out the 0, and it's not a decompiler bug by looking at the assembly.  You can modify `jak1_ntsc_black_label.jsonc` and set `decompile_code` to `true`, then run the decompiler (with `decomp.sh` on linux or `decomp-jak1.bat` on windows) and it will generate tons of debug output.

The part we're looking for is the `trans` handler of `robotboss-yellow-wait`, which the decompiler calls `(trans robotboss-yellow-wait)`:

```asm
    daddiu a1, sp, 32        
    sw s6, 4(a1)             
    addiu v1, r0, 1          
    sw v1, 8(a1)             ;; v1 is the parameter count, which is 1
    daddiu v1, s7, art-joint-anim
    sw v1, 12(a1)            ;; message name
    lui v1, L686                              
    ori v1, v1, L686         
    daddu v1, v1, fp    
    sd v1, 16(a1)           ;; store the single parameter (the string)                        
    lw t9, send-event-function(s7)           
    beq s7, gp, L430   
    or a0, s7, r0

B12:
    lwu v1, 0(gp)      
    lwu a0, 24(v1)     
B13:
L430:
    jalr ra, t9        
    sll v0, ra, 0
```

If we compare to one where it does send a 0:
```asm
    daddiu a1, sp, 16    
    sw s6, 4(a1)         
    addiu v1, r0, 2      ;; sending 2 parameters
    sw v1, 8(a1)         
    daddiu v1, s7, art-joint-anim
    sw v1, 12(a1)           
    daddiu v1, fp, L264     
    sd v1, 16(a1)        ;; first parameter (the string)
    sd r0, 24(a1)        ;; second parameter (the 0)   
    lw t9, send-event-function(s7)
    or v1, gp, r0   
    beq s7, v1, L130
    or a0, s7, r0

B5:
    lwu v1, 0(v1)   
    lwu a0, 24(v1)  
B6:
L130:
    jalr ra, t9     
    sll v0, ra, 0
```

If you check the code for `manipy-idle` `event`, it reads the second parameter without checking the number sent. This is admittedly a place where the decompiler did a very messy job:
```
          ((= v1-0 'art-joint-anim)
           (set! (-> self new-joint-anim)
                 (the-as art-joint-anim (lookup-art (-> self draw art-group) (the-as string (-> arg3 param 0)) art-joint-anim))
                 )
           (set! v0-0 (the-as none (-> arg3 param 1))) ;; here
           (set! (-> self new-joint-anim-blend) (the-as uint v0-0))
           v0-0
           )
 ```

Francessco121:
Wow, you guys are fast! 😄 I've added comments.

PR #1572, opened 2022-06-28
subtitle editor fixes + other smaller fixes

### LINKED ISSUE #1550: `\c##` escape sequences are converted to special characters when loaded into the subtitle editor
### Acknowledgements

### Describe the Bug

This affects the subtitle editor, don't know if it affects anything else. When a subtitle file is parsed that contains a sequence like `\c93` or `\c14` (yes, i found out i'm supposed to use `<TIL>` in this case) the escape sequence is turned into a character, even for `line_utf8`. In the case of `\c93` it becomes a �, but in the case of \c14 it causes a [`Invalid character found on line # of #: 0x14`](https://github.com/open-goal/jak-project/blob/master/common/goos/Reader.cpp#L279) when the repl then tries to make the text segment.

It seems like fixing this might require messing around in the goos code or FontUtils.cpp and I really don't wanna break anything else by trying to fix this. if I'm wrong about that, and this can be avoided just in the case of the subtitle editor, then I'm more than happy to include the necessary changes in #1516 .

### How To Reproduce

Put a line in a subtitle like this:
`  (2226 "SAGE" "YOU CAN'T CONTROL DARK ECO BY ITSELF! EVEN THE PRECURSORS\c93")` (yes, i know, i'm using a chouonpu as an em-dash, but it serves that purpose nicely imo, and a hyphen would look wrong)
Save it, then restart the game. in the file, the escape sequence is present, but in the editor it's now
`  (2226 "SAGE" "YOU CAN'T CONTROL DARK ECO BY ITSELF! EVEN THE PRECURSORS�")` (note: it's rendered as a ? in the editor, but it's not one. also, the character is still rendered as expected in-game.)
if you then save the changes again, the broken character is saved to the subtitle file.

Yes, it's unique to OpenGOAL

### Expected Behavior

`\c##` sequences should remain as they are when loaded into the subtitle editor or any other tool that has you editing GOAL strings. i know that characters such as apostrophe are converted to something else for the game, yet they show up in the subtitle editor as apostrophes and are saved back to the file just fine, so the issue seems to be with these escape sequences specifically.

### Operating System

Windows

### OpenGOAL Version

b8bddb

### Game Version

NTSC 1.0 (black label)

No

### LINKED ISSUE #1566: [extractor] Validate game version and set flags for build info with folder extraction
When using the `-f` flag to extract the game assets using a folder with the extracted ISO contents inside, the extractor does not validate the game info and does not set the proper flags used for the `buildinfo.json`.

This can, for example, result in the user being unable to compile the black label version of the game as the compiler will be looking for the `COMMONJ` soundbank which is not present in that version.

ManDude:
Reworks the way text files work a bit to make it less confusing to work with.

Fixes #1566 and fixes #1550 and fixes texscroll in lag.

Also fixes other game versions not decompiling.


decompiler/data/game_text.cpp
```diff
@@ -150,6 +151,11 @@ std::string write_game_text(
   }
   std::sort(languages.begin(), languages.end());
 
+  // hack
+  if (version == GameTextVersion::JAK1_V1) {
```
water111, line 155: is this supposed to be here?

decompiler/extractor/main.cpp
```diff
@@ -105,6 +105,119 @@ std::pair<std::optional<std::string>, std::optional<xxh::hash64_t>> findElfFile(
...
+    } else {
+      fmt::print(stderr,
+                 "Validation has failed to match with expected values, see the above errors for "
+                 "specifics. This may be an error in the validation database!\n");
+    }
+    return {*error_code, std::nullopt};
+  }
+
+  return {ExtractorErrorCode::SUCCESS, meta_res};
+}
```
water111, line 219: I wonder if we could use this version of `validate` for both the ISO and folder versions. That way we don't have to add the file hashing stuff to the ISO reader, and we have less duplication.

But we don't have to worry about this for now.

water111 (approved):

decompiler/data/game_text.cpp
ManDude: yeah, but now that I think about it I can probably change the black label config's `text_version` to be the same as PAL+ since we apply a patch now and avoid the hack here altogether.

PR #1575, opened 2022-06-29
Remove assets folder, use more std::filesystem

water111:
This removes the `assets` folder. The `fr3`s now go in `out/fr3` and the decompiler output stuff goes into `decompiler_out/<game-name>/whatever`

The downside of this is that tools can't just look in `assets/` to find the outputs from the latest decompiler run. But I think this isn't a bad thing because that approach didn't work too well - if you did a random jak2 decompile, your next jak1 build would grab text files from jak 2.


water111:
I tested jak 1 decomp, texture outputs, streaming audio outputs, building in `jak-project`, and the extractor and it all seems to work.

I think the one thing that could go wrong is if somebody is using a non `jak1/` decompiler_out.

water111:
The remaining issue is how the compiler accesses the output of the decompiler.  Most of that is set in `game.gp`, but there is one path in `game_text.gp` that needs to reference a decompiler output. We should probably pass this path through in `game.gp` directly somehow.

PR #1580, opened 2022-06-30
fix windows scripts for jak 2

ManDude:


ZornTaov:
Does `generate_all_types` need to be set twice in `jak2_ntsc_v1.jsonc`?

PR #1589, opened 2022-07-01
build: get rid of `clang-cl` in favor of actual `clang` among other things

xTVaser:
This greatly speeds up windows builds once again by being able to cache them, does some cleanup of our cmake and builds (more to come eventually) and vendors nasm so we should never have intermittent issues again.

image

Windows builds also use Ninja instead of NMake.

After 2 long years ive finally won the fight
image

TODO:
- [x] update docs
- [x] check release / extractor process


.github/workflows/linux-workflow.yaml
```diff
@@ -105,9 +103,27 @@ jobs:
...
+            ./build/decompiler/decompiler
+            ./build/decompiler/extractor
+            ./build/game/gk
+            ./build/sound/sndplay
+            ./build/goalc/goalc
+            ./build/tools/dgo_packer
+            ./build/tools/dgo_unpacker
+            ./build/tools/memory_dump_tolls
+            ./build/tools/level_tools/level_dump
+
```
Margen67, line 126:
```suggestion
            build/**/*.so
            build/**/*.a
            build/decompiler/decompiler
            build/decompiler/extractor
            build/game/gk
            build/sound/sndplay
            build/goalc/goalc
            build/tools/dgo_packer
            build/tools/dgo_unpacker
            build/tools/memory_dump_tolls
            build/tools/level_tools/level_dump
        if-no-files-found: error

```

.github/workflows/windows-workflow.yaml
```diff
@@ -39,45 +39,52 @@ jobs:
...
+        timeout-minutes: 10
         run: ./build/bin/goalc-test.exe --gtest_color=yes --gtest_filter="-*MANUAL_TEST*"
 
+      - name: Upload artifact
+        if: matrix.compiler == 'clang'
+        uses: actions/upload-artifact@v3
+        with:
+          name: opengoal-${{ matrix.os }}-${{ matrix.compiler }}
+          path: ./build/bin
+
```
Margen67, line 87:
```suggestion
          path: build/bin
          if-no-files-found: error

```

.github/workflows/linux-workflow.yaml
xTVaser: The artifacts are just a nice to have, i don't want to fail the build over them not being found.  I also prefer the leading `./` style, im aware it's not required.

PR #1590, opened 2022-07-02
game: read current date correctly for save data

Hat-Kid:
Makes `sceCdReadClock` pass the correct date values as required by `scf-get-time` in GOAL by converting every decimal value to a hex `u8` with the same digit representation.


xTVaser (approved):
Looks good to me, here was a bad file i already had:
image
A brand new one was correct
image
And overwriting the bad one fixed it as well
image

water111 (approved):
Thanks!

PR #1591, opened 2022-07-02
tests/jak2: Better prepare offline tests for multiple games

xTVaser:
The only thing i believe will need to be updated here is the assumptions around the internal DGO/CGO structure.  Jak 1 was very simple, iirc jak 2 is a bit different in this regard.

Based off my other branch as I can no longer build the project locally after updating VS (draft until it's merged)


common/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ add_library(common
         util/os.cpp
         util/print_float.cpp
         util/FontUtils.cpp
-        util/FrameLimiter.cpp)
```
ManDude: make sure to fix this

xTVaser: Are you referring to how the quoted style slipped in on one of my earlier PRs? https://github.com/open-goal/jak-project/blob/master/common/CMakeLists.txt#L51 there is no change to this file in this PR's diff.

ManDude: oops? I guess might as well fix it now.

xTVaser: I can, just wasn't sure what you were referring to

PR #1617, opened 2022-07-06
add text messages for subtitle toggling + subtitle editor fixes

ManDude:


game/tools/subtitles/subtitle_editor.cpp
```diff
@@ -149,7 +149,6 @@ void SubtitleEditor::draw_window() {
     ImGui::PushStyleColor(ImGuiCol_Text, m_selected_text_color);
   }
   if (ImGui::TreeNode("Currently Selected Cutscene")) {
-    ImGui::PopStyleColor();
     if (m_current_scene) {
```
xTVaser, line 152: without running it locally, im pretty sure that not popping inside the `TreeNode` block will make the contents within incorrectly colored (when opened)

PR #1626, opened 2022-07-08
Sprite Distort

Francessco121:
This PR implements the sprite distortion effect as part of the sprite3 renderer.

I'm still fairly new to a lot of stuff in this project, so if there's anything I need to refactor or clean up please let me know!

The sprite distort effect needs to read the active framebuffer to do its thing. I implemented this by creating a separate framebuffer managed by the sprite renderer. When it comes time to render sprite distortion, the default framebuffer's color buffer is blitted (using `glBlitFramebuffer`) to this framebuffer. Then the color buffer of this framebuffer is bound to the shader as a texture. I'm not sure what the performance implications of this are. I have a feeling there's a faster way to implement this effect, but this was a super simple approach.

Since the sprite renderer managed framebuffer needs to have the same dimensions as the default framebuffer, I added the pixel width/height of the window to `SharedRenderState`. I'm not sure if this is the most appropriate way to pass this information to the renderer. Currently, the renderer will resize the framebuffer on demand and checks to see if the width/height changed each frame.

**Issues:**
~At the moment the effect is not perfectly accurate to the original game. Specifically, the texture coordinates used for sampling the framebuffer texture don't look right unless the Y-axis is flipped and then offset by \~0.06. I know this isn't right but I don't really understand why this is necessary.~

~Additionally, the current implementation issues a lot of draw calls (one for each sprite).~

### DIFF
--- a/docs/scratch/sprite_distort_vu1.txt
+++ b/docs/scratch/sprite_distort_vu1.txt
@@ -1,0 +1,94 @@
+;; VU memory:
+;; 0:   GS vertex data (buffer 1)
+;; 176: GS vertex data (buffer 2)
+;; 352: sinetables.entry[128]
+;; 480: sinetables.ientry[9]
+;; 489: sinetables.giftag
+;; 490: sinetables.color
+;; 511: num sprites
+;; 512: sprite data
+
+;; Init
+  lq.xyzw vf01, 489(vi00)    |  nop                         ;; Load giftag
+  lq.xyzw vf05, 490(vi00)    |  nop                         ;; Load color
+  ilw.x vi01, 511(vi00)      |  nop                         ;; Load number of sprites
+  iaddiu vi04, vi00, 0x200   |  nop                         ;; Address of sprite data (512)
+  iaddi vi02, vi00, 0x0      |  nop                      
+
+;; Loop for each sprite
+L1:
+  ilw.w vi07, 1(vi04)        |  nop                         ;; Get sprite.flag
+                                                            ;; The flag represents the 'resolution' of the sprite,
+                                                            ;; which is a circle. The vertices are constructed below
+                                                            ;; by creating "pie-slices" of a circle where the number
+                                                            ;; of slices is the value of flag. In practice, flag
+                                                            ;; can range from 3-11.
+  ior vi05, vi02, vi00       |  nop                      
+  ilw.x vi06, 477(vi07)      |  nop                         ;; Get ptr to sinetables.entry
+                                                            ;; addr = sinetables.ientry[flag - 3]
+                                                            ;; The smallest value of flag is 3 and 3 + 477 = 480, 
+                                                            ;; which is sinetables.ientry.
+                                                            ;; The values in ientry are an index of entry + 352, 
+                                                            ;; which gives the address of an entry element.
+  sqi.xyzw vf01, vi05        |  nop                         ;; Write giftag to GS data
+  iaddiu vi08, vi07, 0x4000  |  nop                      
+  iaddiu vi08, vi08, 0x4000  |  nop                         
+  isw.x vi08, -1(vi05)       |  nop                         ;; Set giftag.nloop to flag (number of slices)
+  lqi.xyzw vf02, vi04        |  nop                         ;; Get sprite.xyz
+  lqi.xyzw vf03, vi04        |  nop                         ;; Get sprite.st
+  lqi.xyzw vf04, vi04        |  nop                         ;; Get sprite.rgba
+  nop                        |  ftoi4.xyzw vf14, vf02    
+
+;; Loop the number of times denoted by sprite.flag
+;; Here, we build the GS data for a single distort sprite
+L2:
+  lqi.xyzw vf06, vi06        |  nop                         ;; Load 4 sine table vectors
+  lqi.xyzw vf07, vi06        |  nop                      
+  lq.xyzw vf08, 0(vi06)      |  nop                      
+  lq.xyzw vf09, 1(vi06)      |  nop                      
+  iaddi vi07, vi07, -0x1     |  muly.xyzw vf10, vf06, vf04  ;; Decrement loop count
+  nop                        |  mulz.xyzw vf11, vf07, vf04  ;; Calculate vertex positions, texture coords, and color
+  nop                        |  muly.xyzw vf12, vf08, vf04
+  nop                        |  mulz.xyzw vf13, vf09, vf04
+  nop                        |  mulx.xyzw vf06, vf06, vf04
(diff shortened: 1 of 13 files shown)


docs/scratch/sprite_distort_vu1.txt
```diff
@@ -0,0 +1,94 @@
...
+
+;; Finally, tell the GS to draw the current sprite and switch
+;; to the other buffer for the next sprite
+  xgkick vi02                |  nop                         ;; Draw sprite
+  iaddi vi01, vi01, -0x1     |  nop                      
+  iaddiu vi03, vi00, 0xb0    |  nop                         ;; Switch buffer
+  ibne vi00, vi01, L1        |  nop                         ;; Loop for remaining sprites
+  isub vi02, vi03, vi02      |  nop                      
+  nop                        |  nop :e                   
+  nop                        |  nop
```
water111, line 94: thanks for including this documentation!

game/graphics/opengl_renderer/OpenGLRenderer.cpp
```diff
@@ -414,6 +414,9 @@ void OpenGLRenderer::setup_frame(int window_width_px,
   glDepthMask(GL_TRUE);
   glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
   glDisable(GL_BLEND);
+
+  m_render_state.window_width_px = window_width_px;
+  m_render_state.window_height_px = window_height_px;
```
water111, line 419: I think I'm running into issues related to this, when the game doesn't take up the full window:
image

I'm not actually sure this is something that's supposed to be possible or not, but the resolution/aspect ratio settings don't really work for me in the first place.  This is probably a separate issue and shouldn't block this PR.

game/graphics/opengl_renderer/Sprite3.cpp
```diff
@@ -116,15 +126,94 @@ Sprite3::Sprite3(const std::string& name, BucketId my_id)
...
+  glVertexAttribPointer(0,                                         // location 0 in the shader
+                        3,                                         // 3 floats per vert
+                        GL_FLOAT,                                  // floats
+                        GL_TRUE,                                   // normalized, ignored,
+                        sizeof(SpriteDistortVertex),               //
+                        (void*)offsetof(SpriteDistortVertex, xyz)  // offset in array
+  );
+  glEnableVertexAttribArray(1);
+  glVertexAttribPointer(1,                                        // location 1 in the shader
+                        3,                                        // 2 floats per vert
```
water111, line 173: This should be a 2 - it caused some interesting problems where the last vertex had messed up texture coordinates on my GPU:
image

game/graphics/opengl_renderer/Sprite3.cpp
```diff
@@ -116,15 +126,94 @@ Sprite3::Sprite3(const std::string& name, BucketId my_id)
...
 
 /*!
- * Run the sprite distorter.  Currently nothing uses sprite-distorter so this just skips through
- * the table upload stuff that runs every frame, even if there are no sprites.
+ * Run the sprite distorter.
  */
 void Sprite3::render_distorter(DmaFollower& dma,
                                SharedRenderState* render_state,
                                ScopedProfilerNode& prof) {
+  // First, make sure the distort framebuffer is the correct size
```
water111, line 203: It might be nice to add a way to disable sprite distort, just in case it causes massive performance issues for some people.

You could follow the example of `m_3d_enable` to set up a checkbox in the debug GUI.

goal_src/jak1/engine/gfx/sprite/sprite-distort.gc
```diff
@@ -111,92 +117,40 @@
...
-    (set! (-> (the-as (pointer gs-clamp) a2-4) 8)
-          (new 'static 'gs-clamp
-               :wms (gs-tex-wrap-mode region-clamp)
-               :wmt (gs-tex-wrap-mode region-clamp)
-               :maxu #x1ff
-               :maxv (-> *video-parms* screen-masky)
-               )
+  
+  ;; set up GS registers
+  (dma-buffer-add-gs-set dma-buff
```
water111, line 125: nice use of the macro

water111 (commented):
Looks great!  I had just a few comments, but other than those, I think this is good to go

water111:
Also - to clarify - with the `st` fix and the "fit to screen" option, it worked perfectly for me, with very change in frame times

game/graphics/opengl_renderer/Sprite3.cpp
Francessco121: Oops, good catch. I'm surprised that didn't cause issues on my machine...

game/graphics/opengl_renderer/OpenGLRenderer.cpp
Francessco121: Turned out to be a simple issue with the framebuffer blit. The source dimensions needed to respect the XY offset of the renderer.

game/graphics/opengl_renderer/Sprite3.cpp
Francessco121: Done, also made it respect the global sprite `m_enabled` flag.

Francessco121:
I addressed the comments and also did a bit of a refactor. Each 'stage' of the render (dma, setup, draw) are fully separate now and have their own profiling scopes.

Francessco121:
The size of `sprite-aux-list` now scales up when `PC_BIG_MEMORY` is set just like the other sprite stuff. This fixes an issue mainly around the klaww fight and volcanic crater where most distort sprites wouldn't render since without particle sprite fade out the original 256 sprite cap is hit.

I think that's pretty much every issue solved! 😄

game/graphics/opengl_renderer/Sprite3.cpp
```diff
@@ -116,43 +127,375 @@ Sprite3::Sprite3(const std::string& name, BucketId my_id)
...
+    glBindTexture(GL_TEXTURE_2D, m_distort_ogl.fbo_texture);
+
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, m_distort_ogl.fbo_width, m_distort_ogl.fbo_height, 0,
+                 GL_RGB, GL_UNSIGNED_BYTE, NULL);
+
+    glBindTexture(GL_TEXTURE_2D, 0);
+  }
+
+  if (m_distort_stats.total_tris == 0) {
+    // No distort sprites to draw, we can end early
```
water111, line 429: good idea to add this, I imagine there's some places with no sprite distorts

water111 (approved):

PR #1630, opened 2022-07-09
Sprite Distort Instancing

Francessco121:
This adds an additional render path to use instancing for sprite distort instead of one giant mesh. A neat part about this is that a good chunk of the original VU calculations are now in the shader.

The way this works is: vertex data for each possible sprite resolution is calculated ahead of time (and recalculated whenever the sine table changes). This results in much less vertex data and doesn't need to be re-uploaded to the GPU every frame. The pre-calculated vertex data just contains position and texture coordinate information from the sine tables. Then, each frame the sprite-specific data is sent to the GPU as an instanced array containing the per-frame sprite position, scale, and texture coordinates. Finally, the shader then does the original VU calculations with the sine table data and sprite frame data.

A slight catch with moving the calculations to the shader is that the shader needs to unfortunately check the current vertex ID. Depending on which of the 5 vertices that make up each 'slice' of the sprite is being rendered, a different scale value needs to be used. This effectively means the vertex shader just has a few `if` statements. It would be preferable to not have any branches, but I couldn't think of anything and it doesn't seem to affect performance regardless.

Regarding performance, this is (at least on my machine) roughly **twice as fast** in scenarios where a decent number of distort sprites are drawn (fire canyon mid-section, klaww, volcanic crater, etc.). In cases where there isn't too many sprites, the performance is equal to the original approach. Most of the remaining frame time is just the framebuffer blit.

I also left in all the original code and made this togglable via the debug menu. I don't think anyone will need to turn it off, but maybe it's worth keeping it as an option for now?

### DIFF
--- a/game/graphics/opengl_renderer/Shader.cpp
+++ b/game/graphics/opengl_renderer/Shader.cpp
@@ -88,6 +88,7 @@ ShaderLibrary::ShaderLibrary() {
   at(ShaderId::COLLISION) = {"collision"};
   at(ShaderId::MERC2) = {"merc2"};
   at(ShaderId::SPRITE_DISTORT) = {"sprite_distort"};
+  at(ShaderId::SPRITE_DISTORT_INSTANCED) = {"sprite_distort_instanced"};
 
   for (auto& shader : m_shaders) {
     ASSERT_MSG(shader.okay(), "Shader compiled");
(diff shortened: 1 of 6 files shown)


water111 (approved):
Thanks!
I got similar results for performance - it's about 2x faster in heavy places, and a little bit faster when there's only a few distort sprites.

Issue #1524, opened 2022-06-23
2 Powercells Are Impossible To Obtain


EliteZalba:
### Describe the Bug

The 1st powercell that is impossible to obtain is the 1st powercell in the Precursor Basin, obtained by herding all of the moles into their hole.  The moles can be herded, seemingly correctly, but they don't ever burrow back into their hole.  I've seen playthroughs online where there is quite a bit of tolerance with how close the mole has to be to burrow back into the hole, but in my current playthrough, they will literally walk dead center over the hole, without burrowing.

The 2nd powercell that is impossible to obtain is the 7th (2nd to last) in the Volcanic Crater.  The powercell description is hidden, but it is the powercell contained in a metal box that is located 180 degrees behind the opening to Spider Cave.  It requires entering Spider Cave, filling up on a yellow eco vent, and shooting a projectile at the metal box to break it open and reveal the powercell.  The bug is regarding the hitbox of the metal crate.  There is an invisible barrier directly in front of the metal box that prevents any yellow eco projectiles from making contact with and breaking open the metal box.

*Not sure which OpenGOAL version it is, but I followed the windows setup guide exactly on 6/19/22.  I'm assuming this is the released version. 

### How To Reproduce

The steps to reproduce the behavior are simply by attempting to obtain the respective powercells.  Attempting to herd the moles should show that they don't burrow back into their hole, and attempting to fill up on a yellow eco vent and shooting a projectile at the metal crate in the Volcanic Crater should reveal the barrier blocking the yellow eco projectiles from hitting the metal box.

### Expected Behavior

Precursor Basin powercell: forcing a mole to walk very near to or directly over the hole won't trigger the burrow animation.

Volcanic Crater powercell: shooting a yellow eco vent projectile at the metal box will result in the yellow eco projectile being blocked by an invisible barrier, and the metal box will remain unbroken.

### Operating System

Windows 10, Version 21H2, OS Build 19044.1766

### OpenGOAL Version

Release

### Game Version

NTSC 1.0 (black label)

djak250:
I just completed the mole mission, on steam deck. It was difficult but they did eventually find their way home. For what it's worth.

gsilver0:
I had no issues with the mole mission.

BreakPoints:
Sounds like a 150FPS bug, was able to grab all 100 power cells in a pre 150FPS build.

Kringuskrang:
[...]

PR #1634, opened 2022-07-10
Some high FPS fixes

### LINKED ISSUE #1511: [High FPS] Bugged Water Platform Sound
### Describe the Bug

when jumping on these water platforms in Geyser Rock [Tested on 150fps, while comparing to 60fps] the sounds produced come out amplified and multiplied.

Uploading 2022-06-21 21-25-19.mp4…

### How To Reproduce

Jump on (I presume) any water platform like those shown in the video at an fps set higher than 60, although, this was only so far tested in Geyser Rock.

### Expected Behavior

The Audio should come out Multiplied and/or Amplified

### Operating System

Windows 10

### OpenGOAL Version

Release

### Game Version

NTSC 1.0 (black label)

Francessco121:
Fixes high FPS issues with:
- Rigidbody simulation (fixes #1511 and issue with floating platforms snapping back to their neutral position when Jak jumps off them too quickly)
- Lightning moles leaving the 'chase' state too easily (fixes the remaining issue in #1524)
- Swamp rats wiggling too fast
- 'Kermit' zipping to Jak while airborne
- Low health HUD flashing
- Low eco HUD flashing
- High zoomer heat HUD flashing
- Camera rotation when falling from a blue eco launcher

Some of these issues were listed in the mega thread #1499

### DIFF
--- a/goal_src/jak1/engine/common-obs/generic-obs.gc
+++ b/goal_src/jak1/engine/common-obs/generic-obs.gc
@@ -1763,7 +1763,7 @@
              (f1-1 (* -546.13336 f0-0))
              (f0-2 (fmin 546.13336 (fmax -546.13336 f1-1)))
              )
-        (matrix-axis-angle! gp-0 (-> *camera* local-down) f0-2)
+        (matrix-axis-angle! gp-0 (-> *camera* local-down) (* DISPLAY_FPS_RATIO f0-2)) ;; high fps fix
         )
       (vector-matrix*! (-> self view-flat) (-> self view-flat) gp-0)
       )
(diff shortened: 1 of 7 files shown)


water111 (approved):
Nice!
All these look like they have no effect in 60fps, which is good. 
For the `kermit.gc` stuff - I am a little confused with what the original code is trying to do, but the fix seems reasonable.

Francessco121:
The kermit code was a little weird, but I noticed that other AI code (like yakows for example) have roughly the same exact function except they scale the final `transv` based on frame time. `kermit.gc` seems to be the only one that didn't.

PR #1638, opened 2022-07-12
[jak2] decomp gkernel, setup offline tests

water111:
Decompiler/compiler seem to work.

Definitely still some jak 2 specific stuff left to do (like floats) but I want to just start fresh with this problem on a new type analysis pass.  Also, found some candidates in later files for places we could auto-detect label/stack structures.  It seems like this could be a single new system to do better types/stack structure autodetection/label autodetection.


goal_src/jak2/kernel/gcommon.gc
```diff
@@ -861,13 +986,38 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;
+;; printing
+;;;;;;;;;;;;;;;;;;;;;;;;
+
+;; the column that will be printed to by format.
 (define *print-column* (the-as binteger 0))
 
+;; note: normal use of print/inspect will have the compiler pick the appropriate method
+;; for non-basics.  However, it may be useful to have print/inpsect available as a function
```
ManDude, line 998: inspect*

goal_src/jak2/kernel/gkernel-h.gc
```diff
@@ -6,4 +6,686 @@
...
+;;     This counts in real-time, even if the game is lagging.
+;;
+;; 2). A "integral-frame-counter", which counts the number of vsyncs.
+;;     This doens't count the number of frames the game actually manages to draw,
+;;     just the number of vsyncs. It counts at different rates in NTSC/PAL.
+;;     NOTE: changing clock-ratio will make this count faster/slower. This only counts real
+;;           vsyncs if clock-ratio is 1.0.
+;;
+;; 3). The "time ratio", which adjusts based on the actual achieved framerate.
+;;     Unlock the others, this isn't a incrementing counter, but instead ratios:
```
ManDude, line 142: Unlike*

PR #1646, opened 2022-07-14
Fix instanced sprite distort renderer not always using the most recent sine table

Francessco121:
Sometimes the game won't update the sine table on the same frame that the aspect ratio changed (for example if the player is in a cutscene), but the renderer assumed that it was always up-to-date on the next frame. This caused it to occasionally use outdated vertex/tex coord data that usually made the distort sprites a little too small/big.

The GOAL code now sends up a PC-port specific DMA that contains the aspect x/y that is used to determine when to rebuild the sine table. When the renderer gets new values from this, it's guaranteed that the sine tables were updated and the renderer can rebuild the cached vertex data.


water111 (approved):
Looks good, but looks like there's a merge conflict, maybe from the msaa stuff.

Francessco121:
Merge should be all set now. The MSAA update works great with this.

PR #1647, opened 2022-07-14
LSP: initial LSP implementation for IR files to assist with decompilation

xTVaser:
Supports the following features right now:
- hovering mips instructions to get the human readable name of them
- go-to-def on GOAL symbols to jump to all-types files
- file outline - jump to functions in files
- file diagnostic hints -- mark warns and info to see where problems are in the file


lsp/handlers/lsp_router.cpp
```diff
@@ -0,0 +1,138 @@
...
+  header.append("Content-Length: " + std::to_string(content.dump().size()) + "\r\n");
+  header.append("Content-Type: application/vscode-jsonrpc;charset=utf-8\r\n");
+  header.append("\r\n");
+  return header + content.dump();
+}
+
+std::optional<std::vector<std::string>> LSPRouter::route_message(
+    const MessageBuffer& message_buffer,
+    AppState& appstate) {
+  json body = message_buffer.body();
```
water111, line 65: can this be `json& body` to avoid a copy?

lsp/handlers/lsp_router.cpp
```diff
@@ -0,0 +1,138 @@
...
+  auto method = body["method"];
+  lg::info(method);
```
water111, line 67: Does this work? Won't the type of `method` be `json` here?

lsp/state/data/mips_instruction.cpp
```diff
@@ -0,0 +1,531 @@
...
+    {"vu", "VSQRT", "Square Root"},
+    {"vu", "VSUB", "Subtract"},
+    {"vu", "VSUBi", "Subtract I Register"},
+    {"vu", "VSUBq", "Subtract Q Register"},
+    {"vu", "VSUBbc", "Broadcast Subtract"},
+    {"vu", "VSUBA", "Subtract to Accumulator"},
+    {"vu", "VSUBAi", "Subtract I Register to Accumulator"},
+    {"vu", "VSUBAq", "Subtract Q Register to Accumulator"},
+    {"vu", "VSUBAbc", "Broadcast Subtract to Accumulator"},
+    {"vu", "VWAITQ", "Q Register Synchronize"},
```
water111, line 529: It might be possible to reuse the `OpcodeInfo.cpp` stuff, if you eventually want to have more information about the the number/kind of arguments to each op.

lsp/state/workspace.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+void WorkspaceAllTypesFile::parse_type_system() {
+  lg::debug("DTS Loading - '{}'", m_file_path.string());
+  m_dts.parse_type_defs({m_file_path.string()});
+  lg::debug("DTS Loaded At - '{}'", m_file_path.string());
+}
+
+void WorkspaceAllTypesFile::update_type_system() {
+  // TODO - do i actually need to re-create it, or is there a way to "clear" it?
```
water111, line 287: recreating it is fine - I prefer to do things that way because there's no need to write a separate `reset` method, and I think objects that are easy to reconstruct like this are a good pattern.

water111 (changes requested):
Looks good, just a few minor things.

common/versions.cpp
```diff
@@ -12,4 +12,19 @@ GameVersion game_name_to_version(const std::string& name) {
   } else {
     ASSERT_MSG(false, fmt::format("invalid game name: {}", name));
   }
-}
\ No newline at end of file
+}
+
+bool valid_game_version(const std::string& name) {
+  return name != "jak1" || name != "jak2";
```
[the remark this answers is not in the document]
xTVaser: For my use-case I don't want an assert though as that crashes the LSP

.github/workflows/release-pipeline.yaml
```diff
@@ -46,20 +46,24 @@ jobs:
...
+      - name: Prepare Linux Release Assets
         run: |
           mkdir -p ./ci-artifacts/linux
           ./.github/scripts/releases/extract_build_linux.sh ./ci-artifacts/linux ./ci-artifacts/opengoal-linux-static ./
-          cd ci-artifacts/linux
+          pushd ci-artifacts/linux
           TAG_VAL=$(echo ${{ github.REF }} | awk -F'refs/tags/' '{print $2}')
           tar czf ../final/opengoal-linux-${TAG_VAL}.tar.gz .
+          popd
+          strip ./ci-artifacts/opengoal-linux-static/lsp/lsp
```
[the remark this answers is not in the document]
xTVaser: True, although ive yet to figure out a decent way to debug the app.  Since the editor owns the process (starts it and such), it seems very difficult to attach a debugger to it anyway.

lsp/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+include_directories(./)
```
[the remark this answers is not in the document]
xTVaser: That's fair -- to be honest, visual studio took days for the intellisense to correctly pick up on this anyway 😄

lsp/handlers/initialize.hpp
```diff
@@ -0,0 +1,11 @@
+#include <lsp/protocol/initialize_result.hpp>
```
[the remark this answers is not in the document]
xTVaser: i wish i knew how to change this about visual studio...

lsp/state/data/mips_instruction.cpp
xTVaser: Hmm yeah that could be useful, id need a way to go from a mnemonic -> `InstructionKind` though i think?

common/type_system/TypeSpec.h
```diff
@@ -149,4 +150,9 @@ class TypeSpec {
   // arguments (most of the time) and makes the type analysis pass in the decompiler 2x faster.
   std::vector<TypeSpec>* m_arguments = nullptr;
   cu::SmallVector<TypeTag, 1> m_tags;
+
+  // definition information
+  std::string m_defining_file;
+  int m_line_number;
+  int m_line_offset;
```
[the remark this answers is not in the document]
xTVaser: Ah, I agree -- but this looks like a left-over from when i first started the work.  Since im only doing decompiler use-cases for now, this isn't needed for now (the method that sets these vars isn't even called).

Putting it on the `Type` makes more sense, Somewhat related -- I'd like to see more fields as well, mainly the type/functions docstring.  If this is set in `all-types.gc` then we can carry it forward to the output of the decompiler and the LSP can show the docstring on hover information.

lsp/handlers/lsp_router.cpp
xTVaser: Huh, hadn't noticed that because it indeed does work 
image
nlohman::json must define some custom fmt::formatters

water111: ah, nice!

lsp/state/data/mips_instruction.cpp
water111: yeah it would require some work and probably isn't worth it unless you want to do more stuff with the assembly ops.

water111 (approved):

PR #1653, opened 2022-07-15
[graphics] Another possible way to handle windowed

water111:
In this mode, you can resize the window and pick the resolution from the menu.  When the resolution is selected, it will adjust the window to that resolution, but still allows you to resize it as needed.

All the old letterboxing/aspect ratio stuff works like normal.

I tested this for a bit on linux and it seemed to work pretty well, but not sure if there's windows-specific things to watch out for.


ManDude (approved):
This looks like a better way to do it, thanks!

PR #1660, opened 2022-07-16
Menu cleanup

Zedb0T:
1. __Remove "back" option in title-pc:__ The quit game option should replace the back button on the pc port when not in a game state. Having a back option does not really make sense unless you are in a nested progress-menu.

There were 6 options in title-pc I think this looks MUCH cleaner and is much  clearer.
If the end user wishes to close the title-pc progress menu they can press square or triangle to achieve this.

2. __Disable autosave now lives in game-settings__, This might be the speedrunner in me and its old location messing with my menuing muscle memory, however I think this setting makes more sense nested into game-settings.

3. __Move secrets into a newly defined options-pc:__ This is only for title, elsewhere the game uses game-options-pc, however its probably nice to instance this so in the future more optimizations can be added. (also less clutter on the default opening menu)

4. Enable Daxter intro CS skip in retail mode on PC_PORT

5. Enable L2 'boost if cheatmode is on __**and**__ PC PORT 

Just a nice little thing.


ManDude:
I don't think it makes sense to move that stuff to "game options." The secrets are not game options, they're that: secrets, extra features. The other Jak games don't put them inside "game options" so I don't see why they should be here. And all other games I know put this kind of stuff in its own separate menu as well. The option would eventually be placed in the main options menu as well once it gets more features that are safe to be used in normal gameplay. Jak 2 and 3 have this option in the title menu and base menu.

I also don't think it makes sense for disable auto-save to be in the game options. It's not a togglable option, and it's not intuitive for it to be completely separate from the other save/load options. It's... 1 extra button press, I don't think it's worth making the menu layout confusing over this. If your concern is menu clutter here, you can make a save & load menu and put all this stuff there maybe.

Don't really see the point in removing the "back" option from the title screen, it does the same as you pressing start so it's not like it's doing anything special here. Probably shouldn't outright remove options from the original game if they're not being completely useless IMO. Jak 2 and 3 don't have the option, but they also don't let you leave the title menu. 6 options doesn't really make the menu cluttered.

Zedb0T:
Imo the menu is very cluttered but its not a top priority either, Here you mention

> it does the same as you pressing start so it's not like it's doing anything special here. Probably shouldn't outright remove options from the original game if they're not being completely useless IMO. 

Is this button not completely useless now? Don't think of it as removing the back button think of it more as replacing the PS2s version "Back" button with a more appropriate "quit game" button for the PC port. It doesn't make sense to have both a quit game and back button and is unclear imo.

> I also don't think it makes sense for disable auto-save to be in the game options. It's not a togglable option, and it's not intuitive for it to be completely separate from the other save/load options. It's... 1 extra button press, I don't think it's worth making the menu layout confusing over this. If your concern is menu clutter here, you can make a save & load menu and put all this stuff there maybe.

The issue here I don't think is the one extra menu press its just the disable autosave menu feeling super out of place, if I want to enable and or disable a feature that effects the game (vibration/play-hints etc) I think it should all be done from a central location/menu, not  just seemingly randomly showing up on a main pseudo default menu if its on or not. I am always going to be for more uniformed game/ui/menu design.

 I could move it into the save/load progress screens but again I really feel  it doesn't belong in their either. To me this setting/feature is the same type as vibration/play-hints and should be grouped in a menu with them.

> I don't think it makes sense to move that stuff to "game options." The secrets are not game options, they're that: secrets, extra features. The other Jak games don't put them inside "game options" so I don't see why they should be here. And all other games I know put this kind of stuff in its own separate menu as well. The option would eventually be placed in the main options menu as well once it gets more features that are safe to be used in normal gameplay. Jak 2 and 3 have this option in the title menu and base menu.
Yeah this was the only one I did not feel that strongly about, I think its fine on the title screen. That being said I do think eventually it probably still is a good idea to extend out and have a custom options-pc, however there is no need to jump the gun early and add it before its justified

ManDude:
> Imo the menu is very cluttered , but more importantly is is also not clear where to find what settings or options where.

What kind of options are you referring to here?

> Is this button not completely useless now?

Redundant, but not useless. Useless is stuff like "center screen" which does not do anything. All menus can be exited with square or triangle, but they all still have a "back" button and I don't think we should remove that if the original game insisted on having it.

> The issue here I don't think is the one extra menu press its just the disable autosave menu feeling super out of place, if I want to enable and or disable a feature that effects the game (vibration/play-hints etc) I think it should all be done from a central location/menu, not just seemingly randomly showing up on a main pseudo default menu if its on or not. I am always going to be for more uniformed game/ui/menu design.

If this is what you're concerned about then you should just keep the option but have it grayed out when it can't be used. Having it buried in a complete separate menu is confusing and doesn't really address your uniformness issue.

It fundamentally works different to all the other settings, and I don't think many games (if any) put it far away from the usual save & load.

Zedb0T:
Yeah - I just thought this was busy work that was planned on didn't realize you wanted the options there. Was just trying to clean and remove some busy work so you all could focus on more important stuff - we don't need to "waste" time on this.

while I think they are the least important thing in this PR, are you against the title-cutscene skip and PC_PORT cheat mode 'boost?

If you are we can just close this, but if you aren't opposed I can revise this PR.

ManDude:
Nah, I personally don't mind the cheat changes.

PR #1661, opened 2022-07-16
[high fps] fix target-racer

Brent-Hickey:
- Deceleration when holding x + square
- Deceleration when holding square
- Turning/sliding speed

https://github.com/open-goal/jak-project/issues/1499#issue-1277190691


ManDude:
These changes are interesting because that means 50Hz has a lot more differences than we thought.

NolanHewitt:
Turning at 150fps still feels slower than turning at 60fps to me, playing the latest 0.1.2 release.

PR #1663, opened 2022-07-16
Pc port debug small qol

Zedb0T:
- Enable L2 'boost on racer if cheatmode **AND** PC_PORT
- Enable Daxter intro CS skip in title on retail mode if PC_PORT
- Fix typo to be consistent with other instances in code.
- Add empty jak1 folder in ISO_DATA (can revert this if need be)

https://user-images.githubusercontent.com/89345505/179365660-2739f7fe-5a89-48aa-bb1e-85546ed638b6.mp4


xTVaser (changes requested):
If you want to include the `jak1/` folder, you need to duplicate the `.gitignore` not move it, else there is the risk of other folders in `iso_data` being committed / unstaged.

Zedb0T:
I would need to add `!/jak1` 
So the * wildcard doesn't ignore the empty `jak1` directory correct?

xTVaser:
probably better to do a `!**/.gitignore`

water111 (approved):

PR #1667, opened 2022-07-16
[graphics] some blit optimizations

water111:
This is a pretty big performance increase when msaa is off, and a medium increase when it's on. I can play at 2x on my desktop now, and my laptop is closer to getting 60fps,


ManDude (approved):

Adamillo:
Hooefully this optimization improves the waterfall performance on my end...

PR #1669, opened 2022-07-17
Cleanup goalc tests, fix jak2 kernel bugs

water111:
- Split a few of the giant goalc test files. Remove the `all_goalc_template_tests.cpp` file and use individual files. After the cleanup, this no longer helped compile times.
- Clean up includes, restructure so we don't the `inja` include for tests that don't generate GOAL code - this saves a lot of time compiling.
- Set up the jak 2 kernel test, fixed a bug in the `catch-frame`.

### DIFF
--- a/decompiler/config/jak2/all-types.gc
+++ b/decompiler/config/jak2/all-types.gc
@@ -426,20 +426,24 @@
   :flag-assert #x90000000c
   )
 
+;; a "catch" frame is a frame that can be "thrown" to.
+;; the "throw" is a nonlocal control flow back to the state befor the "catch" block.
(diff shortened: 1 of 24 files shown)


decompiler/config/jak2/all-types.gc
```diff
@@ -426,21 +426,25 @@
   :flag-assert #x90000000c
   )
 
+;; a "catch" frame is a frame that can be "thrown" to.
+;; the "throw" is a nonlocal control flow back to the state befor the "catch" block.
```
ManDude, line 430: before

PR #1672, opened 2022-07-17
RNG - mimic Java's Util.Random

dallmeyer:
While working on Zed's checkpoint randomizer, we started to notice some predictable patterns in the checkpoint lists built using the `(rand-vu)` functions (e.g. checkpoint A was often followed by checkpoint B).

This is an initial stab at mimicking the logic from Java's util.Random class to offer some alternative RNG functions. 

https://docs.oracle.com/javase/8/docs/api/java/util/Random.html


goal_src/jak1/engine/math/math.gc
```diff
@@ -317,3 +317,67 @@
...
+)
+
+(defun java-rand-advance-seed ()
+  ;; seed =  (seed * 0x5DEECE66DL + 0xBL) & ((1L << 48) - 1)
+  (set! *java-rand-seed* (the uint64 (logand (+ (* *java-rand-seed* 25214903917) 11) (- (shl 1 48) 1))))
+)
+
+(defun java-rand-next ((bits int))
+  ;; TODO: figure out why we need to offset bits like this, probably getting converted to uint32 somewhere or something
+  (+! bits 16)
```
dallmeyer, line 337: Need some help debugging this, seemed like uint64s were getting downgraded to uint32s possibly, and then the `(shr *java-rand-seed* (- 48 bits))` was not working as it does in the Java library

ManDude:
Global values on the symbol table only have 32 bits of space.

dallmeyer:
> Global values on the symbol table only have 32 bits of space.

Ah, yep that would explain it. I will go re-read the docs, but offhand is there a good way to store 64 bits in a global? Or would I just need to have 2 32 bit ones and shift/mask to concatenate them?

ManDude:
You will have to store them in a more indirect way. You can allocate an 8-byte buffer and keep the `uint64` pointer to that in the symbol table, then dereference that (`(-> *uint64-buf*)`) to get the value.

water111:
I'd recommend making something like
```
(deftype rng-state (structure)
  ((seed uint64)))
  
(define *rng-state* (new 'global 'rng-state))

;; then access it like:
(-> *rng-state* seed)
```

dallmeyer:
Oh perfect, let me try that - thanks both!

goal_src/jak1/engine/math/math.gc
```diff
@@ -317,3 +317,68 @@
   `(+ -1.0 (the-as float (logior #x3f800000 (/ (rand-uint31-gen ,gen) 256))))
   )
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;; mimicking Java's util.Random
+;;;; https://docs.oracle.com/javase/8/docs/api/java/util/Random.html
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(define *java-rand-seed* (the uint64 42069))
```
dallmeyer, line 324: do we have a system time or anything like that accessible, that we could use here as a more dynamic default seed?

dallmeyer: worst case: `0x666edd1e`? 😉

dallmeyer: Hmm I think math.gc is handled before all_types.gc, getting an error `Type Error: Type scf-time is not fully defined.` if I try to use `scf-time` within math.gc 😞

dallmeyer:
> I'd recommend making something like
> 
> ```
> (deftype rng-state (structure)
>   ((seed uint64)))
>   
> (define *rng-state* (new 'global 'rng-state))
> 
> ;; then access it like:
> (-> *rng-state* seed)
> ```

Hm @water111 even with this it still looks like it's sometimes dropped to 32 bits depending on what value is stored?

```
gc> (-> *java-rand-state* seed)
281473529806044        #xffffa9c1fcdc             -0.0000        #<invalid object #xa9c1fcdc>

gc> (shr (-> *java-rand-state* seed) 24)
16777129        #xffffa9              0.0000        #<unknown type 1 @ #xffffa9>
```
VS 
```
gc> (-> *java-rand-state* seed)
1939231159        #x739651b7        23819025000000000000000000000000.0000        #<invalid object #x739651b7>

gc> (shr (-> *java-rand-state* seed) 24)
115        #x73              0.0000        #<invalid object #x73>
```

goal_src/jak1/engine/math/math.gc
Hat-Kid: Yeah, `math.gc` is one of the earliest files that gets compiled and `scf-time` is defined in `settings-h.gc`.
You'd have to move the type definition of `scf-time`.

water111:
Those 2 examples look right to me.

One confusing thing is that multiplication is 32-bit in GOAL, to match the original hardware.  If you want to multiply 64-bit numbers, you should use `(imul64 a b)` instead of `(* a b)`.

dallmeyer:
> Those 2 examples look right to me.
> 
> One confusing thing is that multiplication is 32-bit in GOAL, to match the original hardware. If you want to multiply 64-bit numbers, you should use `(imul64 a b)` instead of `(* a b)`.

Yeah sorry you're right (I was up til 5am last night trying to sort through this, my brain is on E right now)
And okay bingo that multiplication was the issue, before that I was getting biased results but now it looks random. Will push changes in a sec after some more testing.

On another note - any opposition to moving the definition of `scf-time` into math.gc (or something compiled before it) so I can use it for the initial seed?

dallmeyer:
Thanks all for the help - this seems to be working as intended now and is a pretty close mirror of the Java library I think. 

I will defer to y'all if you want to actually include it in vanilla openGOAL, or use it in place of `(rand-vu)` calls throughout.

dallmeyer:
So I did some tests with the `vu-rand` functions, generating ints between [0,100) and while the distribution of generated ints is fairly even, there are terrible patterns in subsequent random values...

with `(rand-vu-int-range 0 100)`
vu RNG

with `(java-rand-nextIntRange 0 100)`
java-mimic RNG

https://docs.google.com/spreadsheets/d/1Mt46rMLzrKA8hyFb2zs7F5E-j0dC6PokQyVQItM7QVI/edit#gid=461023415

xTVaser:
Yeah, this is what we would expect from the current OpenGOAL RNG because we do not properly track the value of the `Q` register which is a key part of making the randomness more distributed in the original game, and even then it still isn't great.

dallmeyer:
Right yeah, makes sense - I imagine there is some hesitance in replacing all the `vu-rand` calls throughout the game with these new methods I've written, but any concern with merging them in so they could be used for any future development?

xTVaser:
imo no nothing major to prevent the merging for that purpose -- having some actually good RNG functions is definitely useful.  I'd probably suggest removing the `java` from the symbols as it kinda looks out of place, even the java doc says that the algorithms come out of Knuth book
> which is modified using a linear congruential formula. (See Donald Knuth, The Art of Computer Programming, Volume 2, Section 3.2.1.)

dallmeyer:
Cool, I can do that. How about `knuth-rand-xyz` then to give credit?

xTVaser:
That's fine by me, seems reasonably descriptive, naming is always hard -- if the names feel weird down the line we can always change them.

dallmeyer:
Alright updated the function names and the related comments

water111:
I'm fine with the random functions, but would rather not move the definition of `scf-time` here

dallmeyer:
> I'm fine with the random functions, but would rather not move the definition of `scf-time` here

- removed redefinition of `scf-time` in `math.gc`
- moved the `scf-time-to-uint64` utility into `settings-h.gc` next to the `scf-time` definition - let me know if there's a better place for this?
- moved all this new RNG stuff into a separate file since it introduces the dependency on `settings-h`, so that I don't need to change the dep order for `math.gc`

goal_src/jak1/game.gp
```diff
@@ -1696,6 +1696,7 @@
  "engine/connect.gc"
  "ui/text-h.gc"
  "game/settings-h.gc"
+ "math/knuth-rand.gc"
```
ManDude, line 1699: Add a note saying it's a new file here.

ManDude (changes requested):

decompiler/config/all-types.gc
```diff
@@ -6286,7 +6286,7 @@
   :size-assert         #x8
   :flag-assert         #x900000008
   )
-
+;;(define-external scf-time-as-uint64 (function none))
```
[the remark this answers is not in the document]
dallmeyer: Ah yeah whoops forgot to clean this up

water111 (approved):
Looks good, sorry this took us so long to get to. Thanks for moving this to a separate file

PR #1676, opened 2022-07-18
Depth Cue

Francessco121:
Everyone's favorite screen blur is here!

The renderer is fairly straight-forward, hopefully I left good enough comments explaining everything. Usually, the game would switch between creating the effect and then overlaying it onto the screen 16 times (it does 1/16 of the screen at a time), however that's super slow in OpenGL so the renderer does the whole effect in just 2 draws by making some assumptions about how the effect works.

As for performance, it's pretty cheap. On my GPU (Vega 56) it takes ~0.05ms @ 1080p. There's a single framebuffer blit that can theoretically be removed (I had some issues with it), but removing it only drops the frame-time to ~0.04ms.

~~I was thinking maybe this could be added as an anti-aliasing option. The 'MSAA' menu option could be made into a general AA setting that includes a 'Depth Cue (PS2)' or just 'PS2' option. The effect does work correctly with MSAA, but it's kinda pointless and just makes it look blurry. Either way, this should definitely be disabled by default.~~
The fully ported effect is barely noticeable above the game's native resolution, not worth keeping as an AA option in its default state.

**Remaining Issues:**
- [x] Renderer doesn't respect alpha value given by the GOAL code
- [x] Renderer doesn't respect `xyoffset` GS register
- [x] Profiler draw count/tri count stats
- [x] Clean up


water111 (approved):
Looks great, thanks for including all the documentation, everything was very easy to follow.

PR #1677, opened 2022-07-18
implement a cheats menu + a few bug fixes

### LINKED ISSUE #1640: [game] dont switch language while streamed audio is playing
Because sound banks get updated depending on language now we should just hold off changing while the loader is busy.

### LINKED ISSUE #1662: [Game] FPS option not respected on boot
### Acknowledgements

### Describe the Bug

After setting the FPS option to anything but 60FPS, when next booting the game it will be limited to 60FPS again and the setting has to be re-selected.

### How To Reproduce

Choose 100FPS or 150FPS option, reboot the game and check FPS.

Not needed, bug is obvious

### Expected Behavior

FPS option to stick.

### Operating System

Windows 10 21H2

### OpenGOAL Version

0.1.19

### Game Version

NTSC 1.0 (black label)

Yes

### LINKED ISSUE #1679: [runtime] vsync doesnt work sometimes

ManDude:
It shows up once you've 100%'d the game, or found any cheat codes yourself. Changed the behavior of the disable auto-save option to always be there but look locked when it's already disabled.

Increases sprite cap from 8x to 12x (needed for LPC), increases DMA buffer sizes (needed for LPC), and forces envmap when PS2 lods aren't being used (generic2 needs more memory). Fixes a crash with the battle hud.

- Fixes #1662 
- Fixes #1640 
- Fixes #1679


ManDude:
dunno what's going on, but `execute-connections` is getting corrupted somehow.

water111 (commented):
Looks good, I'm a little concerned about performance with the lod stuff. From what I understand it will force all envmapped stuff (orbs, wooden crates) to always render with generic (much, much slower than merc) and completely disables the draw distance check.

ManDude:
I can add a separate setting for envmap if the performance hit is too large.

water111:
> I can add a separate setting for envmap if the performance hit is too large.

Yeah, if it's not a huge pain, I think it would be nice to have a separate option to use ps2 defaults for envmap.

PR #1683, opened 2022-07-19
fix borderless on Windows, resolution when letterboxes and envmap setting

ManDude:
Fixes the 1 pixel hack on Windows borderless, makes the letterbox not part of the render framebuffer and correctly sets the default value for the envmap setting (I forgot to change that when I flipped its purpose).

### DIFF
--- a/game/graphics/opengl_renderer/OpenGLRenderer.cpp
+++ b/game/graphics/opengl_renderer/OpenGLRenderer.cpp
@@ -566,11 +566,21 @@ void OpenGLRenderer::setup_frame(const RenderOptions& settings) {
     }
   }
 
-  glBindFramebuffer(GL_FRAMEBUFFER, m_fbo_state.render_fbo->fbo_id);
   ASSERT_MSG(settings.game_res_w > 0 && settings.game_res_h > 0,
              fmt::format("Bad viewport size from game_res: {}x{}\n", settings.game_res_w,
                          settings.game_res_h));
-  glViewport(0, 0, settings.game_res_w, settings.game_res_h);
+
+  if (!m_fbo_state.render_fbo->is_window) {
+    glBindFramebuffer(GL_FRAMEBUFFER, 0);
+    glViewport(0, 0, m_fbo_state.resources.window.width, m_fbo_state.resources.window.height);
+    glClearColor(0.0, 0.0, 0.0, 0.0);
+    glClearDepth(0.0);
+    glDepthMask(GL_TRUE);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+    glDisable(GL_BLEND);
+  }
+
+  glBindFramebuffer(GL_FRAMEBUFFER, m_fbo_state.render_fbo->fbo_id);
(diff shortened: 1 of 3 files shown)


game/graphics/opengl_renderer/OpenGLRenderer.cpp
```diff
@@ -749,16 +751,16 @@ void OpenGLRenderer::do_pcrtc_effects(float alp,
...
+    glBlitFramebuffer(0,                                                          // srcX0
+                      0,                                                          // srcY0
+                      window_blit_src->width,                                     // srcX1
+                      window_blit_src->height,                                    // srcY1
+                      render_state->draw_offset_x,                                // dstX0
+                      render_state->draw_offset_y,                                // dstY0
+                      render_state->draw_offset_x + render_state->draw_region_w,  // dstX1
+                      render_state->draw_offset_y + render_state->draw_region_h,  // dstY1
+                      GL_COLOR_BUFFER_BIT,                                        // mask
+                      GL_LINEAR                                                   // filter
```
water111, line 763: I think this will not work if you go into windowed mode, resize the window to be not-16:9, then select aspect ratio 16:9. The reason for the "squash" thing earlier is so this blit has the same size source/destination, which ends up being way faster, and is required to resolve the multi-sampled stuff.

ManDude: It does work fine:

image

Not sure about performance though.

water111: Hm, if I do:
- delete pc-settings.txt
- start game, it goes to fullscreen
- set to windowed
- resize window
- aspect ratio 16x9
- resize window
I get a black screen with this error:
```
[2022-07-18 21:14:14:072] [error] [] OpenGL error 0x502 S8246 T824C: GL_INVALID_OPERATION error generated. Source and destination dimensions must be identical with the current filtering modes.
```

which is from the `glGlitFramebuffer` with `0, 0, 864, 778, 0, 146, 864, 632` as the arguments.

ManDude: Ah, it does happen if you have MSAA enabled.

ManDude: image

There we go. 👍

game/graphics/opengl_renderer/OpenGLRenderer.cpp
```diff
@@ -548,11 +548,9 @@ void OpenGLRenderer::setup_frame(const RenderOptions& settings) {
           make_fbo(settings.game_res_w, settings.game_res_h, settings.msaa_samples, true);
       m_fbo_state.render_fbo = &m_fbo_state.resources.render_buffer;
 
-      bool resolution_matches =
-          window_fb.width == settings.game_res_w && window_fb.height == settings.game_res_h;
       bool msaa_matches = window_fb.multisample_count == settings.msaa_samples;
 
-      if (!resolution_matches && !msaa_matches) {
+      if (!msaa_matches) {
```
water111, line 553: good catch

water111 (approved):
Thanks, I gave this a try and everything looks good!

PR #1684, opened 2022-07-19
Use libco for cooperative threading in overlord

Ziemas:
PR'd for opinions, just did this for fun. 
It's a bit cleaner than having real threads messing around with cv's and mutexes.

Idk if there's any significant benefit/drawback though.

TODO:
lots of testing
returning from a thread is UB in libco, make sure this doesn't happen


game/overlord/iso.cpp
```diff
@@ -658,11 +658,6 @@ u32 ISOThread() {
     ProcessMessageData();
 
     if (!read_buffer) {
-      // HACK!! sometimes when we want to exit, some other threads will wait for stuff to be loaded
-      // in such cases, we continue running until we're the last thread alive when it's safe to die
-      if (ThreadWantsExit(GetThreadId()) && OnlyThreadAlive(GetThreadId())) {
```
water111, line 663: Unfortunately I think this hack might be needed to shutdown if you close the window during a cutscene.

water111 (commented):
Nice, this looks awesome!  

From what I can see, the advantages are:
- simplifies the IOP_Kernel stuff quite a bit
- probably quite a bit faster (the condition variable wait shows up as like 5% of a core in vtune)
- possibly helps with future overlord, which does more stuff with semaphores

The disadvantages are:
- might cause problems with debuggers/profiling tools? Not sure about this. I'd expect gdb to work fine but who knows about windows. It might mess up stack traces in vtune too, but not really a big deal because the IOP stuff is fast anyway
- There's some platform specific stuff that could cause weird problems or break in the future.  That said, the GOAL->C interface already relies on most of the same stuff.
- Probably requires some testing on windows and linux.

My opinion is that we should do this, but test carefully on linux/windows first.

game/overlord/iso.cpp
Ziemas: What would go wrong in such a case? It seems to close properly.

Ziemas:
Did some quick testing on windows and it seems to function fine, at least with MSVC builds (my clang is broken for some reason).

Ziemas:
MSVC seems to handle debugging these cothreads perfectly fine

game/overlord/iso.cpp
ManDude: Try closing the window while a cutscene is playing while some music is fading out, that's what this fixed originally.

Ziemas: I don't think it can be a problem now. If we're in this branch DelayThread will return control to the kernel and we'll end up at the exit condition check in iop_runner.

water111 (approved):
Looks good, lets give this a try!

PR #1687, opened 2022-07-20
Some more high FPS fixes

Francessco121:
At high FPS, fixes:
- Rat game camera rotation (was too fast)
- Boggy swamp 'balance' platforms (accelerated too fast, see below)
- Baby spiders (wiggle too fast, same issue that swamp rats had)

Regarding the balance platforms however, this doesn't 100% fix it... I'm not entirely sure why, I think it has to due with Jak registering as riding on the platform on more consecutive frames compared to 60. If you print out the platform state, you can see that some frames while the platform is going down Jak will technically be floating above it and the platform won't accelerate for that frame. Either way, this makes the physics much much closer to 60fps. I can remove it from the PR if this isn't good enough however.


goal_src/jak1/engine/camera/cam-states.gc
```diff
@@ -636,15 +636,15 @@
...
           (cond
-            ((< (-> *CAM_BILLY-bank* rot-speed) (-> s5-0 y))
-             (set! (-> s5-0 y) (-> *CAM_BILLY-bank* rot-speed))
+            ((< (* DISPLAY_FPS_RATIO (-> *CAM_BILLY-bank* rot-speed)) (-> s5-0 y)) ;; changed for high fps
+             (set! (-> s5-0 y) (* DISPLAY_FPS_RATIO (-> *CAM_BILLY-bank* rot-speed))) ;; changed for high fps
              )
-            ((< (-> s5-0 y) (- (-> *CAM_BILLY-bank* rot-speed)))
-             (set! (-> s5-0 y) (- (-> *CAM_BILLY-bank* rot-speed)))
+            ((< (-> s5-0 y) (- (* DISPLAY_FPS_RATIO (-> *CAM_BILLY-bank* rot-speed)))) ;; changed for high fps
+             (set! (-> s5-0 y) (- (* DISPLAY_FPS_RATIO (-> *CAM_BILLY-bank* rot-speed)))) ;; changed for high fps
```
ManDude, line 647: I think this should actually take lag into account too. The fp camera has the same issue.

Francessco121: Should be good now, seems to work correctly when lagging and the code is consistent with the other high FPS changes to cam-states.

PR #1689, opened 2022-07-20
Implement scheduling of IOP threads.

Ziemas:
This is an experiment on top of #1684 that implements more realistic scheduling of threads and a new strategy for throttling the IOP.

Given that the current solution is working well I'm not very certain if this is a good idea.

Threads are now scheduled like the kernel on the real IOP does them, the highest priority thread in ready state is always ran. For this to work we now also respect the argument to DelayThread and leave them in wait state until the specified time runs out.

This also requires changing the method by which we throttle the IOP kernel, wait_run_iop now uses information from the scheduler to wait until the kernel has a new thread ready to run.

### DIFF
--- a/game/runtime.cpp
+++ b/game/runtime.cpp
@@ -254,8 +254,8 @@ void iop_runner(SystemThreadInterface& iface) {
   while (!iface.get_want_exit() && !iop.want_exit) {
     // the IOP kernel just runs at full blast, so we only run the IOP when the EE is waiting on the
     // IOP. Each time the EE is waiting on the IOP, it will run an iteration of the IOP kernel.
-    //iop.wait_run_iop();
-    iop.kernel.dispatch();
+    auto delay = iop.kernel.dispatch();
+    iop.wait_run_iop(delay);
   }
 }
 }  // namespace
--- a/game/runtime.cpp
+++ b/game/runtime.cpp
@@ -254,7 +254,7 @@ void iop_runner(SystemThreadInterface& iface) {
   while (!iface.get_want_exit() && !iop.want_exit) {
     // the IOP kernel just runs at full blast, so we only run the IOP when the EE is waiting on the
     // IOP. Each time the EE is waiting on the IOP, it will run an iteration of the IOP kernel.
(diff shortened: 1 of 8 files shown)


Ziemas:
This seems to actually work pretty well. It's possibly even better than just #1684 on its own. I'll compare more against master as well later.

Ziemas:
Played with this for a while on linux and did a quick test on windows, found no issues. Also took a peek in vtune and the IOP thread is basically invisible.

water111 (commented):
Nice, this is a much cleaner solution than what I had before.

My only concern is that the `iop_run_cv.wait_until` will end up sleeping for too long in some cases. (For example, on a windows laptop that's not plugged in, the timer resolution becomes very bad).  But I think this won't really be a problem because so much other stuff is doing a `signal_run_iop` that will wake up that thread sooner.

Tested on linux, CPU usage is around the same or lower (hard to tell because the IOP usage is so much lower after the switch to libco).

I was able to get a black screen on startup (in particular when running the ASAN build):
```
IRX version 2.0
STALL: waiting for IOP on RPC port #1
Loading bank ~/jak-project/out/jak1/iso/COMMON.SBK
Created bank 1
STALL: waiting for IOP on RPC port #1
```

It looks like `sceSifCheckStat` is being called in a loop and always returning 1 (busy). It's for `0xdeb2`, which is the `RPC_Loader` function.

My best guess at this point is that the `WakeupThread` in `IOP_Kernel::sif_rpc` is somehow being overwritten.

Ziemas:
Ah yeah, I can reproduce that after a bit of restarting. I guess I was right to be worried about calling WakeupThread from the EE thread.

Ziemas:
Hmph, even having a lock around changing the thread state I still get a deadlock occasionally. Not really sure why yet.

Ziemas:
I added a queue to put wakeup requests on when they come from the EE thread, this seems to solve it.

water111:
oops, I missed that you updated this. I couldn't recreate the problem after your fix.  Putting the wakeup in a queue like that seems like the right way to fix this.

PR #1695, opened 2022-07-21
Fix typo in `game.gp` when building test zone

ManDude:


LuminarLight:
Wasn't the long name intentional?

ND themselves use 3 characters long DGO names for levels that have vis (vi1, cwi, fea, etc...), and up to 8 characters long DGO names for levels that don't have vis (halfpipe, demo, drillmtn, casext, etc...).

While it technically does not matter if the name is 3 or 8 (or less) characters long, testzone is supposed to be an example level, and it has no vis, so according to ND practices its DGO should have the "long" name.


I doubt people will care once they start making custom levels, but testzone is still an/the example. I am just wondering if the above information was considered before making this change?

ManDude:
The .gd file used the name `TSZ`, there was discrepancy here and a few people got confused when making their levels.

The name here is meant to refer to the name of the DGO that is being built.

ManDude:
We don't really understand how the visless files work yet, and the custom level builder makes vis-format levels (since they're the only ones we know anything about right now).

water111:
The "long name" levels can be created and loaded with the level builder, but they can't be loaded at the same time as a "short name" level. I wanted to support loading custom levels and existing levels at the same time, so I made the "short name" levels that happen to not have visibility data.

LuminarLight:
Ok, thanks for the explanations. Things are clearer now!

PR #1697, opened 2022-07-22
Set thread names of system threads

Ziemas:
Helps when looking at debug tools.

### DIFF
--- a/game/system/SystemThread.cpp
+++ b/game/system/SystemThread.cpp
@@ -1,6 +1,19 @@
 #include "SystemThread.h"
 
 #include "common/log/log.h"
+#include "common/util/unicode_util.h"
+
+#ifdef __linux
+#include <pthread.h>
+#else
+// Include order matters...
+// clang-format off
+#define NOMINMAX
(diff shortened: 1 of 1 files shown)


game/system/SystemThread.cpp
```diff
@@ -82,6 +94,15 @@ bool SystemThreadManager::all_threads_exiting() {
 void* bootstrap_thread_func(void* x) {
   SystemThread* thd = (SystemThread*)x;
   SystemThreadInterface iface(thd);
+
+#ifdef __linux__
+  pthread_setname_np(pthread_self(), thd->name.c_str());
+#else
+  std::wstring tmp = std::wstring(thd->name.begin(), thd->name.end());
+  LPCWSTR thname = tmp.c_str();
+  SetThreadDescription(GetCurrentThread(), thname);
```
xTVaser, line 103: could maybe simplify with https://github.com/open-goal/jak-project/blob/6ce1d2a0c7f59580953cbb2d77d39af7ded24cd4/common/util/unicode_util.cpp#L18

PR #1700, opened 2022-07-24
Add in-game option to switch fullscreen monitor

### LINKED ISSUE #1454: [runtime] Fullscreen/Borderless does work for primary monitor only
When switching from Windowed to Fullscreen/Borderless it automatically fullscreen on the primary monitor detected on boot, instead of the current active monitor like most of pc games. That means in order to change monitor you must close OG, change primary monitor and reopen it.
Even when using programs that let you switch monitor of the active window (e.g. Dual Monitor Tools) it stays fixed on the primary monitor, even with Borderless (that should technically be like a maximized window, and so movable with DMT).

Francessco121:
Worth mentioning that the selected location is referred to as 'monitor' in code but 'display' on the UI. Monitor is consistent with GLFW's terms, but display makes more sense to users. The options are also listed as just 'DISPLAY 1', 'DISPLAY 2', etc. I would have preferred to show the monitor's name but I wasn't sure how to get those strings into GOAL.

Let me know if I went a little overboard on the refactoring, I ran into some issues tracking the previous windowed pos/size when going from fullscreen to windowed, so I changed it quite a bit.  

Also, I had to modify the ImGui source, their GLFW implementation hijacks the monitor changed callback but never calls our registered callback (unlike the other events they steal). I have no idea why that code is even in ImGui's current release since their callback is empty and marked as such...

Closes #1454

### DIFF
--- a/goal_src/jak1/engine/ui/progress/progress-static.gc
+++ b/goal_src/jak1/engine/ui/progress/progress-static.gc
@@ -151,7 +151,7 @@
   )
 
 ;; maps options to a progress screen
-(define *options-remap* (new 'static 'boxed-array :type (array game-option) :length 0 :allocated-length (#if (not PC_PORT) 35 61)))
+(define *options-remap* (new 'static 'boxed-array :type (array game-option) :length 0 :allocated-length (#if (not PC_PORT) 35 60)))
 
 ;; TODO probably an enum.
 ;; maps level-info indices to the appropriate offset in *level-task-data*
--- a/goal_src/jak1/engine/ui/progress/progress-static.gc
+++ b/goal_src/jak1/engine/ui/progress/progress-static.gc
@@ -151,7 +151,7 @@
   )
 
 ;; maps options to a progress screen
-(define *options-remap* (new 'static 'boxed-array :type (array game-option) :length 0 :allocated-length (#if (not PC_PORT) 35 60)))
+(define *options-remap* (new 'static 'boxed-array :type (array game-option) :length 0 :allocated-length (#if (not PC_PORT) 35 61)))
 
 ;; TODO probably an enum.
 ;; maps level-info indices to the appropriate offset in *level-task-data*
--- a/third-party/imgui/imgui_impl_glfw.cpp
+++ b/third-party/imgui/imgui_impl_glfw.cpp
@@ -197,8 +197,12 @@ void ImGui_ImplGlfw_CharCallback(GLFWwindow* window, unsigned int c)
     io.AddInputCharacter(c);
 }
 
-void ImGui_ImplGlfw_MonitorCallback(GLFWmonitor*, int)
+void ImGui_ImplGlfw_MonitorCallback(GLFWmonitor* monitor, int event)
 {
+  ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
+  if (bd->PrevUserCallbackMonitor != NULL)
+    bd->PrevUserCallbackMonitor(monitor, event);
+
 	// Unused in 'master' branch but 'docking' branch will use this, so we declare it ahead of it so if you have to install callbacks you can install this one too.
 }
 
--- a/decompiler/config/all-types.gc
+++ b/decompiler/config/all-types.gc
@@ -1597,6 +1597,8 @@
   (scene-255                    #x12ff)
   (hint-0                       #x1300)
   (hint-511                     #x14ff)
+  (display                      #x1600)
+  (display-fmt                  #x1601)
 ;; GAME-TEXT-ID ENUM ENDS
   )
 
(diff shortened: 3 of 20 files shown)


third-party/imgui/imgui_impl_glfw.cpp
```diff
@@ -197,8 +197,12 @@ void ImGui_ImplGlfw_CharCallback(GLFWwindow* window, unsigned int c)
     io.AddInputCharacter(c);
 }
 
-void ImGui_ImplGlfw_MonitorCallback(GLFWmonitor*, int)
+void ImGui_ImplGlfw_MonitorCallback(GLFWmonitor* monitor, int event)
 {
+  ImGui_ImplGlfw_Data* bd = ImGui_ImplGlfw_GetBackendData();
+  if (bd->PrevUserCallbackMonitor != NULL)
+    bd->PrevUserCallbackMonitor(monitor, event);
+
```
water111, line 205: wow, was imgui just missing this originally?

water111 (commented):
Sorry, I totally missed this PR!  If I forget about stuff like this in the future feel free to ping me on discord or in the PR.

Unfortunately I'm away this weekend so I can't test right now, so if somebody else with multiple monitors could give it a try, that would be helpful.

Some things that I'd recommend trying:
- try the "reboot in debug mode" option a few times, just to make sure we can still recreate the window. (the logic here all looks good, but it's easy to have bugs here)
- make sure we aren't adding too many more glfw function calls per frame.  At least on my computer they take 0.1ms each so we spent a pretty significant amount of time doing these. As far as I can tell this PR doesn't add any.
- unplug/connect monitor as the game is running. make sure that glfw window size/pos callbacks happen if the OS moves the window off of a monitor that was just disconnected. You can get a similar effect by changing the resolution (at least in windows, this forces the window positions to all be reassigned)

third-party/imgui/imgui_impl_glfw.cpp
Francessco121: Yep, looking at the current release you can find the same empty function here https://github.com/ocornut/imgui/blob/v1.88/backends/imgui_impl_glfw.cpp#L397. Absolutely no idea why it's like this.

Francessco121:
No worries! I've been a bit busy myself. I did test the debug mode switch and unplugging monitors while the game was fullscreen on the monitor that got unplugged and such (actually ran into several issues with both tests originally), so it should be fine. It's been a bit tho so I'll double check.

For the additional glfw calls, I tried to avoid making any during each frame using the new glfw callbacks and just keeping track of the state. I'll double check this as well tho.

ManDude (changes requested):
I only have 1 monitor, so I can't do thorough testing, but it seems to work.

You need to add your new menu to the `can-go-back?` function so users can press start or square to exit the menu.

goal_src/jak1/engine/ui/progress/progress-static.gc
```diff
@@ -151,7 +151,7 @@
   )
 
 ;; maps options to a progress screen
-(define *options-remap* (new 'static 'boxed-array :type (array game-option) :length 0 :allocated-length (#if (not PC_PORT) 35 60)))
+(define *options-remap* (new 'static 'boxed-array :type (array game-option) :length 0 :allocated-length (#if (not PC_PORT) 35 61)))
```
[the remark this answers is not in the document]
Francessco121: Yep, turns out 60 is indeed enough.

Francessco121:
Thanks for the feedback, I've addressed each comment.

I changed up how the fullscreen and minimized states are retrieved, the GLDisplay class now tracks them as fields so their states can be retrieved without calling GLFW. Turns out that this PR did originally add an extra GLFW call per frame, doing this removes that. However, I wasn't able to fully remove all GLFW calls used each frame to check if the fullscreen mode needs to change. Unfortunately, GLFW doesn't have a callback for video mode changes and without that or checking it each frame the game has no idea if you change your monitor's resolution for example. At least on Windows however, this call only takes ~0.01ms.

I also did the following tests:
- Clicking 'restart in debug mode' multiple times in each fullscreen mode on different monitors.
- Unplugging the monitor that the game is fullscreened (or borderless) on (game moves over to the primary monitor).
- Changing the resolution of the monitor that the game is fullscreened (or borderless) on.
    - Note: The resolution that the game renders at doesn't get affected by this, but GLFW will be given the correct video mode when this changes so that the game understands the active monitor's dimensions.
- Unplugging the monitor that the game is fullscreened on and forcing it to default to a different monitor with a different resolution.
- Same unplugging tests but with the Windows monitor 'Project' feature.

I think everything should be all set. Please let me know if I committed any C++ sins tho, I'd be happy to change any of that.

xTVaser:
add to whitelist

ManDude (approved):

PR #1701, opened 2022-07-24
Fix bg-custom making level-load-info :music-bank not work.

LuminarLight:
With these changes, I played around in the game, checking ND intro, some cutscenes, some gameplay, and haven't noticed any difference.
BUT now if you set :music-bank in level-info for your custom level, and use bg-custom, the music will work. Before this change it kept playing the music that was loaded at the time when bg-custom was used, even after respawning and things like that.

The `(set! (-> *level* play?) #f)` line in the bg-custom function is probably important, I tried commenting it out and then game became unstable in custom levels. So instead I investigated why/how 'play?' is related to music, and found that one line that I modified.

I use bg-custom very often, and this music issue always bothered me.

This may not be the safest change, but I have no way to see everything that am I affecting with this change (that may be impossible even). So if this is rejected, or a different solution is suggested, I will understand.

If my explanation of the issue is unclear/messy, please let me know, and I will try to give a clearer answer.

### DIFF
--- a/goal_src/jak1/engine/game/main.gc
+++ b/goal_src/jak1/engine/game/main.gc
@@ -635,7 +635,7 @@
           (dotimes (a0-8 (-> *level* length))
             (let ((a1-6 (-> *level* level a0-8)))
               (when (= (-> a1-6 status) 'active)
-                (if (and (= (-> a1-6 name) v1-13) (-> *level* play?))
+                (if (= (-> a1-6 name) v1-13)
                     (set! (-> *setting-control* default music) (-> a1-6 info music-bank))
                     )
                 )


ManDude:
I think it's better to just remove the stuff in here <https://github.com/open-goal/jak-project/blob/master/goal_src/jak1/engine/debug/default-menu.gc#L4822>. If you wanna play a level then it makes more sense that you'd have play mode enabled. If the load boundaries are being a bother then just move the level away from them, I guess.

LuminarLight:
Indeed if I remove the '(set! (-> *level* play?) #f)' line, then music works, but of course if I move around then game crashes due to the load triggers. And indeed, as you suggested, a possible solution to this would be to move my level away from the load triggers.

I am just unsure if removing this line is indeed a good solution. I see all the people putting their custom level right inside Sandover Village for example, maybe that is why the disablation of load triggers was implemented in bg-custom.

LuminarLight:
I don't think this is very important, I can live with it, so I am closing the PR. Indeed there are many easy workarounds for my main issue.

ManDude:
Don't worry we'll eventually have it working properly once custom level implementation is more fleshed out.

Issue #1612, opened 2022-07-05
Camera rotation and attacking by using the mouse


MojoJojoDojo:
### What feature is your idea related to?

The feature relates to controls for the game.

### Describe the solution you'd like.

Have the abillity to use the mouse for rotating the camera and attacking with its 2 (or 3) buttons.
Maybe with the help of some 3rd party library.

### Additional context

Good for people that dont have access to controllers.

PR #1702, opened 2022-07-25
Added first working instance of controller/keyboard re-mapper

### LINKED ISSUE #1600: Movement using arrow keys

Personally I hate using WASD. I much prefer ↑←↓→ .

Please add the ability to use either.

### Additional context

Thanks for all your great work.

animalstyletaco:
Marking as a draft because I want to add unit tests, add some better documentation on my changes, and see if I can iron out some minor inconsistencies first.

In short, this would give users the ability to do the following:
 * Re-map Controller buttons 
 * Re-map Keyboard buttons
 * Re-map keyboard buttons to analog (Fixes #1600)
 * Re-map Mouse buttons
 * Scroll wheel/touch pad (Will need to do some additional testing since I don't have a touch pad available)

On startup, the program will attempt to read the configuration file game/assets/config/controllers-settings.json. The program will generate a json file with the default configuration if a file doesn't exist. The Default configurations are below:

Controller Default Settings are the same.

Keyboard Default Configuration:
> L1 - Q
> R1 - O
> L2 - 1
> R2 - P
> X - Spacebar
> Square - F
> Triangle - R
> Circle - E
> Dpad - arrows
> Left analog - wasd
> Right analog - Mouse
> Start - Enter
> L3 - .
> R3 - ,

Values are stored as integers in the json file. This is subject to change if needed. In most cases, it represent the GLFW_KEY defines and GLFW_GAMEPAD define. Mouse buttons indices are treated as a extension of default GLFW_KEY defines. For example, left mouse button = 348. The same applies for scroll wheel axes and mouse axes. They are treated as extensions of the GLFW_GAMEPAD defines.

Other changes:
* Disables mouse cursor when ImGui debug menu is closed. Allowing for mouse camera control.
* Disables in-game left mouse button actions and re-enables mouse cursor while ImGui debug menu is open.


animalstyletaco:
Would also fix #1612

game/graphics/pipelines/opengl.cpp
```diff
@@ -92,10 +97,64 @@ void SetDisplayCallbacks(GLFWwindow* d) {
...
+      Pad::OnKeyPress(key);
+    } else if (action == GlfwKeyAction::Release) {
+      Pad::OnKeyRelease(key);
+    }
+  });
+
+  glfwSetScrollCallback(d, [](GLFWwindow* window, double xoffset, double yoffset) {
+    Pad::SetAnalogAxisValue(GlfwKeyCustomAxis::SCROLL_WHEEL_X_AXIS, xoffset);
+    Pad::SetAnalogAxisValue(GlfwKeyCustomAxis::SCROLL_WHEEL_Y_AXIS, yoffset);
+    printf("Scroll Callback: xoffset: %lf, yoffset: %lf\n", xoffset, yoffset);
```
water111, line 129: can probably remove this

game/graphics/pipelines/opengl.cpp
```diff
@@ -186,6 +245,9 @@ static std::shared_ptr<GfxDisplay> gl_make_display(int width,
     lg::error("Could not load icon for OpenGL window");
   }
 
+  glfwSetInputMode(window, GLFW_CURSOR, g_cursor_input_mode);
+  glfwSetInputMode(window, GLFW_STICKY_MOUSE_BUTTONS, GLFW_TRUE);
```
water111, line 249: Does this mess up imgui menus?

game/graphics/pipelines/opengl.cpp
```diff
@@ -529,6 +591,12 @@ void GLDisplay::render() {
     ImGui_ImplOpenGL3_RenderDrawData(ImGui::GetDrawData());
   }
 
+  // switch vsync modes, if requested
+  if (Gfx::g_global_settings.vsync != Gfx::g_global_settings.old_vsync) {
```
water111, line 595: why move this (just curious)?

game/system/newpad.cpp
```diff
@@ -107,64 +139,93 @@ int IsPressed(MappingInfo& mapping, Button button, int pad = 0) {
...
   if (key == -1)
     return 0;
   auto& keymap = mapping.buffer_mode ? g_buffered_key_status : g_key_status;
   ASSERT(key < NUM_KEYS);
   return keymap[key];
 }
 
+void SetAnalogAxisValue(int axis, double value) {
+  if (axis == GlfwKeyCustomAxis::CURSOR_X_AXIS || axis == GlfwKeyCustomAxis::CURSOR_Y_AXIS) {
+    value /= 18.0f;  // Arbituary value. Cursor delta generally a lot larger than scroll wheel delta
```
water111, line 153: Will the framerate have an effect on the cursor sensitivity? If the framerate is super high, the delta cursor positions will be smaller.

game/graphics/gfx.cpp
```diff
@@ -53,36 +52,162 @@ GfxGlobalSettings g_global_settings;
...
+                                                           {"Square", Pad::Button::Square}};
+
+const std::pair<std::string, Pad::Analog> analog_map[] = {
+    {"Left X Axis", Pad::Analog::Left_X},
+    {"Left Y Axis", Pad::Analog::Left_Y},
+    {"Right X Axis", Pad::Analog::Right_X},
+    {"Right Y Axis", Pad::Analog::Right_Y},
+};
+
+void DumpToJson(const std::string& filename) {
```
water111, line 78: we've been trying to use `filesystem::path` when possible for filenames. I think you can just switch this argument to use `filesystem::path`

water111 (commented):
This looks awesome, thanks for working on this!

I added a few small comments, let me know when you are ready for more detailed review/testing. I think we should try this on windows/linux with a few different mice, just to make sure the sensitivity is reasonable and there's not weird polling-rate issues.

I can test on linux within the next few days

game/graphics/gfx.cpp
animalstyletaco: Ok, I'll update this when I get a chance

game/system/newpad.cpp
animalstyletaco: Great catch! I only tested with 60 frames on my initial tests. Higher frame rate would reduce the delta cursor positions. I was thinking about making the delta configurable since different mouse(s) could have different DPIs. Would having a mouse sensitivity value in the json be a good idea?

game/graphics/pipelines/opengl.cpp
animalstyletaco: This was an accident, I was moving chunks code around so I could better visualize what was going on in the source code and this change must have snuck into my commit.

game/graphics/pipelines/opengl.cpp
animalstyletaco: Will do

game/system/newpad.cpp
water111: Yeah, that sounds like the best option.

game/graphics/pipelines/opengl.cpp
animalstyletaco: It doesn't mess with the imgui. This code snippet will display the cursor when ever the imgui menu is up and disable it when the menu is closed. Although, there is a weird bug where the cursor is there only on startup regardless if imgui is there or not.
https://github.com/animalstyletaco/jak-project/blob/c5e3ab71b595abe5fdf7ca57782844578e3db989/game/graphics/pipelines/opengl.cpp#L100-L102

animalstyletaco:
This will fix #1712

ManDude (changes requested):
I tested around a bit and it works. Thank you for adding analog stick mapping!

game/graphics/pipelines/opengl.h
```diff
@@ -18,30 +18,38 @@ enum GlfwKeyAction {
...
+  int get_screen_vmode_count() override;
+  int get_monitor_count() override;
+  void set_size(int w, int h) override;
+  void update_fullscreen(GfxDisplayMode mode, int screen) override;
+  void render() override;
+  bool minimized() override;
   bool fullscreen_pending() override;
   void fullscreen_flush() override;
-  void set_lock(bool lock);
+  void set_lock(bool lock) override;
```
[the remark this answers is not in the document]
animalstyletaco: I mainly did this to get rid of the warnings I saw when compiling on Linux

game/system/newpad.cpp
```diff
@@ -43,6 +47,37 @@ u64 input_mode_mod = 0;
 u64 input_mode_index = 0;
 MappingInfo g_input_mode_mapping;
 
+MappingInfo g_mapping;  // Local copy of Gfx mapping info
```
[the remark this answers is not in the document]
animalstyletaco: I remember it was a little awkward in the beginning using the mapping in g_settings for analog settings. Mostly because I wasn't familiar with this part of the code base at the time. I'll see if I can use g_settings now that I'm more familiar with the code

game/graphics/gfx.cpp
```diff
@@ -51,38 +50,184 @@ namespace Gfx {
...
-void SaveSettings() {
-  const auto filename = file_util::get_file_path({GAME_CONFIG_DIR_NAME, SETTINGS_GFX_FILE_NAME});
-  file_util::create_dir_if_needed(file_util::get_file_path({GAME_CONFIG_DIR_NAME}));
-  FILE* fp = file_util::open_file(filename.c_str(), "wb");
-  fwrite(&g_settings, sizeof(GfxSettings), 1, fp);
-  fclose(fp);
+void SavePeripheralSettings() {
+  auto filename =
+      (file_util::get_jak_project_dir() / "game" / "config" / "controller-settings.json");
+  file_util::create_dir_if_needed_for_file(filename);
```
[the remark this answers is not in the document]
animalstyletaco: Ok that makes sense. I just chose a place at random for initial testing.

game/system/newpad.cpp
```diff
@@ -113,66 +149,114 @@ int IsPressed(MappingInfo& mapping, Button button, int pad = 0) {
...
-  }
-
-  if (input == 0) {
+  if (g_gamepads.gamepad_idx[pad] == -1) {
+    input = g_key_analogs[pad][(int)analog];
+    // Hack. Clearing the buffer immediately can lead to inconsistencies on analog input.
+    // If a mouse is disconnected or can't calculate a new delta it would stay stuck at 1.
+    // Decreasing the values gradually seems like a good comprise.
+    g_key_analogs[pad][(int)analog] *= 0.95;
+  } else {  // Gamepad present
```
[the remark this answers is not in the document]
animalstyletaco: Just for clarification. Are you expecting something like this in the code.

```cpp
  float input = 0.0f;
  if (CheckPadIdx(pad) == -1) {
    // Pad out of range, return a stable value
    return 127;
  }

  if (g_gamepads.gamepad_idx[pad] > -1) {
    input += g_gamepad_analogs[pad][(int)analog];
  }
  input += g_key_analogs[pad][(int)analog];
  // Hack. Clearing the buffer immediately can lead to inconsistencies on analog input.
  // If a mouse is disconnected or can't calculate a new delta it would stay stuck at 1.
  // Decreasing the values gradually seems like a good comprise.
  g_key_analogs[pad][(int)analog] *= 0.95;
```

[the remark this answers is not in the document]
ManDude: Sort of. You can track key and controller analogs separately (and thus decay keyboard just fine), but when asked for analog value you pick whichever one is "most active."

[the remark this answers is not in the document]
animalstyletaco: So something like this?

```cpp
  float input = 0.0f;
  if (CheckPadIdx(pad) == -1) {
    // Pad out of range, return a stable value
    return 127;
  }

 float controller_input, keyboard_input;
controller_input = keyboard_input = 0.0f;
  if (g_gamepads.gamepad_idx[pad] > -1) {
    controller_input = g_gamepad_analogs[pad][(int)analog];
  }
  keyboard_input = g_key_analogs[pad][(int)analog];
  g_key_analogs[pad][(int)analog] *= 0.95;

if(fabs(controller_input) > fabs(keyboard_input)){
   input = controller_input;
} else {
   input = keyboard_input;
}
 glfw_to_goal_value(input); //pseudo code

```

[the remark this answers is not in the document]
ManDude: Yeah, that's sort of what I was thinking as well.

game/graphics/gfx.cpp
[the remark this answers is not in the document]
animalstyletaco: Should there be a controller/kbm config file for each game or a global config for all games?

game/graphics/gfx.cpp
```diff
@@ -41,8 +42,6 @@ void InitSettings(GfxSettings& settings) {
   settings.pad_mapping_info.buffer_mode = true;
   // debug input settings
   settings.pad_mapping_info.debug = true;
-  // use a default mapping
-  Pad::DefaultMapping(settings.pad_mapping_info);
```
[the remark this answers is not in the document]
animalstyletaco: I was trying to reduce the amount of calls to this function to the absolute minimum. Although, after re-factoring the code I ended up putting the call back in the startup function.

game/graphics/gfx.cpp
```diff
@@ -51,38 +50,184 @@ namespace Gfx {
...
+    LoadPeripheralSettings(filename);
+    Pad::SetMapping(g_settings.pad_mapping_info);
+    lg::info("Loaded graphics configuration file.");
+    return;
+  }
+
+  file_util::create_dir_if_needed_for_file(filename);
+  Pad::DefaultMapping(g_settings.pad_mapping_info);
+  SavePeripheralSettings();
+  lg::info("Created graphics configuration file {}", filename.string());
```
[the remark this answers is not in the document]
animalstyletaco: I'm a little confused by this. The program would create a new default controller-settings.json only if the file doesn't exist.  If an example controller-settings.json doesn't exist how would users know how to configure their controller/kbm settings?

If this isn't the appropriate approach, what would be the best approach to show users how controller/kbm configuration works?

game/graphics/gfx.cpp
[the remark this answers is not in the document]
ManDude: I imagine global for all games makes the most sense.

game/graphics/gfx.cpp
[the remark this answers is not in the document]
ManDude: There will be an in-game menu for key rebinding. Users won't be editing a text file.

ManDude (approved):

game/system/newpad.cpp
```diff
@@ -113,67 +148,123 @@ int IsPressed(MappingInfo& mapping, Button button, int pad = 0) {
...
   if (key == -1)
     return 0;
   auto& keymap = mapping.buffer_mode ? g_buffered_key_status : g_key_status;
-  ASSERT(key < NUM_KEYS);
+  ASSERT(key < glfw::NUM_KEYS);
   return keymap[key];
 }
 
+void SetAnalogAxisValue(MappingInfo& mapping_info, int axis, double value) {
+  const double sensitivity_numerator = 100.0f;
```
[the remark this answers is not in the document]
animalstyletaco: That's a good idea I'll implement it when I get a chance. Do you remember where the target frame rate is stored in the c++ code?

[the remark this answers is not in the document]
ManDude: Global settings has a `target_fps` value I believe.

game/system/newpad.cpp
```diff
@@ -43,6 +47,51 @@ u64 input_mode_key = -1;
...
+    60.0f;  // frame rate reference so mouse sensitivity can be consistent on frame rate changes
+
+void SetFrameRate(float frame_rate) {
+  const float minimum_frame_rate = 0.0001f;  // Arbituary value
+  if (frame_rate < minimum_frame_rate) {
+    g_frame_rate = minimum_frame_rate;
+  } else {
+    g_frame_rate = frame_rate;
+  }
+}
```
ManDude, line 60: You should just use `Gfx::g_global_settings.target_fps` directly instead of having to manage a copy and a new function that just sets it.

game/system/newpad.cpp
```diff
@@ -114,67 +164,123 @@ int IsPressed(MappingInfo& mapping, Button button, int pad = 0) {
...
+          mapping_info.keyboard_analog_mapping[pad][analog].negative_key < glfw::NUM_KEYS) {
+        if (analog == static_cast<int>(Analog::Left_Y) ||
+            analog == static_cast<int>(Analog::Right_Y)) {
+          input =
+              input +
+              g_buffered_key_status[mapping_info.keyboard_analog_mapping[pad][analog].negative_key];
+        } else {
+          input =
+              input -
+              g_buffered_key_status[mapping_info.keyboard_analog_mapping[pad][analog].negative_key];
```
ManDude, line 252: use shorthand assingments here?

game/graphics/gfx.cpp
```diff
@@ -53,38 +54,179 @@ namespace Gfx {
...
+  }
+}
+
+void LoadSettings() {
+  auto filename = (file_util::get_user_config_dir() / "controller" / "controller-settings.json");
+  if (fs::exists(filename)) {
+    LoadPeripheralSettings(filename);
+    lg::info("Loaded graphics configuration file.");
+    return;
+  }
```
water111, line 227: I think you're missing something to create the settings file on the first run if it doesn't exist.
I added an `else { SavePeripheralSettings(); }` here and it seemed to work.

game/graphics/pipelines/opengl.cpp
```diff
@@ -209,6 +214,11 @@ static std::shared_ptr<GfxDisplay> gl_make_display(int width,
   }
 
   auto display = std::make_shared<GLDisplay>(window, is_main);
+
+  display->set_imgui_visible(Gfx::get_debug_menu_visible_on_startup());
+  display->update_cursor_visibility(window, display->is_imgui_visible());
+  glfwSetInputMode(window, GLFW_STICKY_KEYS, GLFW_TRUE);
```
water111, line 220: Is this needed? This turned out to be the cause of the weird double-alt-callback on linux.
From https://www.glfw.org/docs/3.3/input_guide.html it sounds like this only changes stuff with `glfwGetKey`, which we don't use.

water111 (changes requested):
Sorry I keep forgetting about this...

Just two more comments and I think it is all set.

I'm not personally a fan of taking over the mouse by default when the game first starts, but idk a better way to do this. I guess we already go fullscreen by default so it's not much more annoying to also grab the mouse.

game/graphics/gfx.cpp
animalstyletaco: I originally had a default mapping settings file generated if it couldn't find a controller-settings.json file. Although, it seemed like mandude wasn't very fond of the idea at the time. https://github.com/open-goal/jak-project/pull/1702#discussion_r938782693. I'll re-add the logic to create a new setting file back into the code.

game/graphics/pipelines/opengl.cpp
animalstyletaco: Gotcha, I'll remove it

game/graphics/gfx.cpp
water111: Hm, once there's an in game menu, maybe we can revisit this, but I think for now this is good. Otherwise I have no idea what all the things in the JSON should be called.

PR #1711, opened 2022-07-29
Decompiler: Cleanup CLI and make overriding the config much easier

xTVaser:
I need to be able to specify what objects i want decompiled easily, and the existing overriding code only really worked for flags.  The way i did it before was finding the empty json array and writing the values into the file with a python script (gross).

With this i can just provide a JSON string on the CLI and have it merge with the base config:

```
--config-override '{"allowed_objects": ["ash"]}'
```

I also cleaned up some of our scripts that are either not needed anymore, or won't be used anymore


water111 (approved):
this is much nicer, thanks!

PR #1715, opened 2022-07-30
Automatically switch to dedicated GPU in OpenGOAL runtime

doctashay:
Some machines with dual-GPUs cannot properly detect OpenGOAL as a high-performance application and properly utilize the dedicated GPU. This sets a flag for both Radeon and NVIDIA GPUs to always use the high performance GPU when available. I do not have a dual-GPU machine to test this on, but this should just work out of the box, and doesn't seem to cause any issues for machines that aren't utilizing dual-GPUs.

### DIFF
--- a/game/main.cpp
+++ b/game/main.cpp
@@ -45,8 +45,6 @@ void setup_logging(bool verbose) {
   lg::initialize();
 }
 
-
-
 /*!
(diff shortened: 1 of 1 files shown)


xTVaser:
add to whitelist

PR #1724, opened 2022-08-04
tests: Cleanup decompiler test fixtures and try to improve the test output / add a readable report

xTVaser:
Reports will showup in the wrong spot unfortunately due to an underlying github issue - https://github.com/dorny/test-reporter/issues/67


water111 (approved):
Thanks! This is definitely a cleaner way that what I was originally thinking (single fixture with jak1/jak2 functions)

PR #1725, opened 2022-08-05
decomp/lsp: Differentiate warnings from likely/definite errors

xTVaser:
Many warnings are truly errors (ie. this function has no type) which makes flagging on warnings not very accurate.

This tries to solve that, making many warnings that are typically errors/failures prefixed with `ERROR:` instead.


decompiler/Function/Warnings.h
```diff
@@ -80,14 +86,16 @@ class DecompWarnings {
...
+          return fmt::format("ERROR: {}\n", message);
         case Kind::EXPR_BUILD_FAILED:
-          return fmt::format("WARN: Expression building failed: {}\n", message);
+          return fmt::format("ERROR: Expression building failed: {}\n", message);
         case Kind::CFG_FAILED:
-          return fmt::format("WARN: CFG building failed: {}\n", message);
+          return fmt::format("ERROR: CFG building failed: {}\n", message);
         case Kind::TYPE_PROP_FAILED:
-          return fmt::format("WARN: Type Propagation failed: {}\n", message);
+          return fmt::format("ERROR: Type Propagation failed: {}\n", message);
```
water111, line 98: If you wanted, I think this might be simpler if there were just 3 types: INFO, WARN, ERROR

decompiler/IR2/AtomicOpTypeAnalysis.cpp
```diff
@@ -1433,13 +1433,13 @@ TypeState StackSpillLoadOp::propagate_types_internal(const TypeState& input,
...
   if (info.size != m_size) {
-    env.func->warnings.general_warning(
+    env.func->warnings.general_error(
         "Stack slot load at {} mismatch: defined as size {}, got size {}", m_offset, info.size,
         m_size);
   }
 
   if (info.is_signed != m_is_signed) {
-    env.func->warnings.general_warning("Stack slot offset {} signed mismatch", m_offset);
+    env.func->warnings.general_error("Stack slot offset {} signed mismatch", m_offset);
```
water111, line 1442: this one might be better as a warning - the GOAL compiler actually gets this wrong in a few places

test/decompiler/reference/jak1/engine/camera/cam-layout_REF.gc
```diff
@@ -286,17 +286,17 @@
...
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
+;; ERROR: Stack slot offset 164 signed mismatch
```
water111, line 299: like these errors - I think it's just a bug in the original compiler

water111 (approved):
Looks good! I just had a few suggestions, feel free to merge after taking a look

PR #1728, opened 2022-08-05
all-types: Improve Jak 2's all-types

xTVaser:
This tries to improve the all-types for jak2 in a few ways:
- ~~put an object's method states in `(:states ... )` with the proper name instead of making a dummy method~~
  - just properly name and type the state methods.  It would be nice to also initialize the `:states` table, but this would rely on the naming convention.  Which would probably be accurate but, eh.
- put proper type names on symbols when possible (a basic / function / state) instead of `object` always
- name dummy methods `<type-name>-method-<id>` so it is obvious what type they are associated with and that it is a method call
- hopefully properly omit `heap-base` if it doesn't match the automatic one (the only one that ended up being so seems to be `ocean-wave-frames`

This does _not_ try to fill in types that were not able to have their fields read as I don't see a general solution to this.


ManDude:
Don't our state methods go into the method table and not the state table? State table is just for simpler state forward declaration.

ManDude:
Also maybe we want to use `int64` instead of `uint64` for the 64-bit load (you can't tell whether it's unsigned or not). For Jak 1 I had to manually change a lot of these.

decompiler/analysis/analyze_inspect_method.cpp
```diff
@@ -920,7 +920,18 @@ std::string inspect_inspect_method(Function& inspect_method,
   result.flags = flags.flag;
   result.type_size = flags.size;
   result.type_method_count = flags.methods;
-  result.type_heap_base = flags.heap_base;
+
+  // Only set heap-base if it's different from the automatic one
+  // A child (or child of a child) of process ALWAYS has heap-base set.
+  if (flags.heap_base > 0) {
+    // TODO - how can i get this value in a non-hardcoded way
+    // - this is the size of a jak 2 "process"
```
water111, line 928: Maybe
```cpp
auto process_type = dts.ts->get_type_of_type<BasicType>("process");
auto auto_hb = (flags.size - process_type->size() + 0xf) & ~0xf;
```
I think that this requires us to have `process` in `all-types.gc` but we should have that already.

water111: If you want a per-game constant you can add something like this to `goal_constants.h`
```
constexpr PerGameVersion<int> GOAL_PROCESS_SIZE(0xwhatever, 0x80);
```
then `GOAL_PROCESS_SIZE[game_version]` will give you the right size.

decompiler/analysis/analyze_inspect_method.cpp
```diff
@@ -1148,13 +1163,21 @@ std::string TypeInspectorResult::print_as_deftype(
 
   result.append(fmt::format("  :method-count-assert {}\n", type_method_count));
   result.append(fmt::format("  :size-assert         #x{:x}\n", type_size));
+  if (type_heap_base.has_value()) {
+    result.append(fmt::format("  :heap-base           #x{:x}\n", type_heap_base.value()));
```
water111, line 1167: can you check to see if this case is ever hit or not? If not, want to just replace this with an ASSERT?

decompiler/analysis/analyze_inspect_method.cpp
xTVaser: I remember trying this early on and i got an error, ill give it a shot and if it doesn't work ill use the second option.

decompiler/analysis/analyze_inspect_method.cpp
xTVaser: It's hit once:
```lisp
(deftype ocean-wave-frames (structure)
  ((frame ocean-wave-data 64 :offset-assert 0) ;; guessed by decompiler
   )
  :method-count-assert 9
  :size-assert         #x0
  :heap-base           #x1
  :flag-assert         #x900010000
  )
```

water111: ah ok. Then leave it as is.

The `ocean-wave-frames` thing is a bug in the original compiler - the size of the type is larger than `UINT16_MAX` but they only use a `u16` to store the size.

I think this happened in jak 1 as well.

PR #1736, opened 2022-08-06
goal_src: Get rid of `*OLD*` files in jak1's goal_src

xTVaser:
I also missed adding the placeholder to the only files we've done in jak 2 ( the kernel ones )


ManDude:
thanks (I don't care about these files)

PR #1737, opened 2022-08-07
[wip] game: add remap menu

Hat-Kid:
Adds a new "Input Options" menu to the progress menu that lets you rebind all the cpad buttons to keyboard or controller actions.

- The "device" setting at the top toggles between controller and keyboard bindings.
  - If that setting is set to "Controller" and a controller is connected, the keyboard is completely disabled in-game. Will be toggleable with the "Disable Keyboard" button once that is wired up.
  - If a controller is not connected, however, the keyboard is still enabled regardless of the setting, in order to not completely lock you out without a controller.
- Remap buttons automatically unselect without changing anything after a timeout of 4 seconds (or when pressing Escape).
- Handles swapping binds if the pressed key is already bound to something else.
- Button to reset every binding to its default value.
- Only lets you remap `cpad 0` for now. Most people will not require changing `cpad 1`, anyway.
- Will save all mappings to `pc-settings.gc` (commented out for now).

### Stuff that still needs fixing

- **Big issues**
  - The biggest problem right now is to do with the input mode. When pressing a remap button, input mode is activated and a timeout of 4 seconds will activate.
    If a button/key is pressed in that time, the code for that button should be read, written to `input_mode_joy` or `input_mode_key` in C++ and mapped to the corresponding cpad button.
    However, what actually currently ends up happening is that `input_mode_joy` is either always `-1` even though it is being set in `OnJoyPress` in C++ or it instantly binds itself to the pressed button after selecting an option.
    Keyboard does not have this issue even though it basically works the same way using `input_mode_key`.
- **Small things**
  - The "Disable Keyboard" button does not do anything yet.
  - After binding a key, the first key press is not registered, but everything works fine after that.
  - In order to avoid creating a separate menu for controller and keyboard bindings, I conditionally set the `value-to-modify` of the `pad-remap` buttons based on the value of `(-> *pc-settings* pad?)`.
    However, because `init-game-options` is only called once (when the progress menu is initialised), the options do not update when changing `(-> *pc-settings* pad?)`.
    To remedy this, I call `init-game-options` when `pad-selection` is changed, however, this temporarily messes up the text for the buttons until an option is selected.

### Other notes

- Decided to postpone analog remapping for now (and I imagine we probably don't need analog remapping at all on controllers?).
- Analog keys on keyboard are still hardcoded right now, as I wanted to focus on them once I have the input mode issue fixed.
- It's possible that I broke the `pc-pad-utils.gc` file, but I haven't tried it out yet.


xTVaser:
Merged this branch into https://github.com/open-goal/jak-project/pull/2397 and I'm seeing it through to completion (getting there!) Most of the things that were called out I've not had to deal with yet / have already solved / or have intentionally deviated from because the design is slightly different.

PR #1746, opened 2022-08-08
[WIP] Mouse hook

towai:
Admittedly I wrote all this nearly a month ago and haven't really touched it since, but while I'm updating my repo I figured I may as well make a draft PR so anyone interested can take a look at these changes even though they're very unfinished. It's a requested feature, but at this stage I'm not making any attempt to link it to the camera or menu logic at all, since changing the mouse mode doesn't quite work properly.

Demonstrated best in this video, though regrettably it's probably at bit unclear when exactly I'm clicking and using alt-tab:

https://user-images.githubusercontent.com/12366892/183504586-84c39205-140d-4bc0-a348-c5816bf39da4.mp4

The present issue as I described when I recorded the video, which should help to comprehend what's going on in it:
> - when the game was just paused, if GLFW_CURSOR was GLFW_CURSOR_DISABLED, i have the polling process (same one that prints to the screen; yet to separate these into separate processes) set it to GLFW_CURSOR_NORMAL, and vice versa when the game is unpaused
> - this works as expected on pause, but on unpausing, the cursor sticks around until you move it, at which point it disappears, but you can still move it around on the desktop. however, GLFW does not update the cursor position at all in this state. clicking on the game window does nothing (except the title bar in windows, which moves the window and restores the GLFW_CURSOR_DISABLED behavior when you finish dragging. i don't have a linux machine to test).
> - if you click on a different window (at which point the cursor is moved back to its last position when it was GLFW_CURSOR_NORMAL) and then click back into the game window, then GLFW will start updating the mouse position again.
> - before the process starts, i store a variable that's used to find whether it was paused on the previous frame (by capturing it right before doing (suspend)). literally i do (let* ((last-paused (paused?)) (procp (process-spawn-function [etcetc])))). it seems like this variable is always true when the process first runs, though.

TODO:
- [ ] add a GOAL function that tells the game whether the imgui bar is visible, so that on unpause the game can know to only hide the cursor again if it isn't.
- [ ]  add a function that sets the mouse position, if deemed necessary?


towai:
Closing this as it's sort of obsoleted by #1702, though i still think some work that exists in this and not that such as GOAL functions for interfacing with the mouse should be needed in the future, and that simply emulating the right stick with the mouse is very much the wrong way of going about it — i would just have to create a new PR that builds off those changes instead.

PR #1747, opened 2022-08-09
Implement IOP semaphores.

Ziemas:
These are really only used as locks in jak1, I'm not sure there's even any change in behavior from this. I think we'll want these for jak 2 though.

Might want to give this some testing, I didn't do much tbh.


water111 (approved):
Logic looks good to me, lets give this a try.

PR #1750, opened 2022-08-10
fix invincibility cheat codes + some japan-specific changes

### LINKED ISSUE #1739: [game] check language instead of territory for japanese logo
I think it'd be cooler if we changed this.

### LINKED ISSUE #1740: [game] remove language cheat
Unnecessary.

### LINKED ISSUE #1741: [game] japanese version uses wrong date format
Should be Y/M/D and not M/D/Y.

### LINKED ISSUE #1745: Invincibility bug.
### Acknowledgements

### Describe the Bug

Turning invincibility on causes no damage to be taken, however, if switched off from the secret menu, does not disable invincibility.

however, after a forced death, vincibility is restored, and you can take damage.
Also, saving then reloading also restores the ability to be hurt, both morally and physically.

### How To Reproduce

```
Steps to reproduce:

- One:     Turn on invincibility from the secrets menu.
- B:       Attempt to take damage and fail
- Tree:    Proceed back to the secrets menu to undo your mistake.
- See:     That you cannot be hurt again.
- And:     Go back to the secrets menu and weep as you confirm it is disabled, and yet you cannot be hurt.
```

Didn't check

### Expected Behavior

When the invincibility cheat is disabled, damage can be taken.

### Operating System

Windows 10 21H2 Build:19044.1826

### OpenGOAL Version

v0.1.22

### Game Version

PAL

Yes

ManDude:
- fixes #1745 
- fixes #1741 
- fixes #1740 
- fixes #1739


water111 (approved):
But now we won't hear the sound when entering the language cheat code!

PR #1754, opened 2022-08-12
custom levels: add support for more res types in actor lumps

Hat-Kid:
Adds support for `vector` and `float` in actor `res-lump` tags in custom level JSON files.
Also added some small documentation for currently available types to use in the `test-zone` JSON file.

### DIFF
--- a/goalc/build_level/Entity.cpp
+++ b/goalc/build_level/Entity.cpp
@@ -131,7 +131,7 @@ std::unique_ptr<Res> res_from_json_array(const std::string& name,
       data.push_back(json_array[i].get<float>());
     }
     return std::make_unique<ResFloat>(name, data, -1000000000.0000);
-  } else if (array_type == "floatm") {
+  } else if (array_type == "meters") {
     std::vector<float> data;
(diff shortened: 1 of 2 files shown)


xTVaser:
add to whitelist

goalc/build_level/Entity.cpp
```diff
@@ -113,6 +113,30 @@ std::unique_ptr<Res> res_from_json_array(const std::string& name,
...
+      data.push_back(vectorm4_from_json(json_array[i]));
+    }
+    return std::make_unique<ResVector>(name, data, -1000000000.0000);
+  } else if (array_type == "float") {
+    std::vector<float> data;
+    for (size_t i = 1; i < json_array.size(); i++) {
+      data.push_back(json_array[i].get<float>());
+    }
+    return std::make_unique<ResFloat>(name, data, -1000000000.0000);
+  } else if (array_type == "floatm") {
```
water111, line 134: This might be better as `meters` to be consistent with the GOAL types.

water111 (commented):
Thanks for adding, I didn't realize I forgot these. 
Just one suggestion on naming

PR #1756, opened 2022-08-13
HUD - show total/level counts on dpad up/down

dallmeyer:
Some of my thoughts:
- ~~Should it be a toggle instead of just while holding up/down? Personally I kind of like that it behaves like the original unless you're actively holding another button~~ switched to toggles, see comments below
- ~~Better ways for displaying it? Not crazy about splitting across lines~~ Made it easier to scale text printed with `draw-string-xy`, looks pretty good now I think. Also tested while on zoomer (with heat shield and eco meter visible) and with hud-pc-parts visible (in ambush missions, showing remaining enemies)
- Should probably compute totals dynamically like in [calculate-completion](https://github.com/open-goal/jak-project/blob/8a18072d9794fbd50e0b5b36f4569125535b1f4c/test/decompiler/reference/jak1/engine/ui/progress/progress_REF.gc#L292) instead of hardcoding (I can't steal these values from `*progress-process*` because we're not in the progress menu 😔)

https://user-images.githubusercontent.com/2515356/185327680-f2e43422-066f-49f0-826d-c63c0082e84a.mp4


ManDude:
Personal suggestions:
- use a toggle on press instead of on hold.
- play a sound on a toggle!
- make the HUD hide before changing. makes the change more obvious.
- you can honestly just ditch the totals.
- the code you changed was not properly decompiled. it's better to clean up a function that you're going to change before writing your custom stuff. here's how that original method actually looked more like for example:
```lisp
(defmethod draw-hud hud-money ((obj hud-money))
  (let ((t9-0 (method-of-type hud draw-hud)))
    (t9-0 obj)
    )
  (with-dma-buffer-add-bucket ((buf (-> (current-frame) global-buf))
                               (bucket-id debug))
    (draw-string-xy (string-format "~D" (-> obj value)) buf
            (+ (-> obj text-x) (* (-> obj x-sgn) (-> obj offset)))
            (/ (* (+ (-> obj text-y) (* (-> obj y-sgn) (-> obj offset)) (-> obj y-offset))
                  (the int (-> *video-parms* relative-y-scale))
                  )
              2
              )
            (font-color dark-yellow)
            (font-flags shadow kerning large))
    )
  0
  (none)
  )
```

dallmeyer:
- Cleaned up the decompilation artifacts as suggested in the functions I touched
- I've left the totals in place as I think they help make it clear which view you're seeing. Looking into `print-game-text-scaled` to improve the sizing
- Switched to toggles
  - dpad up for game totals / down for level totals / left or right for original view
  - it functions as one global toggle (so the 3 are always using the same view) that is persisted as long as any collectable (cell/orb/fly) hud part is visible. Resets back to original once all 3 of these hud parts are hidden
  - since the toggle resets when the HUD goes away, I didn't add any sound as I thought that could get annoying - I see these as a nice way to briefly check the totals without having to go into the pause menu and disrupt gameplay

goal_src/jak1/engine/ui/hud.gc
```diff
@@ -270,6 +308,10 @@
   :code (behavior ()
     (logior! (-> self mask) (process-mask sleep-code))
     (loop
+      ;; if all collectable hud parts are hidden, revert collectables view to original
+      (if (and (!= *hud-collectable-view* (hud-collectable-view original)) (hud-collectables-hidden?))
+        (set! *hud-collectable-view* (hud-collectable-view original))
+        )
```
ManDude, line 314: Add some sort of notice that you changed this from the original game.

The counters resetting when they go away feels like only a half-way implementation of this. It's probably way more intuitive and easier to mod if you just change the `hud-update` methods for each of the hud classes separately instead of doing it all in a state which *all* hud classes use. It also probably gives you more control.

ManDude (changes requested):
Sorry, I just really want this feature to be as good as possible so we don't have to make changes to it later down the line and keep built-in modifications to a minimum. :-)

goal_src/jak1/engine/ui/hud.gc
```diff
@@ -207,6 +221,30 @@
...
+      (dotimes (s5-0 9)
+        (when (-> *hud-parts* parts s5-0)
+          (when (not (hidden? (-> (the-as (pointer hud) (-> *hud-parts* parts s5-0)) 0)))
+              (case (-> (the-as (pointer hud) (-> *hud-parts* parts s5-0)) 0 name)
+                  (('hud-money 'hud-buzzers 'hud-fuel-cell)
+                    (set! gp-0 #f)
+                    )
+                )
+              )
+          )
```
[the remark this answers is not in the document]
dallmeyer: Yeah think you are right - I basically just followed the example from `hud-hidden?` (same reason I used `local-vars`)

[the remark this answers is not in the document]
ManDude: Yeah, that one's just gonna check pretty much all huds, and they probably had constants or macros that automatically used the correct amounts to check for.

goal_src/jak1/engine/ui/hud.gc
dallmeyer: Sure, I can move this into just the fuel/money/buzzer `hud-update` methods, that makes sense though it will be a bit of repeated code.

I can remove the reset on when the hud is hidden, that's easy - but I do want to leave the state global so that all 3 hud parts are in sync (i.e. avoid situations like cells showing game-totals while buzzer shows level-totals)

decompiler/config/all-types.gc
```diff
@@ -15630,6 +15636,7 @@
 ;; - Symbols
 
 (define-extern *hud-parts* hud-parts)
+(define-extern *hud-collectable-view* hud-collectable-view)
```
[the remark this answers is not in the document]
dallmeyer: Sorry I always get this wrong, I forget what I was doing but I caused `task type-test` to fail at some point by getting some type out of sync with this... anyways I'll revert

goal_src/jak1/engine/ui/hud-classes.gc
```diff
@@ -745,41 +745,39 @@
...
+      *temp-string*
+      s5-0
+      (+ (-> obj text-x) (* (-> obj x-sgn) (-> obj offset)))
+      (/ (* (+ (-> obj text-y) (* (-> obj y-sgn) (-> obj offset)) (-> obj y-offset))
+            (the int (-> *video-parms* relative-y-scale))
+            )
+          2
+          )
+      (font-color dark-yellow)
+      (font-flags shadow kerning large)
```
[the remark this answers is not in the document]
dallmeyer: Just curious, what cleanup do you have in mind? The logic for x/y positioning? Renaming variables such as `s5-0`?

I'm always just a bit hesitant to change too much and diverge from the original logic

goal_src/jak1/engine/ui/hud.gc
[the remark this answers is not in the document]
dallmeyer: In any case, don't even need this function if we remove the reset when HUD diappears

goal_src/jak1/engine/ui/hud-classes.gc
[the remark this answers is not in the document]
ManDude: Pretty much exactly what my other comment said - the `with-dma-buffer-add-bucket` macro wraps code in those DMA buffer/packet lets, and the `string-format` macro is not handled properly in the decompiler which is causing the function calls here to split (that's why the function name was awkwardly stored in a separately variable before), it calls `format` on a `(clear *temp-string*)` and then returns that, which is why you see that kind of code here. [`string-format` macro here](https://github.com/open-goal/jak-project/blob/master/goal_src/jak1/kernel/gstring.gc#L761)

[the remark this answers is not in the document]
dallmeyer: Ah sorry, I overlooked some of the differences in your example - will clean up

xTVaser:
add to whitelist

dallmeyer:
@ManDude I added a wrapper around `draw-string-xy` to make it easier to scale text, borrowed some logic from `print-game-text[-scaled]`, and renamed some variable over there too for readability. See updated video in the PR description for what it looks like now

goal_src/jak1/engine/gfx/font.gc
```diff
@@ -1830,3 +1830,49 @@
...
+        (+! (-> font-ctxt origin y) (the-as float (/ (the int (- (-> font-ctxt height) scaled-height)) 2)))
+        )
+      
+      ;; scaling globals *font-default-matrix*, *video-parms* (mimicking print-game-text scaling)
+      (*! (-> *font-default-matrix* vector 0 x) scale)
+      (*! (-> *font-default-matrix* vector 1 y) scale)
+      (*! (-> *video-parms* relative-x-scale) scale)
+      (*! (-> *video-parms* relative-y-scale) scale)
+      (set! (-> *video-parms* relative-x-scale-reciprical) (/ 1.0 (-> *video-parms* relative-x-scale)))
+      (set! (-> *video-parms* relative-y-scale-reciprical) (/ 1.0 scale))
```
dallmeyer, line 1864: based on the below, where `sv-136` is the scale https://github.com/open-goal/jak-project/blob/55ad7306205fd7d70313ea3c343079ecfe323444/goal_src/jak1/engine/ui/text.gc#L424-L429

goal_src/jak1/engine/gfx/font.gc
```diff
@@ -1830,3 +1830,49 @@
     (draw-string str buf context)
     )
   )
+
+(defun draw-string-xy-scaled ((str string) (buf dma-buffer) (x int) (y int) (color font-color) (flags font-flags) (scale float))
```
ManDude, line 1834: Write something that says this was added for the pc port.

I think the way you wrote this function is effectively a wrapper for `print-game-text-scaled`. I think you can just copy `draw-string-xy` but simply set a scale on the `font-context` before drawing instead of copying all of this unnecessary text padding stuff.

goal_src/jak1/engine/ui/text.gc
```diff
@@ -327,35 +327,33 @@
...
-    (set! (-> font-ctxt origin x) f24-0)
-    (set! (-> font-ctxt origin y) f22-0)
-    (set! (-> font-ctxt width) f26-0)
-    (set! (-> font-ctxt height) f30-0)
-    (set! (-> font-ctxt scale) f28-0)
+    (set! (-> font-ctxt origin x) orig-x)
+    (set! (-> font-ctxt origin y) orig-y)
+    (set! (-> font-ctxt width) orig-width)
+    (set! (-> font-ctxt height) orig-height)
+    (set! (-> font-ctxt scale) orig-scale)
```
ManDude, line 356: thanks for cleaning this up

ManDude (changes requested):

goal_src/jak1/engine/gfx/font.gc
dallmeyer: Will add a note about modified for PC 👍

So `print-game-text-scaled` calls `print-game-text` which ultimately calls `draw-string`, so this actually feels like a bit more of a direct route to a scaled call to `draw-string`.

I'll try removing the extra padding code and see how things look today 👍

dallmeyer: OK yep as you said, the adjustments/padding were unnecessary - looked exactly the same with and without (I did actually test with `(font-flags left)` and `(font-flags middle)`

goal_src/jak1/engine/gfx/font.gc
```diff
@@ -1830,3 +1830,40 @@
...
+      
+      ;; un-scaling globals *font-default-matrix*, *video-params*
+      (set! (-> *font-default-matrix* vector 0 x) orig-mat-x-scale)
+      (set! (-> *font-default-matrix* vector 1 y) orig-mat-y-scale)
+      (set! (-> *video-parms* relative-x-scale) orig-vid-parms-x-scale)
+      (set! (-> *video-parms* relative-y-scale) orig-vid-parms-y-scale)
+      (set! (-> *video-parms* relative-x-scale-reciprical) (/ 1.0 (-> *video-parms* relative-x-scale)))
+      (set! (-> *video-parms* relative-y-scale-reciprical) (/ 1.0 (-> *video-parms* relative-y-scale)))
+      )
+    )
```
ManDude, line 1867: Can you not just do
```lisp
  (let ((context (new 'stack 'font-context *font-default-matrix* x y 0.0 color flags)))
    (set-scale! context scale)
    (draw-string str buf context)
    )
```
?

dallmeyer: No, `draw-string` doesn't actually respect the `font-context` scale, as far as I could tell. `print-game-text-scaled` needed to take that scale and apply it to the matrix and video-parms, and then revert them back after.

dallmeyer: I could move this scaling logic into draw-string I suppose, but that function is a mess and I was reaaaaally reluctant to touch it

ManDude: Oh okay, it's fine then. Don't touch that function, it's a tangled mess of assembly.

ManDude (approved):

water111 (approved):

PR #1757, opened 2022-08-14
fix bug in knuth RNG

dallmeyer:
- I had a bug where I wasn't adding the min back
- `(sar (* bound (knuth-rand-next 31)) 31)` will always returns -1 or 0 when bound is a power of 2, should use `imul64` instead here

### DIFF
--- a/goal_src/jak1/pc/util/knuth-rand.gc
+++ b/goal_src/jak1/pc/util/knuth-rand.gc
@@ -59,11 +59,13 @@
          (bits 0)
          (val 0))
 
+    #|  For some reason this doesn't seem to be working correctly :(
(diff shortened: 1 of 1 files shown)


ManDude:
maybe you needed to make it `imul64` for the 64-bit multiplication? Also, you can actually do `(- var)` to just negate a value.

dallmeyer:
> maybe you needed to make it `imul64` for the 64-bit multiplication? Also, you can actually do `(- var)` to just negate a value.

d'oh you are almost certainly right, I thought I had fixed this everywhere but I guess not. Let me try that.

dallmeyer:
Yep that was the issue @ManDude, thanks for catching!

PR #1758, opened 2022-08-14
fix a game bug and add a new compiler feature

ManDude:
Adds support on the goal compiler for inline arrays of value types which work similarly to inline arrays of structure types, allowing you to write something like `(new 'global 'inline-array 'uint32 10)` which would have thrown an error.

Also fixes a bug with the gondola not working sometimes.

Also moved some of our older custom code to a better folder and did clean-up on it (and fixed some bugs with it...)


common/type_system/TypeSystem.cpp
```diff
@@ -296,26 +297,23 @@ DerefInfo TypeSystem::get_deref_info(const TypeSpec& ts) const {
 
   if (ts.base_type() == "inline-array") {
     auto result_type = lookup_type(ts.get_single_arg());
-    auto result_structure_type = dynamic_cast<StructureType*>(result_type);
-    if (!result_structure_type || result_structure_type->is_dynamic()) {
+    if (auto result_structure_type = dynamic_cast<StructureType*>(result_type);
+        result_structure_type && result_structure_type->is_dynamic()) {
```
water111, line 301: I'd kind of prefer to leave this as it is, I think it is extra confusing that there are two ways to do the same thing, and it's hard to know if this really handles all cases.

goalc/compiler/compilation/Type.cpp
```diff
@@ -1003,9 +1003,14 @@ Val* Compiler::compile_heap_new(const goos::Object& form,
...
-    array->set_type(ts);
-    return array;
+    auto new_array = compile_real_function_call(form, malloc_func, args, env);
+    if (info.is_struct) {
+      new_array->set_type(ts);
+    } else {
+      // allocs of value types are exactly the same as a pointer deref
+      new_array->set_type(TypeSpec("pointer", {ts.get_single_arg()}));
+    }
+    return new_array;
```
water111, line 1013: Another way to do this would be to modify the condition on line 985 from `is_inline` to
```
is_inline && m_ts.lookup_type(elt_type)->is_reference()
```
then I think you can remove the modifications to `get_deref_info`.

I think this would be nicer because it means that we don't add more cases to `get_deref_info`.

water111 (changes requested):
I had a suggestion on the type stuff, everything else looks good

water111 (approved):

PR #1760, opened 2022-08-15
i18n: init translation files that can be used by crowdin

xTVaser:
image

This should make it easier for people to submit / as well as track what needs to be translated for our custom strings.  We can't do everything because there is no base string in english to share (it's in the ISO).

Crowdin monitors and updates the `*.json` files (sync'd manually so we don't go insane with PRs)

I've added support, and transitioned text translations to a json format so there is a single source of truth for crowdin to use.  I changed my mind about having an intermediate format just to retain the GOAL format.  The JSON format does not yet support credits translation, but nothing is using it and it's not documented anywhere yet either.  This should probably done independently as it is radically different from simple key-value, and a translation service is not going to work with it well.


MineSAT:
Can you fix this synchronization failure with the repository so that I can start with the translation? :)

image

xTVaser:
Yes, after the PR is merged and it targets this repo instead of the temporary one.  I don't want anyone editing a proof of concept project as it will likely get reset in the coming days.

MineSAT:
> Sí, después de fusionar el PR y apunta a este repositorio en lugar del temporal. No quiero que nadie edite un proyecto de prueba de concepto, ya que es probable que se reinicie en los próximos días.

Perfect, entonces se puede traducir ya sin problema?

MineSAT:
> Yes, after the PR is merged and it targets this repo instead of the temporary one. I don't want anyone editing a proof of concept project as it will likely get reset in the coming days.

Perfect, let me know when you can start translating :)

MineSAT:
Hi, it's been a while since I've seen this, but is the page now available to send translations?

xTVaser:
Give me a week or so, I have to reach out to Crowdin to get an OSS license and I probably want to write up some documentation about what characters are supported.  I'll make an announcement in the discord/website when it's ready.

xTVaser:
fyi https://opengoal.dev/blog/announce-crowdin-text/

PR #1764, opened 2022-08-15
Add dockerfiles for Linux environments with documentation

liamh101:


water111 (approved):
Thanks! I am pretty happy with how simple these dockerfiles are.

PR #1769, opened 2022-08-19
scripts: Add a (hopefully) better goal_src updating script

xTVaser:
My latest python monstrosity

The highlevel overview of the script is as such:
- read in your current goal_src file, annotating every comment/comment-block/explicitly retained block with as much metadata as possible
- read in the decompiled file, ignoring superfluous lines and explicitly mentioned forms
- merge the files, while doing a _variety_ of things, many of which are best-effort heuristics to try to put the retained code back in the right spot
  - if all else fails, just use the line number

Why would I torture myself with this?
- While doing jak 1, there was hesitation to comment code as we went along.  Mostly because we could improve/fix the decompiler after the fact, creating a very painful manual task of having to put back all the comments in the right spot.
- This tries to remedy that, removing the fear of re-applying the comments means we can document as we go

I've been able to prove this out using jak 2's gkernel.gc but I'll very likely have to iterate as jak 2's decompilation effort progresses and bugs are revealed when it's being used for real.

I also removed more of the now irrelevant scripts from the jak 1 days.


scripts/update-goal-src.py
```diff
@@ -1,131 +0,0 @@
...
-  # write the damn thing
-  os.remove(file_path)
-  with open(file_path, "w") as f:
-    f.writelines(new_lines)
-
-  print("Copied - {}!".format(file))
-
-
-if throw_error:
-  exit(1)
```
water111, line 131: RIP `update-goal-src.py`

PR #1770, opened 2022-08-19
add a bunch of new cheat codes & re-do all pc cheat codes

ManDude:
Changes the pc cheat code encoding so that there's less inputs and adds a bunch of new cheats:
- big head mode for Jak & NPC's
- big fist mode for Jak
- small head mode for Jak
- HUGE head mode for Jak
- mirror mode
- no textures mode

Wrote a bunch of code in the new `pc-anim-util.gc` for easier joint manipulation.

Also cleaned up some random other things.

The no texture mode added some branches to the shaders, it should be checked if there was a performance dip from that.


water111:
Looks awesome!

> it should be checked if there was a performance dip from that.

I checked, it looks the same. My understanding is that if you have a branch like this that uses a uniform, the cost is basically nothing (it can recompile a version of the shader for each value of the uniform).

PR #1773, opened 2022-08-20
game: add a speedrunner mode which enables a faster new-game/resetting experience

### LINKED ISSUE #1670: Speedrun Mode/Cutscene skipping

I think having a "speedrun mode" option in the game would be great for letting some other things exist. For example, when speedrun mode is disabled, you can skip cutscenes. Thus allowing speedrunners to keep the run as close to original as possible, but letting normal users skip annoying cutscenes. I'm sure there's other things that could go into a "speedrun mode" as well.

Add an option to enable an exclusive mode that disables some features when enabled, but enables some things while disabled.

### Additional context

_No response_

xTVaser:
- skips intro cutscene
- removes auto-save prompt

https://www.youtube.com/watch?v=kyLDUQR_UbY

closes #1670 

Other ideas that im not 100% sure on yet (so didn't implement):
- constant fish rng seed
- an even faster way to reset the current save file (couldn't easily figure this out, very coupled with the progress screen right now it seems)
- skipping the "good training boys..." cutscene as well

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -36,12 +36,19 @@ tasks:
     cmds:
       - '{{.DECOMP_BIN_RELEASE_DIR}}/decompiler "./decompiler/config/{{.DECOMP_CONFIG}}" ./iso_data ./decompiler_out --config-override "{\\\"decompile_code\\\": false}"'
   boot-game:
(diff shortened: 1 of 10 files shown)


ManDude:
I was gomna do this next as well. Kinda beat me to it.

PR #1775, opened 2022-08-21
game/speedrunning: Add struct with relevant information to facilitate Auto Splitting

xTVaser:
Draft until this is confirmed to have everything needed to facilitate splitting current runs -- being tested by some people currently.

This adds a struct that an auto-splitter can find and watch to split/reset/start the timer on various things.  Right now, all `need-resolutions` are exposed, as well as stats on power-cells/scout-flies/orbs.

This is quite verbose mainly because:
- for LiveSplit, all settings have to be setup before you connect to the process, so I can't dynamically figure out these lists
- order of the struct has to be consistent as to not cause breaking changes.

This does not handle load removals yet.


game/features/auto_splitter.cpp
```diff
@@ -0,0 +1,247 @@
+#include "auto_splitter.h"
+
+#include "game/kernel/jak1/kmachine.h"
+#include <game/kernel/common/Ptr.h>
+#include <game/kernel/common/kmachine.h>
+#include <game/kernel/jak1/kscheme.h>
```
water111, line 6: use `"` here

game/features/auto_splitter.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+#include "third-party/fmt/core.h"
+
+AutoSplitterBlockJak1 gJak1AutoSplitterBlock;
+
+void update_autosplitter_block_jak1(u32 jak1_autosplit_info) {
+  auto info = jak1_autosplit_info ? Ptr<jak1::AutoSplitInfoJak1>(jak1_autosplit_info).c() : NULL;
```
water111, line 13: I think this check is redundant - doing `Ptr<Whatever>(0).c()` will give you `NULL`:

```
  /*!
   * Convert to a C pointer.
   */
  T* c() {
    if (!offset) {
      return nullptr;
    }
    return (T*)(g_ee_main_mem + offset);
  }
  ```

File: game/features/auto_splitter.h [hunk 3]
```diff
@@ -0,0 +1,138 @@
+#pragma once
+
+#include <common/common_types.h>
```
water111, line 3: `"`

game/kernel/jak1/kmachine.cpp
```diff
@@ -44,6 +44,7 @@
 #include "game/sce/sif_ee.h"
 #include "game/sce/stubs.h"
 #include "game/system/vm/vm.h"
+#include <game/features/auto_splitter.h>
```
water111, line 47: `"`

water111 (approved):
Looks good, just a few minor things.

One idea is that maybe this struct doesn't have to be in C at all? Maybe it can just be in GOAL?  You could make a struct with the exact same memory layout.

Issue #1765, opened 2022-08-16
[runtime] quick screenshot button


ManDude:
F8? F2? Any ideas?

RandomPrimary2004:
Steam screenshot button is f12, so I don't know if it would be better to use that or not considering that some people run the game as a non steam game. f8 makes sense to me as an in-game screenshot function, but then again such a thing may be entirely unneeded considering that most GUI operating systems come with or can be configured to have a screenshot feature.

Zedb0T:
F2/F3

xTVaser:
Whatever is picked can it be easily toggable until we have a proper UI to remap buttons.  For example I have my AMD driver recording hotkeys setup to involve function keys.

sardap:
Should we just save the screenshot as what is set in the debug gui ("screenshot.png") or do we want it to do a steam style name + timestamp if you trigger it with the hotkey so you don't override it each  screenshot?

ManDude:
The debug GUI will eventually go away completely for normal users, so the filename would have to be something automatic. I really don't want to program an entirely new UI for this (and nobody else probably does either) so I think that's what we're gonna stick with on that aspect: hotkey and it just saves a screenshot to some user folder.

sardap:
I think "opengoal_YYYY_MM_DD_hh_mm_ss.png" does the job this does limit to one screenshot per second though. 

For disabling screenshot with hotkey  is a simple checkbox in the screenshot debug GUI menu acceptable? 
![gk_Zka8MAlhS4](https://user-images.githubusercontent.com/22311832/185105227-e582ffe7-85ee-4044-b252-906c3fa768da.png)

sardap:
Thinking about this more it really should be mappable though the existing control mapping UI I will try to get it working though that.

xTVaser:
I would replace `opengoal` with the actual game name, ie. `jak1`.  That way we are future proofed for jak 2 and beyond, and opengoal is becoming a bit overloaded of a term.

PR #1776, opened 2022-08-21
Added screenshot hotkey

sardap:
Added screenshot hotkey and added a toggle to debug GUI to disable screenshot hotkey #1765

### DIFF
--- a/game/graphics/pipelines/opengl.cpp
+++ b/game/graphics/pipelines/opengl.cpp
@@ -403,8 +403,8 @@ void GLDisplay::on_iconify(GLFWwindow* /*window*/, int iconified) {
 
 namespace {
 std::string make_full_screenshot_output_file_path(const std::string& file_name) {
-  file_util::create_dir_if_needed(file_util::get_file_path({"gfx_dumps"}));
-  return file_util::get_file_path({"gfx_dumps", file_name});
+  file_util::create_dir_if_needed(file_util::get_file_path({"screenshots"}));
+  return file_util::get_file_path({"screenshots", file_name});
 }
 }  // namespace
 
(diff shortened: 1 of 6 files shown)


game/graphics/pipelines/opengl.cpp
```diff
@@ -435,7 +460,9 @@ void render_game_frame(int game_width,
     options.save_screenshot = false;
     options.gpu_sync = g_gfx_data->debug_gui.should_gl_finish();
     options.borderless_windows_hacks = windows_borderless_hack;
-    if (g_gfx_data->debug_gui.get_screenshot_flag()) {
+    want_hotkey_screenshot =
+        want_hotkey_screenshot && g_gfx_data->debug_gui.screenshot_hotkey_enabled;
+    if (g_gfx_data->debug_gui.get_screenshot_flag() || want_hotkey_screenshot) {
       options.save_screenshot = true;
       options.game_res_w = g_gfx_data->debug_gui.screenshot_width;
       options.game_res_h = g_gfx_data->debug_gui.screenshot_height;
```
ManDude, line 468: Don't use this code for quick screenshots, as they need to be set using imgui which is exactly what we want to avoid.

game/graphics/pipelines/opengl.cpp
```diff
@@ -389,12 +402,24 @@ void GLDisplay::on_iconify(GLFWwindow* /*window*/, int iconified) {
 }
 
 namespace {
-std::string make_output_file_name(const std::string& file_name) {
+std::string make_full_screenshot_output_file_path(const std::string& file_name) {
   file_util::create_dir_if_needed(file_util::get_file_path({"gfx_dumps"}));
   return file_util::get_file_path({"gfx_dumps", file_name});
```
ManDude, line 407: we can probably rename this folder to `screenshots` now.

ManDude (changes requested):

game/graphics/pipelines/opengl.cpp
sardap: Okay I have separated them

game/graphics/pipelines/opengl.cpp
sardap: Done

sardap:
@ManDude Do I resolve the conversation after addressing the comment? Or do I leave that to the person who left the comment? 

I have only used bitbucket for pull requests and it doesn't have a concept of resolving a conversation.

xTVaser:
Doesn't really matter, resolving the comment just collapses it to minimize vertical space for everyone.  It can be unresolved again if deemed that it wasn't actually resolved.

game/graphics/pipelines/opengl.cpp
```diff
@@ -389,12 +402,24 @@ void GLDisplay::on_iconify(GLFWwindow* /*window*/, int iconified) {
...
 }
 }  // namespace
 
+static std::string get_current_timestamp() {
+  auto current_time = std::time(nullptr);
+  auto local_current_time = *std::localtime(&current_time);
+  std::ostringstream oss;
+  oss << std::put_time(&local_current_time, "%Y_%m_%d_%H_%M_%S");
+  return oss.str();
+}
```
ManDude, line 417: Looking at `sceCdReadClock`, we do the following to get the time:

```cpp
  time_t t = time(0);
  std::tm* date = localtime(&t);
```

Stringstreams are kinda janky and we haven't used them anywhere, so it might be better to just format the string using that code.

sardap: I opted to use strftime instead.  fmt has date formatting support but we are missing "fmt/chrono.h" https://fmt.dev/latest/api.html

ManDude (approved):

sardap:
Fixed clang formatting error sorry for missing it have set up a pre hook to catch it next time

water111 (approved):
Thanks!

PR #1778, opened 2022-08-21
game: windowed mode related fixes

xTVaser:
- restore the windowed mode size properly from `pc-settings.gc`
- when switching to windowed mode from fullscreen, update the size properly
- don't auto-center the window on startup (this was due to a side-effect where it would always adopt the full-screen settings when the fullscreen update was forced)
- save and restore the windowed mode position

I've tested that this seems backwards compatible / doesn't effect full-screen modes but...should be somewhat tested before merging because this stuff has so many edge-cases historically.

### DIFF
--- a/common/util/FileUtil.cpp
+++ b/common/util/FileUtil.cpp
@@ -33,8 +33,8 @@
 #include <cstring>
 #include <unistd.h>
 #endif
+#include "common/log/log.h"
 #include "common/util/Assert.h"
-#include <common/log/log.h>
 
 namespace file_util {
 fs::path get_user_home_dir() {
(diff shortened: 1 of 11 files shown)


water111:
If I change to windowed mode while already in windowed mode, the resolution of the window changes. Not sure if this is an intended behavior or not, or possibly some linux-specific weird thing (most of this stuff doesn't work quite right on linux)

RandomPrimary2004:
that sort of behavior is probably dependent on your window manager, xserver, and other minor variables of that nature. Window manager behavior is sketchy on Linux as it is.

PR #1786, opened 2022-08-22
game: fix transition from fullscreen (not borderless) to windowed

xTVaser:
There seems to be no reason to update the values here, they are updated by callbacks when moving the window.

There might be value in adding a check to reset to defaults if the monitor changes and it's off-screen, but that's a bit out of scope for this fix (original code only did this by accident).

When restoring to windowed mode, actually use the last position instead of the center of the monitor.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -98,16 +98,6 @@ tasks:
   offline-tests:
     cmds:
       - '{{.OFFLINETEST_BIN_RELEASE_DIR}}/offline-test --iso_data_path "./iso_data/{{.GAME}}" --game {{.GAME}}'
-  # TODO - update or replace
-  # add-reference-test:
-  #   cmds:
-  #     - task: decomp-file
-  #     - python ./scripts/add-reference-test.py --file "{{.FILES}}"
-  #     - task: offline-tests
-  # add-reference-test-no-decomp:
-  #   cmds:
-  #     - python ./scripts/add-reference-test.py --file "{{.FILES}}"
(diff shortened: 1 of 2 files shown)


Francessco121:
With this change if I:
1. Start in windowed
2. Go fullscreen
3. Switch to another display while still in fullscreen
4. Go back to windowed

then the game window will end up back on the first monitor I originally had it windowed on, which in my opinion is a bit strange. Is this intentional? The code was originally centering the window to avoid this. Not a big deal, just curious.

xTVaser:
Not really intentional, I hadn't considered that scenario.  Trying to fix it I'm running into a few rough edges -- some I think I can fix, others I'm not so sure if I can reliably / will end up doing in this PR.
1. The reverse is awkward as well -- if I drag the window to another monitor it doesn't update the display.  So going from windowed -> fullscreen -> windowed puts it on the wrong monitor
2. The calculation for resolving the center of the screen fails if your game resolution doesn't match the native resolution of your monitor.  ie. 1280x720 game on a 4k monitor.  The width/height seems to be equal to the game's scaled resolution and not the true monitor's.  I tried another API function `getVidMode` but it seems to produce similar results
3. Because of this, I figure I can just default to top-left.  The problem with that is....I have stacked monitors and the X,Y returned by GLFW is actually the _bottom left_ for my top monitor, and the _top left_ for my normal monitor.  This I can probably workaround with some simple conditions, but maybe im using GLFW wrong?

PR #1787, opened 2022-08-22
[decomp] Decompile first batch of files in engine

water111:


goal_src/jak1/engine/math/quaternion.gc
```diff
@@ -281,7 +281,8 @@
     (init-vf0-vector)
     (.lvf vf1 (&-> arg1 vec quad))
     (.mul.vf vf2 vf1 vf1)
-    (.sub.vf vf3 vf3 vf3)
+    ;; (.sub.vf vf3 vf3 vf3)
+    (.xor.vf vf3 vf3 vf3)
```
water111, line 285: this `.sub` in jak 1 has the potential to cause problems and I'm surprised it didn't. Subtracting two garbage floats doesn't result in 0 if you have NaNs. The `xor` trick is better because it always gives you 0.

PR #1803, opened 2022-08-26
d/jak2: finish `settings`

xTVaser:


goal_src/jak2/engine/sound/gsound.gc
```diff
@@ -550,9 +550,9 @@
   )
 
 (defun sound-volume-off ()
-  (set-setting! 'music-volume 'abs 0 0)
-  (set-setting! 'sfx-volume 'abs 0 0)
-  (set-setting! 'ambient-volume 'abs 0 0)
+  ;; (set-setting! 'music-volume 'abs 0 0)
```
water111, line 553: why comment these out?

water111 (commented):
Looks good, I had one question, and then the merge conflicts.

water111 (approved):
:+1: Just merge conflicts

goal_src/jak2/engine/sound/gsound.gc
xTVaser: no idea why i did this

xTVaser: ah, has to do with the confusion around how we have a `set-setting!` macro instead of just using the method, and how there's a build in form analysis set that looks for `set-setting`.

ManDude: the macro hides away the process register and global that are used 100% of the times.

PR #1808, opened 2022-08-27
removed copied vmodes and vmode_idx argument from functions

sardap:
I know there hasn't been a lot of discussion on this issue if this is not the route you think we should go I have another branch where I just changed the array into a vector. So no worries


ManDude:
I don't think there's any point to outright removing this code. It might eventually be useful for fullscreen/display mode stuff, which keeps having issues and we keep having to make changes for it.

PR #1811, opened 2022-08-27
Decompiled enemy-h, idle-control, and nav-enemy-h

animalstyletaco:
I wasn't fully able to decompile enemy-h and idle-control. I got a vast majority of these file done but I'm having trouble trying to determine what arguments should be pass in for idle-control method 10 and enemy-focus method 12.


decompiler/config/jak2/all-types.gc
```diff
@@ -29661,11 +29664,11 @@
   :size-assert         #x10
   :flag-assert         #xb00000010
   (:methods
-    (idle-control-method-9 () none 9)
-    (idle-control-method-10 () none 10)
+    (idle-control-method-9 (_type_ idle-control-frame idle-control-frame) idle-control 9)
+    (idle-control-method-10 (_type_ uint32 uint32) none 10)
```
ManDude, line 29668: use register semantics types such as `uint` (and `int`) instead of memory semantics types.

ManDude (changes requested):

decompiler/config/jak2/all-types.gc
```diff
@@ -29690,21 +29693,36 @@
 ;; enemy-h                        ;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-#|
-(deftype enemy-focus (focus)
+;;Added hopefully temp structure since "focus" used the 12 method slot instead of enemy-focus
```
water111, line 29696: It should be possible to make this work without needing a separate type like this.

The type of method 12 in focus and enemy-focus should be the same (though it's possible that I got it wrong in `focus` and you'll need to add an additional argument).

I think it's probably going to be something like:
```
(_type_ process-focusable enemy)
```

decompiler/config/jak2/all-types.gc
```diff
@@ -29690,21 +29693,36 @@
...
   ((aware uint64  :offset-assert 16)
    (flags uint64  :offset-assert 24)
    )
   :method-count-assert 14
   :size-assert         #x20
   :flag-assert         #xe00000020
   (:methods
-    (enemy-focus-method-13 () none 13)
+    (enemy-focus-method-12 (_type_ uint32 enemy) symbol 12)
+    (enemy-focus-method-13 (_type_ uint32 uint32) symbol 13)
```
water111, line 29722: For arguments and return values, we should use `int` and `uint`, and not any of ones with sizes. The integer types with sizes should be used in arrays and fields of types (any place where they are stored in memory) and the ones without sizes should be used for any variable/argument/return value (any places where they are stored in registers)

goal_src/jak2/engine/ai/enemy-h.gc
```diff
@@ -5,5 +5,759 @@
...
+        (format #t "lock-focus ")
+        )
+    (if (= (logand (enemy-flag recover) s5-0) (shl 8 32))
+        (format #t "death-start ")
+        )
+    (if (= (logand s5-0 (enemy-flag jump-check-blocked)) (enemy-flag jump-check-blocked))
+        (format #t "enable-on-active ")
+        )
+    (if (= (logand (enemy-flag directed-ready) s5-0) (enemy-flag directed-ready))
+        (format #t "checking-water ")
```
water111, line 549: hm, it looks like these flag values aren't quite right - I'd expect to see the print statements match the flag they are testing (so this one should be `directed-ready`)

decompiler/config/jak2/all-types.gc
```diff
@@ -47055,9 +47148,40 @@
...
+  (attached 2)
+  (no-slowing-for-turns 3)
+  (recovery-mode 4)
+  (blocking-dest-node 5)
+  (direct-mode 6)
+  (ignore-others 7)
+  (debug 8)
+  (on-straightaway 9)
+  (draw-marks 10)
+  (left-turn 11)
```
water111, line 47165: I think this order isn't right - unfortunately the order they are printed in doesn't match the order, so you have to look at the assembly and figure out which bit it's testing.

For example:
```

    lwu s5, 0(gp)             ;; [ 16] (set! s5 (l.wu gp))
    addiu v1, r0, 16          ;; [ 17] (set! v1 16)
    andi a0, s5, 16           ;; [ 18] (set! a0 (logand s5 16))
    bne a0, v1, L53           ;; [ 19] (b! (!= a0 v1) L53 (set! v1 #f))
    or v1, s7, r0

B4:
    lw t9, format(s7)         ;; [ 20] (set! t9 format)
    daddiu a0, s7, #t         ;; [ 21] (set! a0 #t)
    daddiu a1, fp, L225       ;; [ 22] (set! a1 L225) "do-turn "
    jalr ra, t9               ;; [ 23] (call!)
    sll v0, ra, 0

    or v1, v0, r0             ;; [ 24] (set! v1 v0)
 ```
is `and`ing with 16 = `0b10000`, so `do-turn` would be 4.

File: decompiler/config/jak2/all-types.gc [hunk 6]
```diff
@@ -47091,8 +47214,124 @@
 |#
 
 #|
-(deftype vehicle (rigid-body-object)
-  ()
+(defenum vehicle-flag
+    :bitfield #t
+    :type int64
+    (lights-update 0)
```
water111, line 47220: same here

decompiler/config/jak2/all-types.gc
```diff
@@ -47198,13 +47437,60 @@
 ;; citizen-h                      ;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-#|
+;; kernel types
+(defenum citizen-flag
+    :bitfield #t
+    :type uint16
+    (female 0)
```
water111, line 47444: and here

water111 (changes requested):

xTVaser:
ok to test

PR #1816, opened 2022-08-29
[decomp2] `joint-mod`, `chain-physics`, `loader`

### LINKED ISSUE #1824: [decompiler] jak 2 `process->handle` support
```lisp
      (let ((v1-61 (process->ppointer arg0)))
        (set! (-> sv-16 handle) (the-as handle (logior (if v1-61
                                                           (shl (-> v1-61 0 pid) 32)
                                                           0
                                                           )
                                                       (.asm.sllv.r0 v1-61)
                                                       )
                                        )
              )
        )
```

ManDude:
Fixes #1824 

There was one method in `loader` that I couldn't figure out. The CFG is broken but it doesn't look like an assembly function.


goal_src/jak2/engine/load/loader.gc
```diff
@@ -25,5 +93,2377 @@
...
+                             (27 . wait)
+                             (26 . wait)
+                             (28 . wait)
+                             (31 . wait)
+                             (35 . wait)
+                             )
+          )
+    gp-0
+    )
+  )
```
water111, line 2455: what the heck is this

water111 (approved):
looks great, thanks for fixing the ja and handle macros!
I am kinda confused about what `gui-control` does.

goal_src/jak2/engine/load/loader.gc
ManDude: The numbers are actually `gui-channel`s. I alMost want to write a hack for thos function so that it prints those instead.

PR #1817, opened 2022-08-30
decomp: `board-h`, `board-util`, `gun-util`, `darkjak-h`

Hat-Kid:
- `board-util` seems good
- `gun-util` has a few problems:
  - I couldn't figure out `draw-beam` and `(method 9 gun-info)`
  - Some functions used static lists of pairs which was causing errors, fixed by removing a check in `data_decompile.cpp` (`L225` as an example for such a list)
- ~~`joint.gc` will fail the ref tests because `root-channel` in `joint-control` was changed to `(inline-array joint-control-channel)` from `uint32` and this change was needed for a lot of stuff in `gun-util` and `board-util` as these deal with jetboard and gun animations a lot~~


goal_src/jak2/engine/target/gun/gun-util.gc
```diff
@@ -5,5 +5,1180 @@
...
+             (vector<-cspace! s3-0 (-> (the-as gun self) mag gp-1 joint))
+             (vector-normalize-copy!
+               s4-1
+               (the-as vector (&-> (-> (the-as gun self) mag gp-1 joint bone) transform data 4))
+               40960.0
+               )
+             (set! (-> s4-1 y) (+ 81920.0 (-> s4-1 y)))
+             (set! (-> s5-2 data 0) (the-as float (-> (the-as gun self) entity)))
+             (set! (-> s5-2 data 1) 1.0)
+             (s.d! (+ s5-2 16) 0)
```
water111, line 309: hm, I don't think this will actually compile successfully. Maybe for now we should leave out this file - I think there's a bunch of stuff in here that's not handled well yet.  But just having the types filled out in `all-types.gc` will still be helpful for when we finally get to this file

Hat-Kid: Emptied the contents of `gun-util.gc` for now

water111 (approved):

PR #1826, opened 2022-08-30
add project path option to the compiler

trippjoe:


xTVaser:
add to whitelist

trippjoe:
Realized this isn't useful because we can just set the current working directory for goalc.

trippjoe:
Turns out I was wrong to close this.

goalc/main.cpp
```diff
@@ -45,15 +46,24 @@ int main(int argc, char** argv) {
...
                "Attempt to automatically deduce the user, overrides '-user'");
   app.add_option("-g,--game", game, "The game name: 'jak1' or 'jak2'");
+  app.add_option("--proj-path", project_path_override,
+                 "Specify the location of the 'data/' folder");
   app.validate_positionals();
   CLI11_PARSE(app, argc, argv);
 
-  if (!file_util::setup_project_path(std::nullopt)) {
-    return 1;
-  }
```
xTVaser, line 53: you can't get rid of this completely, this needs to still run if a project path is not provided, change it to an `else if` and add it after the condition you added below.  With that change, seems to work as expected
image

xTVaser (changes requested):
Tested with the suggested change made, works fine.  I was able to:
- open the REPL within the project folder with no args and build the game
- run the game
- move `goalc` to an entirely different folder on an entirely different drive, and have it work with the `--proj-path`:
image
image

xTVaser (approved):
Did the same test as above, didn't have to make any code modifications this time -- looks good to me.

water111 (approved):
Thanks!

PR #1830, opened 2022-08-31
jak1/speedruns: Some final touches for speedrunning in jak 1

xTVaser:
- Display the release/revision until you get your first cell, if you are using speedrunning mode
  - This is to make it obvious what version people are using when verifying a run
- Enforce 60fps while speedrunning, as the other options are experimental
- Add an explicit setting (which defaults to true) for skipping cutscenes
  - Some people _may_ want to run without the ability to skip cutscenes, but I feel like this also has casual use (people that don't want to accidentally skip the cutscene as there is no way to replay them in jak 1).
- Ensure `force all actors` is disabled when starting a run as well.


ManDude:
> Ensure force all actors is disabled when starting a run as well.

just wondering, why?

xTVaser:
There are suspicions that enabling the actors early causes them to react to jak earlier than normal.  No idea if this is wrong or not, but from what I could tell it's not even an exposed option in the UI (only via debug menu) and it's safer to just disable it than worry about it.

ManDude:
and those issues were actually unrelated - it simply made them easier to trigger: https://github.com/open-goal/jak-project/pull/1758/files#diff-1d79954e17ea44f741d4d4e64c5f288e97ec02fdb35a1cfc28ff274784633f06R155
which I think is a good thing, personally.

And yes, enabling all actors does cause them to react earlier. After all, everything is spawned at once regardless of the visibility status (which was already not entirely accurate to the PS2 due to various reasons). I don't know why we should be the ones arbitrating whether it should be enabled or not, though, is the thing.

A menu option can be added (there's even text for it already). The only reason I didn't do it before was due to concerns over the game lagging with it on, which seem have largely gone away by now due to the faster renderers.

xTVaser:
> I don't know why we should be the ones arbitrating whether it should be enabled or not, though, is the thing.

We aren't really arbitrating anything, rather I am implementing what was discussed and decided upon, which imo isn't invasive to our code-base as it is isolated inside a speedrunning specific function.  Even if there was a menu option, it would still make sense to force the value to be `false` otherwise from a verifying standpoint -- everyone has to show their settings at one point in their recording which is just a pain for everyone involved.

I'm not going to personally add a menu option for it, as it's unrelated to the goal of this PR and I don't want to add anything beyond what is required that could potentially cause more support issues as we are trying to focus our efforts away from jak 1.

ManDude:
I agree that it makes sense to maintain some consistency with the settings, but I wasn't aware there was any discussion on this at all, so it felt to me like you had decided to keep it disabled for your own reasons. I don't really have a problem with it being invasive or not here.

I think it is a good option to have since it pretty much eliminates pop-in in levels and lets you do nice full level shots (all my promo screencaps have had it on for this reason). But we can tackle that later.

goal_src/jak1/pc/features/speedruns-h.gc
```diff
@@ -0,0 +1 @@
+(define-extern speedrun-start-run (function none))
```
ManDude, line 1: missing goal header

```lisp
;;-*-Lisp-*-
(in-package goal)
```

goal_src/jak1/pc/features/speedruns.gc
```diff
@@ -0,0 +1,16 @@
+(defun speedrun-start-run ()
```
ManDude, line 1: missing goal header
```lisp
;;-*-Lisp-*-
(in-package goal)
```

xTVaser:
Yeah I'm not doing any of this in isolation, as I mentioned a few days ago -- im part of the discussion (actually will be a moderator as well) for the new leaderboards.  These 3/4 things were the only things we identified as must-haves -- which im happy about since none of these felt too onerous (and as i mentioned one of which is probably nice for casual users as well)

goal_src/jak1/pc/pckernel.gc
```diff
@@ -98,6 +98,9 @@
 
 (defmethod set-frame-rate! pc-settings ((obj pc-settings) (rate int))
   "set the target framerate."
+  ;; ignore changes to frame-rate while in speedrunner mode
+  (when (= (-> obj speedrunner-mode?) #t)
+        (set! rate 60))
```
ManDude, line 103: possibly redundant? You already set it to 60 when starting a run. Since you need to go to the menu to change fps anyway, rejecting the fps change unless you go to a *different* menu first seems a bit awkward.

ManDude (changes requested):
there's some headers missing from the source code files (I think we should have those on all our files at least).

goal_src/jak1/pc/pckernel.gc
xTVaser: Possibly, i figured I would do this because this was the only setting exposed in the UI. But you are right, it would be visible in the recording so i can probably keep it simple

Zedb0T: Would it be better to just freeze this value in the function in autosplit.gc that keeps track of orbs/flies etc constantly?

It seems odd to allow a user to change a setting that would invalidate their run if we can just not?

This was it would all be separate from main code and only run when speedrun mode is on.

ManDude: There are many things that a user can do to invalidate their run in the menu, such as loading a save file.

xTVaser: I think the nice (from both a user and dev's perspective) approach will be adding support to the progress menu code to allow disabling / hiding options -- which is a pretty basic GUI programming feature that would likely be nice to have regardless in the long-term.

Freezing the value -- aka writing it constantly -- has perf implications (writing to the settings file every frame), but the current approach of ignoring the value is pretty hacky as well.

I think for now it's for the best to keep it simple and be a part of verification until we solve it in a better way -- as dass says, there are plenty of ways someone can invalidate the run and a verifier should atleast scrub through to catch obvious things like this.

Zedb0T: Yeah it's hacky not ideal and confusing, but doesn't the settings file only update when you close the progress menu or autosave?

xTVaser: Menu option is now nicely disabled when speedrunner mode is active, this feature could be taken advantage of for any menu option as well.

ManDude (changes requested):
Just some stuff with the lambda return values checking that I think will work better.

goal_src/jak1/pc/progress-pc.gc
```diff
@@ -1777,6 +1769,10 @@
               (option-y y-off)
               (locked? #f)
               )
+          ;; Check if the option should be disabled
+          (when (and (!= (-> options index option-disabled-func) 0)
+                     (= ((-> options index option-disabled-func)) #t))
```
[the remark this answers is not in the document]
xTVaser: I'm not clear on what you're suggesting.

Are you saying I can just do:
```lisp
(when (and (!= (-> options index option-disabled-func) 0)
               ((-> options index option-disabled-func)))
```
instead?

The idea was for the lambda added to the option to explicitly return a simple `true` or `false`, so that's why I check against `#t` explicitly.

goal_src/jak1/pc/progress-pc.gc
```diff
@@ -1660,18 +1660,10 @@
               (set! (-> *game-info* mode) 'play)
               ;; Start a new game differently if speedrunning mode is active
               (cond ((= (-> *pc-settings* speedrunner-mode?) #t)
```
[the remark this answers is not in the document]
xTVaser: For the same reason as above, I was under the impression that I had to explicitly check because OpenGOAL doesn't really have a true boolean -- am i mistaken and i can just do `(-> *pc-settings* speedrunner-mode?)`?

goal_src/jak1/pc/progress-pc.gc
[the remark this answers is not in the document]
ManDude: Yes, that is what I'm suggesting.

There's no benefit to checking `#t` directly for when you want a true or false response, and it's also non-standard. Another example where it would break unexpectedly is with using `or` and/or `and` (lol):
```lisp
(or *target* *sidekick*)
(and *target* *sidekick*)
```
`or` will never return `#t` here. `or` returns the value of the first truthy operand, or `#f` if all are not truthy.
`and` will also never return `#t`. similar to `or`, it returns the value of the last operand, or `#f` if any operand is not truthy.

see https://opengoal.dev/docs/reference/syntax/#truth

ManDude (approved):

water111 (approved):
:+1:
I like this approach for type consistency

PR #1837, opened 2022-09-02
Basic hash table and jump table

Brent-Hickey:
I wanted to try to build a jump table through macros since the goal compiler doesn't support them. I figured the hash functions would be useful for other things as well.

This ports the crc32 function from c++ just because and adds the murmur3 hash function for better performance. It supports a really really basic hash table and then adds support for integer jump tables and symbol jump tables.

In my testing the symbol jump table doesn't start to outperform a normal (case) until ~20 cases so it might not be helpful anywhere for now


xTVaser:
ok to test

xTVaser (approved):
Looks good to me, if for nothing else then it's always nice to have examples of OG code that aren't related to the game.  I moved it to the `pc/utils` folder which feels more appropriate.

PR #1839, opened 2022-09-03
Fix crash in light jak cutscene skip

### LINKED ISSUE #1836: [Cutscene skips] Skipping Light Jak cutscene too quickly crashes
### Acknowledgements

### Describe the Bug

For the cutscene right after you pick up Light Eco, if you press triangle to skip the cutscene too quickly after it starts, gk.exe crashes. I was only able to trigger this in retail mode, in debug mode it skips without issue. Also if you wait a few seconds before pressing triangle, then you can skip without crashing. 

### How To Reproduce

In retail mode, collect light eco at end of final boss fight and press triangle to skip the cutscene quickly after it starts, and the game should crash.

Yes, it's unique to OpenGOAL

### Expected Behavior

Cutscene should skip w/o crashing

### Operating System

Windows 11

### OpenGOAL Version

909bde8

### Game Version

NTSC 1.0 (black label)

Yes

dallmeyer:
Fixes #1836

From some testing, it seems the crash happens on the below line, because `(-> self tasks current-stage)` is not set to something valid (`current-stage` is accessed within `current-status`), so then `task-status->string` crashes when trying to convert the enum to string.

https://github.com/open-goal/jak-project/blob/2a1dc7bd2b97c5f5244d29f230c1ff694a38742c/goal_src/jak1/levels/finalboss/sage-finalboss.gc#L607


ManDude:
just remove that line, it's not necessary

xTVaser (approved):
tested, seems to work as described -- just removing a log

PR #1840, opened 2022-09-03
d/jak2: finish the majority of `sparticle` and `sparticle-launcher`

xTVaser:
The vast majority of the sparticle code remains the same (i wonder why lol) but there are a handful of new functions and effects in `sparticle-launcher` -- many of which ive left unfinished because they require essentially guesses at this point (they access arbitrary data).  The rest of the functions we used mip2c for in jak 1, which I don't know how to use yet.

In some cases in `sparticle` it seems like we manually fixed the functions -- like `memcpy`?  Perhaps these should also be mips2c'd this time around?


decompiler/config/jak1_ntsc_black_label.jsonc
```diff
@@ -18,7 +18,7 @@
   "disassemble_code": false,
 
   // Run the decompiler
-  "decompile_code": false,
+  "decompile_code": true,
```
ManDude, line 21: set this to false

goal_src/jak2/engine/gfx/sprite/particles/sparticle-h.gc
```diff
@@ -15,6 +16,7 @@
   (bit2 2)     ;; cleared after an aux has its func set to add-to-sprite-aux-lst
   (bit3 3)
   (bit4 4)
+  (unk5 5)
```
ManDude, line 19: unique name here

goal_src/jak2/engine/gfx/sprite/particles/sparticle-h.gc
```diff
@@ -30,14 +32,21 @@
   (left-multiply-quat 18)
   (right-multiply-quat 19)
   (set-conerot 20)
+  (unk-21 21)
```
ManDude, line 35: and here

test/decompiler/reference/jak2/engine/anim/joint_REF.gc
```diff
@@ -183,8 +183,10 @@
   )
 
 ;; definition for method 11 of type joint-control
+;; ERROR: Function may read a register that is not set: a2
 (defmethod debug-print-channels joint-control ((obj joint-control) (arg0 symbol))
-  (format arg0 "~0K")
+  (local-vars (a2-0 none))
+  (format arg0 "~0K" a2-0)
```
ManDude, line 189: looks like you broke something here

decompiler/util/DecompilerTypeSystem.cpp
```diff
@@ -422,7 +422,7 @@ int DecompilerTypeSystem::get_format_arg_count(const std::string& str) const {
 
   static const std::vector<char> single_char_ignore_list = {'%', 'T'};
   static const std::vector<std::string> multi_char_ignore_list = {
-      "0L", "1L", "3L", "1k", "1K", "2j", "0k", "0K", "30L", "1T", "2T"};
+      "0L", "1L", "3L", "1k", "1K", "2j", "0k", "30L", "1T", "2T", "100", "200", "350"};
```
ManDude, line 425: You removed `0K` from here.

Also what are the numbers for?

ManDude (changes requested):

goal_src/jak2/engine/gfx/sprite/particles/sparticle-h.gc
xTVaser: it is unique in the context of this enum? Do you want something globally unique?

decompiler/util/DecompilerTypeSystem.cpp
xTVaser: Mis-identified strings for format strings I encountered -- instead of adding them to hacks I added them here.  But didn't add the entire token because the code i wrote here more than a year ago is dumb and only supports 2/3 char multi-char strings.

goal_src/jak2/engine/gfx/sprite/particles/sparticle-h.gc
ManDude: globally yeah. It makes it way easier to search for usages later when we want to give them normal names.

decompiler/util/DecompilerTypeSystem.cpp
ManDude: you should probably add the specific strings to `bad_format_strings` instead then? so it's less confusing.

xTVaser: I intend to this this part of code in the coming days / if not before this PR merges

ManDude (approved):

water111 (approved):
Looks good, we're going to have to do a bunch of scratchpad/PC modifications/mips2c stuff.

I think we should wait until textures are working, so you can actually run this stuff and debug things. Let me know if you ware interested in working on this, I think it would be a good way to learn about mips2c stuff if you want.

PR #1846, opened 2022-09-05
[decomp] jak2: debug, debug-sphere

Francessco121:
I wasn't able to 100% complete `debug` due to a bunch of level boundary debug stuff I couldn't figure out. But, I added a ref test and documented/copied over everything else into goal_src.

Most of the functions that existed in Jak 1 are identical, others got an extra param or 2 and some now make copies of arguments instead of modifying them. There's a bunch of new functions, including all of the debug functions that used to be in `level-boundary` are now in `debug`. At the very end of `debug` there's also some weird asm functions checking for some EE memory controller bug (not sure what's up with that)?


water111 (approved):
Changes look good, there's just some merge conflicts (sorry, this was probably the PR I just merged...)

The EE memory controller bug they mention is a bug on the development TOOL.  In rare cases, there would be some type of data corruption/DMA issue if you transferred data from the upper 96 MB (or was it 64 MB?) of memory. There was a workaround provided by sony that reconfigured some hardware, but reduced the performance, and I think this function times a DMA transfer to see if the bug fix is enabled or not.

Francessco121:
Oh that's interesting... and also sounds painful lol.

The merge should be good now, but I might add a little to this PR tomorrow if it doesn't get merged before then. There's two functions I left marked as asm (`internal-draw-debug-line`, `add-debug-flat-triangle`) which aren't actually asm but they give errors like `ERROR: Unsupported inline assembly instruction kind - [srl v1, s5, 24]` even tho they decompile fine. I thought the misc fixes branch that just got merged would fix this but I guess I misunderstood. Either way, I want to take a second look at these soon.

water111:
looks like `(define-extern draw-string (function string dma-buffer font-context float))` in `debug.gc` needs to be updated to pass the type consistency check.

You can run just this check with a filter like `--gtest_filter="*Jak2*Type*"`.

I'll have to look more closely at the srl thing - I think it should end up accessing a field of an `rgba` and the warning should go away, but it sounds like that's not working.

water111:
:+1:

PR #1848, opened 2022-09-06
[speedrun] Show speedrun information at start of run

Zedb0T:
This shows the current settings needed to help provide information when verifying a run, also it removes the two calls to show the version in main.gc and instead does it within the speedrun/auto splitting scope.

- Shows Version
- Shows if Speedrunner mode is on (technically pointless since the text only displays when it is enabled this more just serves to show that there IS a speedrunner mode to viewers and stuff (josh) So hopefully it may cut down on submissions with the mode off)
- Shows if cutscene skips is enabled/disabled.

Tested on 16x9 4x3 the really weird one and borderless/fullscreen

[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=qDC-beEg5Es)

### DIFF
--- a/goal_src/jak1/pc/features/speedruns-h.gc
+++ b/goal_src/jak1/pc/features/speedruns-h.gc
@@ -2,3 +2,22 @@
 (in-package goal)
 
 (define-extern speedrun-start-run (function none))
+
+
+(defun draw-speedrun-settings ()
+  "debug draw"
+(if (and (-> *pc-settings* speedrunner-mode?)(< (-> *autosplit-info-jak1* num-power-cells) 1))
+(begin
+(format *pc-temp-string* "OpenGOAL Version: ~S ~%Speedrun mode: ~A ~%Cutscene Skips ~A" *pc-settings-built-sha* (-> *pc-settings* speedrunner-mode?) (-> *pc-settings* cutscene-skips?))
+  (with-dma-buffer-add-bucket ((buf (-> (current-frame) global-buf))
+                                     (bucket-id debug-no-zbuf))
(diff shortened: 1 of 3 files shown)


goal_src/jak1/pc/features/speedruns-h.gc
```diff
@@ -2,3 +2,22 @@
...
+                                     (bucket-id debug-no-zbuf))
+          (draw-string-xy *pc-temp-string*
+                          buf
+                          0
+                          (* 200 (-> *video-parms* relative-y-scale))
+                          (font-color flat-yellow)
+                          (font-flags shadow kerning)))
+                          (clear *pc-temp-string*)))
+                          (none)
+                          )
```
ManDude, line 22: Needs indent. Also, please use `when` instead of... whatever unreadable mess is going on here. I'm not sure what the `relative-y-scale` is for either since it doesn't have much to do with the framebuffer coords (but I could be wrong).

You can also use `string-format` and put that format call as one of the arguments, instead of putting it outside and then re-using the string global.

ManDude (changes requested):

goal_src/jak1/pc/features/speedruns-h.gc
Zedb0T: `relative-y-scale` seems to make it place in the same relative place regardless of the height of the screen?? I am not sure I just borrowed it from https://github.com/open-goal/jak-project/blob/master/goal_src/jak1/pc/pckernel.gc#L1162-L1170

Ill get the rest cleaned up now thanks!

ManDude (approved):

PR #1858, opened 2022-09-07
[decompiler] rewrite set lets as just sets

### LINKED ISSUE #1854: [decompiler] `(set! foo bar) bar` could be cleaner
In particular, we see stuff like
```
      (let ((v1-2 (the-as object (-> v1-1 mood-context state)))
            (f0-0 arg0)
            )
        (set! (-> (the-as forresca-states v1-2) electricity arg1 scale) f0-0)
        f0-0
        )
```
which could be
```
      (let ((v1-2 (the-as object (-> v1-1 mood-context state))))
        (set! (-> (the-as forresca-states v1-2) electricity arg1 scale) arg0)
        )
```

This is somewhat tricky to fix for a few reasons.  It wouldn't be too hard to eliminate the `f0-0` after expressions and before `let`s, but knowing if it's safe to eliminate `f0-0` is a lot harder - we really only have the information and tools to do this as part of the expression building pass.

I _think_ we could identify all of these cases by looking for a
```
<store> val, addr
<move> val2, val
```
and then have this become "push the `set!` expression we'd normally do to `val2` on the expression stack". Then the usual expression stack logic would handle all the logic to eliminate temporaries.

But there's probably a lot of details to get right here.

ManDude:
Rewrites specific kinds of lets where the return value of the `set!` itself is meant to be used to just be `set!`s. Implemented at the let rewrite level. Seems to work for Jak 2 so far.

Fixes #1854 .


decompiler/analysis/insert_lets.cpp
```diff
@@ -699,6 +699,56 @@ FormElement* rewrite_empty_let(LetElement* in, const Env&, FormPool&) {
...
 }
 
+FormElement* rewrite_set_let(LetElement* in, const Env& env, FormPool& pool) {
+  /*
+   * (let ((dest-var src))
+   *   (set! something dest-var)
+   *   dest-var
+   *   )
+   * to:
+   * (set! something src)
```
water111, line 709: I think this change is unsafe to make - for example:
```
(let ((temp (foo)))
  (set! (-> (bar) field) temp)
  )
```
will call `foo` before `bar`, but

```
(set! (-> (bar) field) (foo))
```
will call `bar` before `foo`.

Which is why I think this detection has to happen before expression building.
If we do it here, we also lose the ability to detect things like `*!`

ManDude: I was wondering that too, but it looks like `set!` is an exception to the left-to-right eval order. The source is eval'd before the destination.
image

ManDude: Math macros like `*!` also seemed unaffected, looking at the refs.

ManDude:
Only problem might be that it won't cast things like enums or `time-frame` properly (see  `seagull`). Hasn't caused actual issues so far though so I won't spend time fixing it right now.

decompiler/analysis/insert_lets.cpp
water111: Yeah it looks like you're right - both the compiler and decompiler agree about this, and I made a comment about it being flipped here
image

I think that means this is safe, but I do want to think about it a little more.

The `*!` thing would only apply in cases where we don't detect `*!` currently - if you had something that was originally `(*! foo bar)`, but it got split up due to this set/let issue, then this fix will give us `(set! foo (* foo bar))`. Which I wouldn't worry about too much.

A few other things I want to think about
- `uint128`/`vf` stuff : seems like it doesn't get applied for vf stuff because that doesn't become a `set`
- dropping casts for `float`/`int` convers: I think it won't but I want to review some more
- dropping casts for `(the-as ` - not sure, I kinda think this will be wrong
- bitfield stuff: the compile makes `set!` on a bitfield return `none`, but we could change this

PR #1861, opened 2022-09-08
decomp: `logic-target`

### LINKED ISSUE #1859: [decompiler] special case `(the bitfield-enum -1)` with all 64 bits set
Instead of some terrible `(bitfield-enum bit0 bit1 ... bit63)`

Hat-Kid:
Also adds all names for `focus-status`.

Closes #1859 and updates ref tests for Jak 1 for this edge case.


goal_src/jak2/engine/collide/collide-shape-h.gc
```diff
@@ -78,6 +78,43 @@
...
+  (unknown-54 54)
+  (unknown-55 55)
+  (unknown-56 56)
+  (unknown-57 57)
+  (unknown-58 58)
+  (unknown-59 59)
+  (unknown-60 60)
+  (unknown-61 61)
+  (unknown-62 62)
+  (unknown-63 63)
```
ManDude, line 117: don't need this anymore I think

water111 (approved):
looks good!
I think there's some cleanup that we can eventually do (both decompiler bug fixes/improvements, and type stuff), but we can't really do it yet.

PR #1864, opened 2022-09-09
d/jak2: finish `task-arrow` | `carry-h` | `projectile` | `gun-[red|yellow|blue]-shot`

xTVaser:


decompiler/IR2/OpenGoalMapping.cpp
```diff
@@ -11,13 +11,12 @@ typedef OpenGOALAsm::InstructionModifiers MOD;
...
-    //    {InstructionKind::MTPC, {".nop", {MOD::SKIP_IT}}},
-    //    {InstructionKind::MFPC, {".nop", {MOD::SKIP_IT}}},
-    //    {InstructionKind::SYNCP, {".nop", {MOD::SKIP_IT}}},
-    //    {InstructionKind::SYNCL, {".nop", {MOD::SKIP_IT}}},
+    // they only use this for performance counters / Count / Debug
+    {InstructionKind::MTC0, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::MTPC, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::MFPC, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::SYNCP, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::SYNCL, {".nop", {MOD::SKIP_IT}}},
```
xTVaser, line 19: specifically calling out this change, it allows you to decompile function that have small amounts of asm (perf counter stuff) without having to manually rewrite/edit the output.  `spatial-hash` has a ton of this

decompiler/IR2/OpenGoalMapping.cpp
```diff
@@ -11,13 +11,12 @@ typedef OpenGOALAsm::InstructionModifiers MOD;
...
-    //     Debug
-    //    {InstructionKind::MTPC, {".nop", {MOD::SKIP_IT}}},
-    //    {InstructionKind::MFPC, {".nop", {MOD::SKIP_IT}}},
-    //    {InstructionKind::SYNCP, {".nop", {MOD::SKIP_IT}}},
-    //    {InstructionKind::SYNCL, {".nop", {MOD::SKIP_IT}}},
+    // they only use this for performance counters / Count / Debug
+    {InstructionKind::MTC0, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::MTPC, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::MFPC, {".nop", {MOD::SKIP_IT}}},
+    {InstructionKind::SYNCP, {".nop", {MOD::SKIP_IT}}},
```
water111, line 18: I'd rather keep these in if possible `.sync.l`/`.sync.p` is not just for performance counters, `mtc0` is for all different things.

In `gcommon.gc` there is a "fake assembly" thing that you can use to make the compiler ignore the assembly, which I'd rather use:
```
;;;;;;;;;;;;;;;;;;;;;;;
;; PC Port asm macros
;;;;;;;;;;;;;;;;;;;;;;;
(#when PC_PORT
  ;; SYNC is an EE instruction that waits for various memory access and DMA to be completed
  ;; DMA will be instant in the PC port, so these are no longer necessary
  (fake-asm .sync.l)
  (fake-asm .sync.p)
  ;; Copies the contents of a cop0 (system control) register to a gpr
  (fake-asm .mfc0 dest src)
  ;; Copies the contents of a gpr to a cop0 (system control) register
  (fake-asm .mtc0 dest src)
  )
```
Maybe we can add these here and also to `decompiler-macros.gc` (I think I forgot this)

goal_src/jak2/engine/common_objs/projectile.gc
```diff
@@ -7,3 +7,720 @@
...
+  (let ((v1-0 (-> obj root-override)))
+    (let ((a1-0 (new 'stack-no-clear 'collide-query)))
+      (set! (-> a1-0 radius) (-> v1-0 root-prim prim-core world-sphere w))
+      (set! (-> a1-0 collide-with) (-> v1-0 root-prim prim-core collide-with))
+      (set! (-> a1-0 ignore-process0) obj)
+      (set! (-> a1-0 ignore-process1) (handle->process (-> obj ignore-handle)))
+      (set! (-> a1-0 ignore-pat) (-> v1-0 pat-ignore-mask))
+      (set! (-> a1-0 action-mask) (the-as uint 1))
+      )
+    (collide-shape-moving-method-65 v1-0 (-> v1-0 transv) -977272832 0 -989855744)
```
water111, line 180: Maybe the weird `-977272832` is actually a `float`? In jak 2 we can't tell, can you try changing `collide-shape-moving-method-65` to take `float`s and see if it looks better?

goal_src/jak2/engine/common_objs/projectile.gc
```diff
@@ -7,3 +7,720 @@
...
+;; WARN: Return type mismatch int vs none.
+(defbehavior projectile-bounce-falling-post projectile-bounce ((arg0 projectile-bounce))
+  (let ((gp-0 (-> self root-override))
+        (s4-0 (new 'stack-no-clear 'collide-query))
+        (s5-0 (new 'stack-no-clear 'vector))
+        )
+    (projectile-bounce-update-velocity self)
+    (set! (-> s5-0 quad) (-> gp-0 trans quad))
+    (vector-v++! s5-0 (-> gp-0 transv))
+    (when (collide-shape-moving-method-55 gp-0 s4-0 #x80249 #x45800000 #x47a00000 #x44800000)
```
water111, line 570: The `#x4...` integers are probably floats

goal_src/jak2/engine/game/task/task-arrow.gc
```diff
@@ -5,5 +5,371 @@
...
+;; WARN: Return type mismatch int vs none.
+(defmethod deactivate task-arrow ((obj task-arrow))
+  (send-event (handle->process (-> obj hud-dist)) 'hide-and-die)
+  ((method-of-type process-drawable deactivate) obj)
+  0
+  (none)
+  )
+
+;; WARN: Return type mismatch int vs none.
+(defmethod task-arrow-method-23 task-arrow ((obj task-arrow) (arg0 vector))
```
water111, line 89: oh, I wonder if this function is for the giant yellow light thing. It might do this check to see the height of the ground, so it stops drawing once it hits something.

goal_src/jak2/engine/target/logic-target.gc
```diff
@@ -677,11 +681,11 @@
                       )
                   )
              )
-    (if (>= (- (-> self clock frame-counter) (-> self control unknown-time-frame21)) (seconds 0.1))
-        (set! (-> self control unknown-time-frame22) (-> self clock frame-counter))
+    (if (>= (- (-> self clock frame-counter) (-> self control unknown-time-frame22)) (seconds 0.1))
+        (set! (-> self control unknown-time-frame23) (-> self clock frame-counter))
```
water111, line 685: I'd rather not rename fields like this, it makes it very easy to leave behind one that uses the wrong name, and it would be really hard to catch it later.

water111 (commented):
I only went through the first files, but I'm wondering if we should wait on the gameplay code for a bit, until we have some decompiler issues fixed and more stuff figured out.
(like all the `run-function-in-proces`, the `skelgroup`, some issues with `(the-as meters <some integer>)`, the missing `vector` inline functions that turn into assembly, stuff like
```
        (new
          'process
          'carry-info
          obj
          41
          (the-as vector (new 'static 'array float 1 0.0))
          (the-as vector (new 'static 'array float 1 0.0))
          12743.111
          )
```
which doesn't really make sense.

I think once we have a bit more of `process-drawable` and collision filled out, and fix up the decompiler, this will be much easier to decompile, debug, and understand.

goal_src/jak2/engine/game/task/task-arrow.gc
xTVaser: oh you know what, that probably makes more sense than the arrow in the demo like i thought -- since the code is related to the minimap / the "rod-of-god" etc.  I guess they just left the name alone perhaps? It's certainly not an arrow anymore atleast.

goal_src/jak2/engine/target/logic-target.gc
xTVaser: There's not much i can do here, I didn't change any of the types of `control-info` in this PR, I just re-generated the code for the file

water111: hm, that's very scary then - we should try to figure out what happened.

goal_src/jak2/engine/common_objs/projectile.gc
```diff
@@ -7,3 +7,720 @@
...
+    )
+  )
+
+(defstate die (projectile)
+  :virtual #t
+  :code (behavior ()
+    (let ((v1-0 (-> self notify-handle)))
+      (send-event (handle->process v1-0) 'notify 'die)
+      )
+    ((method-of-object self process-drawable-method-17))
```
water111, line 399: looks like method is missing types

goal_src/jak2/engine/common_objs/projectile.gc
```diff
@@ -7,3 +7,720 @@
...
+  (none)
+  )
+
+;; WARN: Return type mismatch int vs none.
+(defmethod init-collision! projectile ((obj projectile))
+  "Init the [[projectile]]'s [[collide-shape]]"
+  (let ((s5-0 (new 'process 'collide-shape-moving obj (collide-list-enum hit-by-player))))
+    (set! (-> s5-0 dynam) (copy *standard-dynamics* 'process))
+    (set! (-> s5-0 reaction)
+          (the-as (function control-info collide-query vector cshape-moving-flags) cshape-reaction-projectile)
```
water111, line 418: it looks like the type of `reaction` is wrong. Can you change it to `(function control-info collide-query vector vector cshape-moving-flags)`? Casting functions like this is scary because it's easy for arguments to be lost/garbage. (I made a mistake on this exact function in jak 1 and it caused the jerkiness on the lpc slide that took forever to figure out, so I'm a bit paranoid)

goal_src/jak2/engine/common_objs/projectile.gc
```diff
@@ -7,3 +7,720 @@
...
+    (set! (-> s5-0 nav-radius) (* 0.75 (-> s5-0 root-prim local-sphere w)))
+    (let ((v1-10 (-> s5-0 root-prim)))
+      (set! (-> s5-0 backup-collide-as) (-> v1-10 prim-core collide-as))
+      (set! (-> s5-0 backup-collde-with) (-> v1-10 prim-core collide-with))
+      )
+    (set! (-> s5-0 max-iteration-count) (the-as uint 2))
+    (set! (-> s5-0 event-self) 'touched)
+    (set! (-> obj root-override) s5-0)
+    )
+  (let ((t9-3 (method-of-object (-> obj root-override) collide-shape-method-47)))
```
water111, line 692: Looks like this method is missing types. I think the type of the argument is `collide-spec`.
A good name for the method might be `set-collide-with`

goal_src/jak2/engine/common_objs/projectile.gc
```diff
@@ -7,3 +7,720 @@
...
+    #x806fd
+    (t9-3)
+    )
+  (let ((t9-4 (method-of-object (-> obj root-override) collide-shape-method-48)))
```
water111, line 696: same here. I think the argument is a `collide-spec` and the name could be `set-collide-as`

goal_src/jak2/engine/target/logic-target.gc
```diff
@@ -2737,7 +2734,9 @@
             )
           0
           )
-        (set! (-> self control reaction) target-collision-reaction)
+        (set! (-> self control reaction)
+              (the-as (function control-info collide-query vector cshape-moving-flags) target-collision-reaction)
```
water111, line 2738: same here about function type. I think adding another `vector` argument to end will get rid of this cast.

goal_src/jak2/engine/target/logic-target.gc
```diff
@@ -2746,7 +2745,7 @@
             (set! (-> a2-3 ignore-pat) (-> v1-52 pat-ignore-mask))
             )
           (set! (-> a2-3 action-mask) (collide-action solid))
-          (collide-shape-method-32 (-> self control) (-> self control transv) (the-as uint a2-3) #x45800000)
+          (collide-shape-method-32 (-> self control) (-> self control transv) a2-3 #x45800000)
```
water111, line 2748: The `#x45800000` argument should be a float.

goal_src/jak2/engine/target/logic-target.gc
```diff
@@ -3269,7 +3268,9 @@
     )
   (let ((s0-0 (new 'process 'control-info obj (collide-list-enum hit-by-others))))
     (set! (-> s0-0 dynam) (copy *standard-dynamics* 'process))
-    (set! (-> s0-0 reaction) target-collision-reaction)
+    (set! (-> s0-0 reaction)
+          (the-as (function control-info collide-query vector cshape-moving-flags) target-collision-reaction)
```
water111, line 3272: another place we should make sure to get rid of a cast

goal_src/jak2/engine/target/logic-target.gc
```diff
@@ -3384,16 +3385,17 @@
...
-        (a0-65 'process)
-        (a1-23 carry-info)
+  (set! (-> obj carry)
+        (new
+          'process
+          'carry-info
+          obj
+          41
+          (the-as vector (new 'static 'array float 1 0.0))
+          (the-as vector (new 'static 'array float 1 0.0))
```
water111, line 3395: This is probably going to cause a bug. The label types in this file for these values should probably be `vector`, not an array of 1 float. As written, the `new` method will get junk if it tries to access y, z, or w from this vector.

goal_src/jak2/engine/target/mech_suit/carry-h.gc
```diff
@@ -17,27 +18,28 @@
...
+   (max-distance     meters                   :offset-assert  68)
+   (max-pull         meters                   :offset-assert  72)
+   (min-pull         meters                   :offset-assert  76)
+   (grab-trans-blend float                    :offset-assert  80)
+   (carry-radius     meters                   :offset-assert  84)
+   (backup-radius    meters                   :offset-assert  88)
+   (joint            int8                     :offset-assert  92)
+   (mode             carry-mode               :offset-assert  93)
+   (face-dir         int8                     :offset-assert  94)
+   (local-point      vector           :inline :offset-assert  96)
```
water111, line 42: below this, `grab-quat` should be a quaternion.

goal_src/jak2/engine/target/mech_suit/carry-h.gc
```diff
@@ -47,13 +49,577 @@
...
+        (+ f26-0 (* 409.6 (- 32768.0 f30-0)))
+        )
+      )
+    )
+  )
+
+;; WARN: Return type mismatch int vs none.
+(defmethod drag! carry-info ((obj carry-info) (arg0 carry-info))
+  (let ((v1-0 (-> arg0 process)))
+    (set! (-> obj other) (the-as handle (logior (if v1-0
```
water111, line 193: darn, looks like process -> handle isn't working here. But at least this code will work for now.

water111: https://github.com/open-goal/jak-project/issues/1883

test/decompiler/reference/jak1/engine/entity/entity_REF.gc
```diff
@@ -1881,7 +1879,7 @@
       (reset-actors 'debug)
       0
       (disable-irq)
-      (.mtc0 Count r0)
+      (.mtc0 Count 0)
```
water111, line 1882: here's a good example where we'd actually want to see the `.mtc0` instruction to understand what's happening. They measure elapsed time using the `Count` register, then use this to decide how many actors to spawn.

goal_src/jak2/engine/target/logic-target.gc
water111: ok yeah this change is correct. In the future lets be careful about renaming stuff. This file is currently wrong on master.

goal_src/jak2/engine/game/task/task-arrow.gc
```diff
@@ -5,5 +5,337 @@
...
+              :version #x7
+              )
+            )
+      )
+  (set! (-> a0-0 jgeo) 0)
+  (set! (-> a0-0 janim) 2)
+  (set! (-> a0-0 mgeo 0) 1)
+  (set! (-> a0-0 lod-dist 0) 4095996000.0)
+  (add-to-loading-level a0-0)
+  )
```
water111, line 37: this looks like it should have become a `defskelgroup`... any chance this is from an older version of the decompiler?

goal_src/jak2/engine/game/task/task-arrow.gc
```diff
@@ -5,5 +5,337 @@
...
+  (set! (-> self event-hook) (-> (method-of-object self idle) event))
+  (the-as task-arrow (go-virtual idle))
+  )
+
+;; WARN: Return type mismatch process vs none.
+(defun task-arrow-spawn ((arg0 task-arrow-params) (arg1 task-arrow))
+  (let* ((gp-0 (the-as process #f))
+         (s3-0 (get-process *default-dead-pool* task-arrow #x4000))
+         (v1-1
+           (when s3-0
```
water111, line 327: I'd also expect this function to look a little nicer on the current version of the decompiler.

test/decompiler/reference/jak2/engine/game/task/task-arrow_REF.gc
```diff
@@ -0,0 +1,368 @@
+;;-*-Lisp-*-
+(in-package goal)
+
+;; failed to figure out what this is:
+(defskelgroup skel-task-arrow ctywide-arrow 0 2
```
water111, line 5: Ah yeah, I think you just need to update the actual goal source file for this one.

.github/workflows/linting.yaml
```diff
@@ -19,10 +19,13 @@ jobs:
...
+          sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-15 main'
+          sudo apt update
+          sudo apt install clang-format-15
           clang-format -version
 
       - name: Check Clang-Formatting
         run: |
           chmod +x ./third-party/run-clang-format/run-clang-format.py
-          ./third-party/run-clang-format/run-clang-format.py -r common decompiler game goalc test tools lsp --color always
+          ./third-party/run-clang-format/run-clang-format.py --clang-format-executable $(which clang-format-15) -r common decompiler game goalc test tools lsp --color always
```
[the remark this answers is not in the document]
xTVaser: I had to setup a fresh environment so I got the latest clang-format -- this time the only differences seemed valid (unlike the last time where i think clang14 broke something?).

I can downgrade on my end if we don't want to upgrade

test/decompiler/reference/jak2/engine/game/task/task-arrow_REF.gc
xTVaser: Is it not possible to eventually improve the decompiler to handle these types of `defskelgroup`s?

ManDude: It does. Check the REF file, it's detected.

xTVaser: ah k, i misinterpreted

water111 (approved):

PR #1865, opened 2022-09-09
decomp: `merc`, `merc-vu1`

Hat-Kid:
The merc VU1 program is identical to Jak 1. Jak 2's merc is doing some new texture login stuff and the DMA buffer setup is a bit different, but overall very similar.


water111 (approved):
looks good, there's just a few merge conflicts

PR #1874, opened 2022-09-12
Reduced Default Graphics Settings

himham-jak:
Per https://discord.com/channels/756287461377703987/756287461830688810/1018649195927838830
I just turned on particle culling by default, set msaa to 1 and moved some values to constants up top for easy editing. Let me know if I made any mistakes.

### DIFF
--- a/common/math/Vector.h
+++ b/common/math/Vector.h
@@ -77,7 +77,17 @@ class Vector {
     return true;
   }
 
+  bool operator==(const T other) const {
+    for (int i = 0; i < Size; i++) {
+      if (m_data[i] != other) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   bool operator!=(const Vector<T, Size>& other) const { return !((*this) == other); }
+  bool operator!=(const T other) const { return !((*this) == other); }
 
   const T length() const { return std::sqrt(squared_length()); }
 
--- a/common/util/print_float.cpp
+++ b/common/util/print_float.cpp
@@ -30,6 +30,14 @@ std::string meters_to_string(float value, bool append_trailing_decimal) {
 }
 
 /*!
+ * Wrapper around float_to_string, for printing degrees. Unlike float_to_string, it does not append
+ * decimals by default.
+ */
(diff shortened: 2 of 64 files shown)


xTVaser (changes requested):
Seems like you've intermixed swedish subtitle changes in with this PR -- things will fail because the files it references doesn't exist, you should pull those changes out into a separate PR.

goal_src/jak1/pc/pckernel-h.gc
```diff
@@ -477,7 +478,7 @@
   "Set the default ps2 settings"
 
   (set! (-> obj ps2-read-speed?) #f)
-  (set! (-> obj ps2-parts?) #f)
+  (set! (-> obj ps2-parts?) #t)
```
xTVaser, line 481: what does `ps2-parts?` do?

ManDude: yes

xTVaser:
Your experience for using git/github would be a lot better if you didn't do everything from the browser
- https://docs.github.com/en/desktop/installing-and-configuring-github-desktop/overview/getting-started-with-github-desktop

himham-jak:
Ugh, github irks me. I'll comment out the swedish stuff

himham-jak:
I do have the desktop client. I need to use it more.

xTVaser:
Don't just comment it out, remove it entirely from the changeset.  The reason this is happening is because you aren't using a separate branch -- you're adding things to `master` (which is being used for this PR)

himham-jak:
> Don't just comment it out, remove it entirely from the changeset. The reason this is happening is because you aren't using a separate branch -- you're adding more things to `master` (which is being used for this PR)

should be fixed. my bad

xTVaser:
ok to test

xTVaser (approved):
Tested and seemed to work as expected -- I think it's better for the defaults to be conservative, though it's a little weird with the particle culling as everyone seems confused why it helps performance at all.

The new `defconstants` could probably be named a little better `DEFAULT_FRAME_RATE` instead of just `FRAME_RATE` for example but im not strongly opinionated on this.

PR #1875, opened 2022-09-12
Turn off hints by default and fix the description of hints

himham-jak:
This is just my preference. I get if this change is rejected. But the description of ps2-hints was just backwards either way.


ManDude:
That setting doesn't do anything. It was originally if we wanted to add extra hints to the game, since there are a bunch of unused ones. It turned out to be too complicated too implement and not at all worth the effort.

himham-jak:
Dang. Where are the normal hints?

ManDude:
https://github.com/open-goal/jak-project/blob/master/goal_src/jak1/engine/game/settings.gc#L445

We're not gonna be disabling the tutorials by default though.

PR #1876, opened 2022-09-12
Disable hints if you restart in debug mode?

himham-jak:
pls this would make testing so much less annoying

### DIFF
--- a/goal_src/jak1/engine/game/settings.gc
+++ b/goal_src/jak1/engine/game/settings.gc
@@ -418,7 +418,7 @@
   )
 
 
-(when (zero? *setting-control*)
+(when (zero? *setting-control*) ;; initialize default settings
   (set! *setting-control* (new 'global 'setting-control 32))
   (let ((gp-0 (-> *setting-control* default))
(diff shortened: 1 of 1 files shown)


ManDude:
why not just turn them off in the settings? or load a savefile with them off.

himham-jak:
Most of the time when I'm testing, I'm not in a save file at all. Going through and testing features can mean rebooting dozens of times. Turning off hints each time is a chore

dallmeyer:
I will say it's a little frustrating having to turn them off manually very time the game reboots while you're testing stuff. I would love to just `task boot-game` and not have to deal with this.

Alternatively, if the play-hints setting could be saved globally vs per save file, that would solve this headache too

ManDude:
Nothing stops you from keeping this change locally, but I'm not sure there's really any point to merging to master. The behavior is unintuitive unless you see this single line of code. Conversely, I tested various things with hints so I'd rather keep them enabled. And I'd rather not flip flop settings around like this.

You can also load a file with just 4 button presses that you can chain very quickly.

PR #1877, opened 2022-09-13
d/jak2: get `script` decompiling, no ref tests yet

xTVaser:
Has boxed array accessing that prevents me from adding anything to ref tests (the entire file is lambdas so the access pattern that i would like to ignore happens at the top-level, can't ignore it.

This code actually already has quite a bit of original docstrings so it's not too bad in that regard considering a `script-context` can have 16 arbitrary objects.  It seems they rarely put more than a single object in the context and the types are usually obvious / are actually type checked!


decompiler/config/jak2/all-types.gc
```diff
@@ -7421,8 +7421,8 @@
   :flag-assert         #x1300000fb0
   (:methods
     (new (symbol type int) _type_ 0)
-    (add-setting (_type_ process symbol object object object) none 9)
-    (persist-with-delay (_type_ process time-frame object object float object) none 10)
+    (add-setting "Originally called `setting-set` see (anon-function 48 script)" (_type_ process symbol object object object) none 9)
+    (persist-with-delay "Originally called `setting-pers` see (anon-function 46 script)" (_type_ symbol time-frame symbol symbol float int) none 10)
```
ManDude, line 7425: we should just change the names then. I believe `remove-setting` is also meant to be `unset-setting`.

ManDude (changes requested):

decompiler/config/jak2/all-types.gc
ManDude: and `set-setting` is meant to be `reset-setting`

ManDude: see:
https://github.com/open-goal/jak-project/blob/ef22563d36e3d7c826e40e7781e9210ebde73c6a/goal_src/jak1/levels/village_common/oracle.gc#L127

xTVaser: Are we also going to change jak 1's code then?  It's a pretty large refactor for something that doesn't really add much value -- our name is still decently descriptive (with that said `add` is probably not the best).
I figured it would atleast be interesting to document but not enough to pursue -- for example `setting-pers` is a horribly descriptive name even if it is original.

It's not going to happen in this PR so you should make an issue if it's something you actually want to do.

ManDude: We should do it in Jak 1 as well yes. It's a simple find and replace so it wouldn't be too hard. I really think we should stick to the original names when we can, plus that's half the point of this project really.

If you want to do it later then that's fine.

xTVaser: imo we shouldn't stick to the original names when they can be improved for readability sake.

> If you want to do it later then that's fine.
I'm not interested in doing this work at all, I'd much rather spend my time doing something else.

ManDude (approved):

goal_src/jak2/engine/debug/editable-player.gc
```diff
@@ -15,9 +15,6 @@
 
 ;; DECOMP BEGINS
 
-(format 0 "SKIP: editable-player setup~%")
-(format #t "SKIP: editable-player setup~%")
```
water111, line 19: Does the game start with this removed?

goal_src/jak2/levels/common/airlock.gc
```diff
@@ -200,7 +205,7 @@
                                     )
                                 (and (eval!
                                        (new 'stack 'script-context (the-as basic (process->ppointer self)) self (the-as vector #f))
-                                       (the-as structure (-> obj open-test))
+                                       (the-as pair (-> obj open-test))
```
water111, line 208: I wonder if we should make this `eval!` function just take an `object` here - I kinda think it can be either a pair or a symbol.

goal_src/jak2/engine/debug/editable-player.gc
xTVaser: I wasn't sure where these came from, looks like the entire file was commented out?  Should probably just be cleared completely if it's causing an issue, we already have enough to keep track of

Issue #1894, opened 2022-09-17
Unable to decompile `history::draw-history` due to rgba related code


xTVaser:
```mips
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
; .function history-draw
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
  ;stack: total 0x50, fp? 1 ra? 1 ep? 1
  ;stack_vars: 32 bytes at 16
  ;gprs: gp s5
;; Warnings:
;; WARN: Return type mismatch int vs none.
;; ERROR: Expression building failed: In history-draw: Unmatched field: start bit 24, size 40, signed? true, type rgba
```

I can't figure out which rgba in particular the decompiler doesn't like / if it's even possible to cast around this (doesn't seem to be so far).

I was presuming it had to do with the creation of an `rgba` from a `-1` immediate, but when casting these it did nothing

```
    lw t9, add-debug-line(s7) ;; [ 41] (set! t9-3 add-debug-line)
                              ;; [] -> [t9: (function symbol bucket-id vector vector rgba symbol rgba symbol) ]
    ...
    **addiu t2, r0, -1          ;; [ 50] (set! t2-0 -1) [] -> [t2: <integer -1> ]**
```

PR #1897, opened 2022-09-17
[decompile] subdivide, wind-work, tie-work, bsp, focus

water111:
- decompile `subdivide`, `wind-work`, `tie-work`, `bsp`, `focus`
- support `ppacb` in compiler
- don't assert when bitfield stuff fails due to constant propgataion weirdness
- finish up history
- div/mod unsigned assert fix in decompiler
- empty assert fix in decompiler for failed `add` type prop
- make jak 1 performance counters "work" (just measure time)
- fix cast/typos on pcgtb/vftoi15


water111:
fixes https://github.com/open-goal/jak-project/issues/1894

PR #1898, opened 2022-09-17
[decomp] jak2: shrubbery

Francessco121:
Some notes:
- `draw-prototype-inline-array-shrub` can probably be decompiled but I'm not comfortable doing this function myself. It's got a ton of inline assembly (or at least I think it's inline assembly) and is very long. I left this marked as asm for now but I am confident in the function definition.
- The VU program is identical to Jak 1 (yay)
- There's some new `dma-test` stuff at the end of the file, not really sure what it's doing but it runs in the top-level.


water111 (approved):
Looks good!
We may be able to skip some of those more complicated functions - I think one of those is assembly, and we never fully ported it for Jak 1.

PR #1900, opened 2022-09-20
Fix infinite pause buffer bug

### LINKED ISSUE #1886: Toggling the Hud counter type resets HUD duration and allows for infinite pause buffer state
### Acknowledgements

### Describe the Bug

https://cdn.discordapp.com/attachments/183602484621606912/1017900450676215808/2022-09-09_13-51-47.mp4

If you flip between the toggles for the hud counters the hud never goes away

### How To Reproduce

Do a "pause buffer" and then flip between the hud toggles

Not needed, bug is obvious

### Expected Behavior

Should not extend the duration of the HUD

### Operating System

Windows

### OpenGOAL Version

Launcher Latest Release

### Game Version

NTSC 1.0 (black label)

No

dallmeyer:
Fixes #1886 which was introduced in #1756 
 
Changed to only read the dpad inputs for hud toggle if L2/R2 were considered held (they're not read during a pause buffer)


goal_src/jak1/engine/ui/hud.gc
```diff
@@ -64,6 +64,21 @@
                   (-> obj first-init)
                   )
               )
```
dallmeyer, line 66: `(cpad-hold? 0 l2)` / `(cpad-hold? 0 r2)` checks for keeping HUD up are a few lines up from here^

ManDude: The check that prevents the HUD from being pulled up is the `(not *progress-process*)`.

dallmeyer: Right yeah, just wanted to show that this is now behind the same barrier as L2/R2 triggering the HUD in the first place. (also no longer needed the calls to update `trigger-time`)

PR #1901, opened 2022-09-21
d/jak2: some work in `memory-usage` | `nav-graph-h` and finish `pov-camera` and `aligner`

xTVaser:


decompiler/config/jak2/all-types.gc
```diff
@@ -1766,8 +1768,8 @@
 (define-extern *cos-poly-vec* vector)
 (define-extern vector-cos-rad! (function vector vector vector))
 (define-extern vector-sincos-rad! (function vector vector vector int))
-(define-extern sincos-rad! (function (pointer float) float int))
-(define-extern sincos! (function (pointer float) float int))
+(define-extern sincos-rad! (function vector float int))
+(define-extern sincos! (function vector float int))
```
water111, line 1772: I'm trying to remember why I used `(pointer float)` here...
I think the function only writes to the first two values.

decompiler/config/jak2_ntsc_v1.jsonc
```diff
@@ -99,6 +99,8 @@
   ////////////////////////////
 
   // turn this on to extract level background graphics data
+  // NOTE/TODO - you have to comment out 2 asserts in `extract_merc` for this to work
```
water111, line 102: will be fixed soon (tm), as soon as the tfrag stuff goes in.

goal_src/jak2/engine/entity/entity-table.gc
```diff
@@ -72,7 +72,7 @@
                           :heap-size #x10000
                           )
                         (new 'static 'entity-info
-                          :ptype (type-ref nav-graph :method-count 47)
+                          :ptype (type-ref nav-graph :method-count 45)
```
water111, line 75: this looks super suspicious that it changed, but it's actually fine.

For background, in order to insert a reference to a type, the linker has to know the number of methods. This is because the linker needs to get a pointer to a type object, and if it doesn't exist already, it will allocate an "empty" type, and that needs to have the right size.

Here's what happened:

Prior to this PR, the `nav-graph` type didn't exist in `all-types.gc`. The decompiler still has an idea of how many methods the type has because it's contained in the link data. However, in Jak 2, to save space, they store number of methods / 4. So to be safe, the original runtime (and the decompiler) will make room for (stored_value * 4) + 3 methods.

Then, you added `nav-graph` type. The `deftype` flags do have the exact method number, which was only 45.  (45/4) == (47/4), so this change doesn't actually change the output for this file, and everything is ok.

goal_src/jak2/engine/gfx/hw/video.gc
```diff
@@ -6,10 +6,18 @@
...
 ;; DECOMP BEGINS
 
-(defun set-video-mode ((arg0 symbol))
-  (case arg0
+;; WARN: Return type mismatch int vs none.
+(defun set-video-mode ((tv-format symbol))
+  "Set related settings to the video mode in the settings, [[*video-params*]] and the [[*video-mode*]]
+   `ntsc` has a [[*video-mode*]] value of `0`, where as `pal` has a value of `1`
+
+   Will also set a bunch of common settings related to profiling and the camera to finalize the switch
```
water111, line 17: and, very annoyingly, if you don't call this function, the camera settings will use the default from jak 1, which makes everything stretched in jak 2 :upside_down_face:

goal_src/jak2/engine/math/transformq-h.gc
```diff
@@ -73,7 +73,7 @@
     (global-y-angle-to-point (_type_ vector) float 23)
     (relative-y-angle-to-point (_type_ vector) float 24)
     (roll-relative-to-gravity (_type_) float 25)
-    (set-and-limit-velocity (_type_ int vector float) trsqv 26)
+    (set-and-limit-velocity (_type_ int vector float) trsqv :behavior process 26)
```
water111, line 76: If this is just looking at process clocks, I want to eventually turn this into a macro that doesn't require making things a behavior.

goal_src/jak2/levels/city/common/nav-graph-h.gc
```diff
@@ -7,3 +7,224 @@
...
+    (get-angle (_type_) float 20)
+    (nav-node-method-21 (_type_) float 21)
+    )
+  )
+
+
+(defmethod nav-branch-method-11 nav-branch ((obj nav-branch))
+  "TODO
+      @returns `density * 0.0078125` - is this some kind of trick?"
+  (* 0.0078125 (the float (-> obj density)))
```
water111, line 84: I think this method (and the one below) are just to help make the `nav-branch`s smaller in memory.

They store the density as an `int8`, and this rescales that value to some more meaningful scale.

water111 (approved):
looks good, just merge conflicts!

goal_src/jak2/engine/entity/entity-table.gc
xTVaser: Ah good to know the reason, i was similarly incredibly confused/suspicious with this lol

goal_src/jak2/engine/math/transformq-h.gc
xTVaser: It is yeah, ill make an issue

PR #1909, opened 2022-09-24
[decomp] background and tfrag

water111:


water111:
tfrag rendering works, but no time of day yet.
```
(lt)
(mi)
(test-play)
(ml "goal_src/jak2/examples/debug-jak2.gc")
(launch-wasd-process)
```
then use render menu to turn on manual time of day

image

PR #1913, opened 2022-09-24
More array special cases

water111:
also make jak 1 and jak 2 behave the same way, to reduce confusion. It wasn't too bad to update jak 1.


water111:
eg:
```
g > (define *test-array* (new 'global 'boxed-array float 4))

g > (set! (-> (&+ (-> *test-array* data) (* (* 4 8) 4)) 0) 1.0)

```
is now possible, and there are fewer stupid casts related to arrays.

PR #1915, opened 2022-09-25
decomp: `target`, `target-board`, `board-states`

Hat-Kid:
Just missing `target-board-handler` because it errored with `Bad delay slot in clean_up_cond_no_else_final` and I wasn't able to figure out where the asm branch was.

Commented out `target-board-clone-anim` because it crashes even after adding `clone-anim` and `clone-anim-once`.


water111 (approved):
Awesome, this is a huge chunk of code!

I think there's possibly some cleanup still to do here (mostly decompiler improvements and maybe type stuff), but lets merge this now, then worry about that later.

PR #1934, opened 2022-09-29
[decompiler] make `(not (logtest?` work

ManDude:


water111 (approved):
thanks! No idea why some of this code in `FormExpressionAnalysis` was there but commented out...

PR #1935, opened 2022-09-30
d/jak2: finish `region` as well as some typedef files - `ctywide-obs-h` | `vehicle-h` | `rigid-body-h`

xTVaser:


decompiler/util/data_decompile.cpp
```diff
@@ -604,6 +604,17 @@ goos::Object nav_mesh_nav_control_arr_decompile(
...
 
+goos::Object xz_height_map_data_arr_decompile(const std::vector<LinkedWord>& words,
+                                              const std::vector<DecompilerLabel>& labels,
+                                              int my_seg,
+                                              int field_location,
+                                              const TypeSystem& ts,
+                                              const std::vector<std::vector<LinkedWord>>& all_words,
+                                              const LinkedObjectFile* file) {
+  return decomp_ref_to_inline_array_guess_size(words, labels, my_seg, field_location, ts, all_words,
+                                               file, TypeSpec("vector4b"), 4);
```
xTVaser, line 615: discussed before, the type is indeed likely a `(pointer int8)` but I'm not sure if that is something we can easily support in the decompiler right now or if that should be a new issue.

decompiler/config/jak2/all-types.gc
```diff
@@ -8111,15 +8111,16 @@
...
+  :bitfield #t
+  (all-off)
+  (channel)
+  (anim-name)
+  (dma-size)
+  (bounds-spheres)
+  (actors)
+  (actor-marks)
+  (special-fma-spheres)
+  (all-on))
```
water111, line 8123: Is this change right? `all-off` will now be 1 instead of 0, `channel` will be 2... and `all-on` is definitely not 95 any more.

water111 (commented):
Looks good - I think you just need to remove `all-on` and `all-off` from the scene control enum.

decompiler/config/jak2/all-types.gc
xTVaser: Yeah I'm not so sure about this either...but the naming came from the debug menu code https://github.com/open-goal/jak-project/blob/master/goal_src/jak2/engine/debug/default-menu.gc#L5579-L5580 Perhaps I'm interpreting it wrong / it's not related?

PR #1936, opened 2022-10-01
[decomp] decompile decomp.gc

water111:
A few issues:
- lwidea's fr3 is getting loaded and unloaded all the time
- the debug line drawing clipping is wrong (doesn't seem wrong in pcsx2, so I think this is on us)
- nothing actually using vis data yet
- at a large distance, our view frustum culling seems slightly too aggressive (might be that viewport scissoring is wrong)
- in the city, things seem darker as you move away. unclear how this is happening (fog?)


water111:
- lwidea issue was because `Loader` was using Jak 1's 2 level limit
- debug line clipping isn't implemented yet (`debug.gc` has some unused stuff - I want to revisit this once the region PR is in)
- darker when moving away is fog

I'll actually hook up visibility data in a second PR. I'd also like to clean this up for jak 1.

water111:
the frustum culling being off seems to be consistent with both tfrag/tie and the debug line clipping:
image

So I think we're supposed to be cutting off the bottom of the image (very possible, we're missing the whole blit-displays thing, and we had this wrong in jak 1 forever).

PR #1942, opened 2022-10-04
d/jak2: decompile `conveyor` | `elevator` | `plat` | `bouncer` | `basebutton` | `base-plat` | most of `sampler` | `simple-nav-sphere` | `simple-focus` | `elec-gate` and `blocking-plane`

xTVaser:
Most of these have been atleast partially documented / named / cleaned up as well.


goal_src/jak2/engine/common_objs/conveyor.gc
```diff
@@ -7,3 +7,492 @@
...
+            )
+          )
+        )
+      #f
+      )
+    )
+  )
+
+(defmethod conveyor-method-27 conveyor ((obj conveyor))
+  "TODO - collision related, has some dead code as well (previous iteration?)"
```
water111, line 175: This is super interesting.
This code has a `(when #t` and a `(when #f` case. The `(when #t` uses Jak 2's new collision system, the `(when #f` looks like it uses jak 1's style.

water111 (approved):
looks good!
I wouldn't worry too much about `sampler.gc` - I think it's a sampling based profiler they used to see how much time was spent in each GOAL function. Not very useful for us, and would proably require support in the compiler anyway.

Issue #1888, opened 2022-09-16
Decompiler is not properly handling negative/positive infinities


xTVaser:
```
Assertion failed: 'std::isfinite(value)'
        Source: C:/Users/xtvas/Repositories/opengoal/jak-project/common/util/print_float.cpp:90
        Function: int __cdecl float_to_cstr(float, char *, bool)
```

In `fma-sphere::fma-sphere-init-by-other` - https://github.com/xTVaser/jak-project/tree/v/fma-sphere
Also `nav-mesh::35` (-inf)

```asm
    lui v1, -128              ;; [ 64] (set! v1-18 -8388608) [] -> [v1: <uninitialized> ]
    mtc1 f0, v1               ;; [ 65] (set! f0-6 (gpr->fpr v1-18)) [v1: <uninitialized> ] -> []
```

`point-to-poly-boundary` (+inf)

```asm
    lui v1, 32640             ;; [  4] (set! v1 #x7f800000)
    mtc1 f0, v1               ;; [  5] (set! f0 (gpr->fpr v1))
```

`(method 51 nav-state)`

```asm
    lui v1, 32640             ;; [296] (set! v1-98 #x7f800000) [] -> [v1: <uninitialized> ]
    mtc1 f30, v1              ;; [297] (set! f30-0 (gpr->fpr v1-98)) [v1: <uninitialized> ] -> []
```

(method 24 warp-gate)
(post idle air-train)

water111:
Should be (mostly) fixed by https://github.com/open-goal/jak-project/pull/2039
The decompiler will output code like below, but it's up to us to make sure that this will actually work right on x86

```lisp
    (set! (-> obj dust-y) (cond
                            ((= v1-2 'nest)
                             -1105.92
                             )
                            ((= v1-2 'caspad)
                             114647.04
                             )
                            (else
                              (the-as float #x7f800000)
                              )
```
(in this case it does, it just uses this as a special flag, checked with `(when (!= f0-9 (the-as float #x7f800000))` later on)

PR #1954, opened 2022-10-11
d/jak2: decompile `ambient` | `speech` and `bigmap-data`

xTVaser:
fma-sphere not completed due to #1888


decompiler/config/jak2/all-types.gc
```diff
@@ -11156,7 +11159,13 @@
    (giftag      generic-gif-tag :inline  :offset-assert 64)
    (adnops      uint32         2        :offset-assert 80) ;; guess
    (num-strips  uint32                   :offset 76)
+   (unknown-gcf-jki1n2ik1jn2  uint32                   :offset 80)
+   (unknown-gcf-dk12n1d  uint32                   :offset 84)
+   (unknown-gcf-kj2n423  uint32                   :offset 88)
    (num-dps     uint32                   :offset 92)
+   (unknown-gcf-kjn345  uint32                   :offset 96)
+   (unknown-gcf-asdas  uint32                   :offset 100)
+   (unknown-gcf-asdkh  uint32                   :offset 104)
```
water111, line 11168: I think if you change `adnops` to `(adnops gs-adcmd 2 :inline)`, you can get rid of these unknowns.

water111: Also I am very curious to see what in ambient/speech/bigmap actually touches this...

goal_src/jak2/engine/ambient/ambient.gc
```diff
@@ -9,3 +9,537 @@
...
+      )
+    (go-virtual exit)
+    (none)
+    )
+  )
+
+(defstate exit (talker)
+  :virtual #t
+  :code (the-as (function none :behavior talker) nothing)
+  )
```
water111, line 545: the decompilation looks good, but to be honest I have 0 idea what's going on in this file.

goal_src/jak2/engine/gfx/lightning-h.gc
```diff
@@ -88,8 +89,8 @@
    (line           vector-array           :offset-assert 128)
    (meet           vector-array           :offset-assert 132)
    (path           vector-array           :offset-assert 136)
-   (start-color    uint32                 :offset-assert 140)
-   (end-color      uint32                 :offset-assert 144)
+   (start-color    gs-packed-rgba         :offset-assert 140)
+   (end-color      gs-packed-rgba         :offset-assert 144)
```
water111, line 93: I think this is probably an `rgba`.  As written, this is a reference to a `gs-packed-rgba` structure somewhere else in memory because `gs-packed-rgba` isn't a bitfield type like `rgba`.

water111: In here and in `lightning-spec`.

I don't have any evidence, but just based on how their other renderers work, I think it's more likely `rgba`s.

goal_src/jak2/engine/sound/speech.gc
```diff
@@ -7,3 +7,492 @@
...
+  (set! (-> obj request handle) (the-as handle #f))
+  (set! (-> obj request priority) -10000000000000000000000000000000000000.0)
+  0
+  (none)
+  )
+
+(defmethod speech-channel-method-13 speech-channel ((obj speech-channel))
+  (set! (-> obj id) (new 'static 'sound-id))
+  (set! (-> obj last-request handle) (the-as handle #f))
+  (set! (-> obj last-request priority) -10000000000000000000000000000000000000.0)
```
water111, line 20: wow, an even bigger negative placeholder value than they usually use!

goal_src/jak2/engine/ui/bigmap-data.gc
```diff
@@ -7,3 +7,2587 @@
 
 ;; DECOMP BEGINS
 
+(define *bigmap-compressed-layers*
+  (new 'static 'bigmap-compressed-layers
+    :layer0 (new 'static 'bigmap-info-array :data (new 'static 'inline-array bigmap-info 21
```
water111, line 12: I think that `layer0` isn't a `bigmap-info-array` - these floats don't make any sense.

Looking at the file, it seems like each layer is actually a different size sadly.
So this is probably a `(inline-array uint32)` and we'll need to modify `data_decompile.cpp` (though this is probably an easy one)

goal_src/jak2/engine/ui/bigmap-h.gc
```diff
@@ -199,57 +208,18 @@
...
-          )
+    (set! (-> gp-0 draw-tmpl gif0) (the-as uint #x90aa400000008001))
+    (set! (-> gp-0 draw-tmpl gif1) (the-as uint #x535353531))
     (set! (-> gp-0 adgif-tmpl dma-vif dma) (new 'static 'dma-tag :qwc #x6 :id (dma-tag-id cnt)))
     (set! (-> gp-0 adgif-tmpl dma-vif vif0) (new 'static 'vif-tag))
     (set! (-> gp-0 adgif-tmpl dma-vif vif1) (new 'static 'vif-tag :imm #x6 :cmd (vif-cmd direct) :msk #x1))
-    (set! (-> gp-0 adgif-tmpl gif0) (the uint64 (new 'static 'gif-tag64 :nloop #x5 :eop #x1 :nreg #x1)))
-    (set! (-> gp-0 adgif-tmpl gif1) (the uint64 (new 'static 'gif-tag-regs :regs0 (gif-reg-id a+d))))
+    (set! (-> gp-0 adgif-tmpl gif0) (the-as uint #x1000000000008005))
+    (set! (-> gp-0 adgif-tmpl gif1) (the-as uint 14))
```
water111, line 222: might be worth keeping the old ones - they are a lot nicer to read.
I've been meaning to add a hack to the decompiler so we don't need to manually do this though...

water111 (commented):
looks good, I just have a few suggestions plus one static data thing.

goal_src/jak2/engine/gfx/lightning-h.gc
xTVaser: Found my justification -- it's because of what i found in `lighting` where `gs-packed-rgba`'s are pulled off the `lightning-control` types and such to be lerp'd:
image

decompiler/config/jak2/all-types.gc
xTVaser: I found this in `lightning` as well, and the unknowns were my one reason for not finishing the file, so maybe i can now!

goal_src/jak2/engine/ui/bigmap-data.gc
xTVaser: I agree, was just a guess.  I'll wait for https://github.com/open-goal/jak-project/pull/1957 until fixing this as adding all the special cases for the fields will be a lot easier with my cleanup there.

goal_src/jak2/engine/gfx/lightning-h.gc
water111: looks like you're right!
Can you make `lightning-spec` and `lightning-state` use `gs-packed-rgba` - that should get rid of a bunch of casts.

xTVaser: When I tried, I ran into static decompilation issues (trying to create a structure from a plain value -- `gs-packed-rgba`s are subtypes of `vector4w`s).  I'd imagine it's similar to however we currently handle decompiling `rgba`s....which im not familiar with yet.

I'll try! :P

water111: hm, I just looked a little closer and the `lightning-spec`s are definitely `rgba`...
Maybe `gs-packed-rgba-lerp` actually takes in `rgba`s, and outputs a `gs-packed-rgba`?

Issue #1959, opened 2022-10-11
[decomp] `progress-static.gc` missing data


water111:
Just so we don't forget:
`progress-static.gc` is missing a bunch of data due to the weird `:dynamic` boxed array field at the end of the type.

PR #1961, opened 2022-10-11
add workaround for menu-option-list weirdness

water111:
The type makes no sense when compared to the actual data there. This modifies the type slightly, which I think is better than spending a bunch of effort to special case this in the compiler/decompiler. 

If we try this, and it turns out it doesn't work, we can always revisit it. But I think this will work.


xTVaser:
This will close https://github.com/open-goal/jak-project/issues/1959 i suspect?

xTVaser (approved):

PR #1966, opened 2022-10-14
scripts: add a script to generate `game.gp` code, comment out generated code for now

xTVaser:
Generates the `game.gp` code for all DGOs, code added to the project file has been commented out for now.

In my opinion, this is way too much in a single file -- it would be nice if this code would live directly inside the `.gd` files themselves, then everything is nicely organized.  But this approach might have issues I'm not aware of.


ManDude:
Agreed that it's too big right now. We are definitely able to put everything needed to build a single DGO file inside the `.gd` file.

water111 (approved):
Thanks!
I agree this is a lot for a single file.

I think we could easily split this up - maybe one file per level or something like that. I haven't tried, but I think you could just put `(read-file "path-to-some-other-file.gd")` in the `game.gp` file and it would include that file.

PR #1971, opened 2022-10-15
d/jak2: finish `scene`

xTVaser:
`scene-player-init` has some weird `rtype-of` usage (mostly in the `pair` clause)


decompiler/config/jak2/all-types.gc
```diff
@@ -6953,6 +6953,8 @@
   (text-x24d #x024d)
   (text-x24e #x024e)
   (text-x24f #x024f)
+  (text-781 781)
+  (text-782 782)
```
ManDude, line 6957: can you use hex for these instead?

ManDude (changes requested):

decompiler/config/jak2/all-types.gc
xTVaser: I'd rather not, since all of our enums use decimal, it creates very awkward merge conflicts for people to follow the convention used everywhere else and then have to reorder decimal/hex when merging.

I will give these proper names though.

ManDude: All the other entries in the `text-id` enum are hexadecimal though, and hexadecimal is also what's used in the text tools.

game/graphics/display.cpp
```diff
@@ -75,6 +75,7 @@ void GfxDisplay::save_display_settings() {
   json["window_ypos"] = m_last_windowed_ypos;
   std::string file_path =
       (file_util::get_user_settings_dir(g_game_version) / "display-settings.json").string();
+  file_util::create_dir_if_needed_for_file(file_path);
```
water111, line 78: thanks, I've been meaning to fix this forever

water111 (approved):
Looks good. The `rtype` stuff is a bit ugly, but I think it's fine. The normal type functions don't work if the object is a `pair`, so it makes sense that they use `rtype` here.

PR #1980, opened 2022-10-18
Spanish Game Text

xsm2:
Spanish subtitles are already merged, so this should be the only text still missing for the language. The game compiles and it's playable without problem
imagen


xTVaser:
ok to test

game/assets/jak1/text/game_text_es.gs
```diff
@@ -0,0 +1,198 @@
...
+(#x1106 "VARIEDAD 1")
+(#x1107 "VARIEDAD 2")
+
+(#x1110 "INGLÉS (UK)")
+(#x1111 "PORTUGUÉS")
+(#x1112 "PORTUGUÉS (BRASIL)")
+(#x1113 "FINÉS")
+(#x1114 "SUECO")
+(#x1115 "DANÉS")
+(#x1116 "NORUEGO")
```
ManDude, line 190: These should not be translated.

ManDude (changes requested):
Make sure the text doesn't have extra line breaks than usual at 4x3 aspect ratio.

xsm2:
I can't test it right now, but is there an easy way to check the line breaks at 4x3? Or just to navigate through all the menus?

ManDude:
You need to check everything manually, yeah.

xsm2:
I have checked all that i can right now, but there are a lot of texts that are not currently used, at least in the master branch, like #x1038. I can't check the line breaks if I can't see the text.

I have fix every line break that I could, but there are still some that are not possible to remove: it's not like the subtitles where you have a lot of characters to play with and change the translation, for example the string "SUBTITLE LANGUAGE"(17 chars) is translated to "IDIOMA DE SUBTITULOS"(20 char) and that 3 char difference in lenght creates a line break, something that happens in other current translations like the german one.

game/assets/jak1/text/game_text_es.gs
xsm2: Fixed

ManDude:
If text doesn't fit and you absolutely cannot find a way to shorten the string, you can change the size of the font by modifying the parameters passed to the calls [around here](https://github.com/open-goal/jak-project/blob/master/goal_src/jak1/pc/progress-pc.gc#L2224).

xsm2:
Wouldn't change that file change the size of the font for all the languages? Is there any way to change the size of the text only for one language like with an if or something like that? To be fair I'm a little lost here and I would really a hand, an example snippset of how to use it would be incredibly usefull. Thanks for the help!

xsm2:
Done :)

ManDude:
Are you sure you can't rephrase some options to be shorter? You can change `MOSTRAR PERSONAJE EN SUBTÍTULOS` to just `PERSONAJE EN SUBTÍTULOS` and set the `game-settings` menu size to `0.71`, and change `BRILLO DE LAS ESFERAS PRECURSOR` to just `BRILLO DE LAS ESFERAS` and keep the original size.

xsm2:
I can change it, but in the `MOSTRAR PERSONAJE EN SUBTÍTULOS` case, I think that the german translation had a similar size, so if I put only `PERSONAJE EN SUBTÍTULOS ` and change the size, there will be a line break in the german translation. 

In the `BRILLO DE LAS ESFERAS PRECURSOR` I think that maybe it's not as clear as the long choice, but it could be a good balance for the menu size as that frase it's specially long.

xsm2:
`game-settings` can't be `0.71` because the `IDIOMA DE SUBTITULOS` would have a line break so I can leave it on `0.70` and change the text or leave it to `0.62`

With `game-settings` menu size at `0.70` I confirm that the german menu would have a break line, with `0.62` there is no problem
imagen

ManDude:
Ignore german for now.

xsm2:
`game-settings` size is `0.70` and `misc-settings` is restored to it's original size

MineSAT:
> I can change it, but in the `MOSTRAR PERSONAJE EN SUBTÍTULOS` case, I think that the german translation had a similar size, so if I put only `PERSONAJE EN SUBTÍTULOS ` and change the size, there will be a line break in the german translation.
> 
> In the `BRILLO DE LAS ESFERAS PRECURSOR` I think that maybe it's not as clear as the long choice, but it could be a good balance for the menu size as that frase it's specially long.

In "BRILLO DE LAS ESFERAS PRECURSOR" change for "ESFERAS BRILLANTES" for short text maybe?

ManDude (approved):

xsm2:
> In "BRILLO DE LAS ESFERAS PRECURSOR" change for "ESFERAS BRILLANTES" for short text maybe?

In the end I left it as `BRILLO DE LAS ESFERAS`, it's true that `ESFERAS BRILLANTES` is shorter, but the main issue was not about "BRILLO/BRILLANTE" but about manteining "ESFERAS PRECURSOR" which is the translation of "PRECURSOR ORBS". 

`PRECURSOR ORB GLOW`(18) could be translated to `ESFERAS PRECURSOR BRILLANTES`(28), it's 3 characters shorter than the previous translation but even then there are 10 more characters than in the english text, and in the menu it appears with the option, something like `ESFERAS PRECURSOR BRILLANTES: NO` which would be 32 characters, it's just too much lenght for it to appear in only one line. With `BRILLO DE LAS ESFERAS` there is no need to alter the size of the menu and it's closer to the original text so I think it's better to leave it like that.

PR #1987, opened 2022-10-24
d/jak2: finish `scene-actor`, `scene-looper`, `race-part`, `rigid-body-plat`, `debug-part`, `voicebox`

xTVaser:


Taskfile.yml
```diff
@@ -120,15 +120,15 @@ tasks:
       - '{{.OFFLINETEST_BIN_RELEASE_DIR}}/offline-test --iso_data_path "./iso_data/{{.GAME}}" --game {{.GAME}}'
   offline-tests-fast:
     cmds:
-      - '{{.OFFLINETEST_BIN_RELEASE_DIR}}/offline-test --iso_data_path "./iso_data/{{.GAME}}" --game {{.GAME}} --dump_current_output --num_threads 10'
+      - '{{.OFFLINETEST_BIN_RELEASE_DIR}}/offline-test --iso_data_path "./iso_data/{{.GAME}}" --game {{.GAME}} --dump_current_output --num_threads 32'
```
water111, line 123: kinda unrelated, but if we start adding lots of dgos, there's a lot of duplicated work/memory usage because each thread will end up reading all the dgos

goal_src/jak2/levels/common/scene-actor.gc
```diff
@@ -5,5 +5,1193 @@
...
+  )
+
+
+(defmethod get-art-elem krew-npc ((obj krew-npc))
+  "Checks various things such the current actor, task status, etc to determine the right art-group data to use
+       @returns the appropriate [[art-element]] for the given NPC"
+  (-> obj draw art-group data 4)
+  )
+
+(defmethod init-skeleton! krew-npc ((obj krew-npc))
```
water111, line 534: seems kinda confusing to have both `initialize-skeleton` and `init-skeleton!`. Maybe `init-skeleton!` could be `init-art` or something like that?

water111 (approved):
looks good, just merge conflicts

Taskfile.yml
xTVaser: Yeah I somewhat expected that tradeoff, do you think it's something that could be easily improved though?  Or perhaps when we have many dgos i can try to partition the work by the dgos themselves so there isn't somewhat-random access within a single thread.

water111: yeah, I think we can either partition by dgo, or share the ObjectFileDB somehow.

I tried running jak 1 with 16 threads and very quickly ran out of ram

PR #1990, opened 2022-10-25
d/jak2: finish `sidekick`, `target-carry` and `battle`.  Close to finishing `target-death` and `warp-gate`

xTVaser:


goal_src/jak2/levels/common/battle.gc
```diff
@@ -7,3 +7,1875 @@
...
+  (let ((s3-0 (+ (-> arg0 creature-index) 1))
+        (s4-0 (new 'stack-no-clear 'vector))
+        )
+    (if (and (= s3-0 (-> arg0 attack-index)) (battle-method-25 obj arg0))
+        (return #f)
+        )
+    (get-point-in-path! (-> arg0 intro-path) s4-0 (the float s3-0) 'exact)
+    (let ((s3-1 (-> arg0 mode)))
+      (set! (-> arg0 mode) (the-as uint 2))
+      (set! (-> arg1 target)
```
water111, line 1258: I think this isn't a handle.  No idea what it is though.

water111 (approved):
Looks good, other than a mystery type

goal_src/jak2/levels/common/battle.gc
water111: this should still work, I left a note for this file and we can come back later. It'll be easier once we understand more of this code.

PR #1991, opened 2022-10-27
989snd: Support version >= 2 sound effects

Ziemas:
Supports most of the grain types now while maintaining compatibility with the old stuff (at least the subset of things jak1 uses)

Would benefit from some testing in Jak 1 to make sure I didn't break anything.

Sorry the git history is a mess, I'll do something about it later.


water111 (approved):
This is awesome, I did some testing on jak 1 and jak 2 and it seems good in both!

PR #1992, opened 2022-10-27
Fix GCC build

Ziemas:
Not sure why but using the static members here caused a linker error, but they're only used in one place so I just passed them as constants.


water111 (approved):
Good catch. I think this would work if it was `static constexpr`, or if you have something like this outside the class:
```
int FormRegressionTestJak1::game_version = GameVersion::Jak1;
```

PR #1994, opened 2022-10-28
decomp: `ocean`, `ocean-mid`, `ocean-near`, `ocean-texture`, `ocean-transition`, `ocean-vu0`

Hat-Kid:
The VU programs for the ocean renderer have changed a bit and `ocean-texture` has a bunch of new stuff, otherwise things are relatively similar to Jak 1.

This is the first time I used mips2c and I'm not sure I did it 100% right, so that should be double-checked.


game/mips2c/jak2_functions/ocean.cpp
```diff
@@ -0,0 +1,292 @@
...
+  // nop                                            // sll r0, r0, 0
+  c->daddiu(sp, sp, -8);                            // daddiu sp, sp, -8
+  c->mov64(t6, s7);                                 // or t6, s7, r0
+  c->sd(ra, 0, sp);                                 // sd ra, 0(sp)
+  c->load_symbol2(t9, cache.clip_polygon_against_positive_hyperplane);// lw t9, clip-polygon-against-positive-hyperplane(s7)
+  c->mov64(a2, t4);                                 // or a2, t4, r0
+  c->mov64(a3, t5);                                 // or a3, t5, r0
+  call_addr = c->gprs[t9].du32[0];                  // function call:
+  c->daddu(t2, a2, r0);                             // daddu t2, a2, r0
+  c->jalr(call_addr);                               // jalr ra, t9
```
water111, line 26: Unfortunately, this is a place that requires some manual modifications because they don't actually use a normal calling convention.

I think there are 3 functions: `render-ocean-quad` calls `draw-large-polygon-ocean`, which calls `clip-polygon-against-positive-hyperplane`. Calling `draw-large-polygon-ocean` and `clip-polygon-against-positive-hyperplane` don't use the normal calling convention. For example, this function is reading the value of `t4` and `t5`.

So the modification here is to comment out this `jalr` and replace it with
```cpp
clip_polygon_against_positive_hyperplane::execute(ctxt);
```
which is a shortcut that allows you to bypass to usual assembly stuff and just directly jump to that function, while keeping all the register values. You'll have to add
```cpp
#include "game/mips2c/jak2_functions/sky.h"
```
See `debug.cpp` for an example of something similar.

game/mips2c/jak2_functions/ocean.cpp
```diff
@@ -0,0 +1,292 @@
...
+  bc = c->sgpr64(t0) == 0;                          // beq t0, r0, L124
+  c->load_symbol2(t9, cache.clip_polygon_against_negative_hyperplane);// lw t9, clip-polygon-against-negative-hyperplane(s7)
+  if (bc) {goto block_11;}                          // branch non-likely
+
+  c->mov64(a2, t4);                                 // or a2, t4, r0
+  c->mov64(a3, t5);                                 // or a3, t5, r0
+  call_addr = c->gprs[t9].du32[0];                  // function call:
+  c->daddu(t2, a2, r0);                             // daddu t2, a2, r0
+  c->jalr(call_addr);                               // jalr ra, t9
```
water111, line 44: this `jalr` should be replaced with
```cpp
clip_polygon_against_negative_hyperplane::execute(ctxt);
```
too

game/mips2c/jak2_functions/ocean.cpp
```diff
@@ -0,0 +1,292 @@
...
+  c->sqc2(vf1, 192, a3);                            // sqc2 vf1, 192(a3)
+  // nop                                            // sll r0, r0, 0
+  c->sqc2(vf4, 48, a3);                             // sqc2 vf4, 48(a3)
+  // nop                                            // sll r0, r0, 0
+  c->sqc2(vf7, 96, a3);                             // sqc2 vf7, 96(a3)
+  // nop                                            // sll r0, r0, 0
+  c->sqc2(vf10, 144, a3);                           // sqc2 vf10, 144(a3)
+  c->load_symbol2(t9, cache.draw_large_polygon_ocean);// lw t9, draw-large-polygon-ocean(s7)
+  // Unknown instr: jr t9
+  return draw_large_polygon_ocean::execute(c);
```
water111, line 273: nice, it looks like you already patched this one.

goal_src/jak2/engine/gfx/ocean/ocean-mid.gc
```diff
@@ -7,3 +7,1191 @@
...
+
+(defmethod ocean-mid-setup-constants ocean ((obj ocean) (arg0 ocean-mid-constants))
+  (let ((v1-0 *math-camera*))
+    (set! (-> arg0 hmge-scale quad) (-> v1-0 hmge-scale quad))
+    (set! (-> arg0 inv-hmge-scale quad) (-> v1-0 inv-hmge-scale quad))
+    (set! (-> arg0 hvdf-offset quad) (-> v1-0 hvdf-off quad))
+    (set-vector! (-> arg0 fog) (-> v1-0 pfog0) (-> v1-0 fog-min) (-> v1-0 fog-max) 3072.0)
+    )
+  (set-vector! (-> arg0 constants) -0.25 -0.5 0.0 393216.0)
+  (let ((t9-0 matrix-local->world)
```
water111, line 20: I think this function changed in one of Vaser's recent PRs. You might need to regenerate this file.

goal_src/jak2/engine/gfx/ocean/ocean-near.gc
```diff
@@ -7,3 +7,725 @@
...
+  (set! (-> arg0 drw2-tex0 dword 0)
+        (logior (logior (the-as uint #x4000082a0) (shr (shl (log2 128) 60) 34)) (shr (shl (log2 128) 60) 30))
+        )
+  (set! (-> arg0 drw2-tex0 dword 1) (the-as uint 6))
+  (set! (-> arg0 drw2-frame dword 0) (the-as uint #xffffff00080198))
+  (set! (-> arg0 drw2-frame dword 1) (the-as uint 76))
+  (set! (-> arg0 drw3-adgif tag) (new 'static 'gif-tag64 :nloop #x1 :eop #x1 :nreg #x1))
+  (set! (-> arg0 drw3-adgif regs) (new 'static 'gif-tag-regs :regs0 (gif-reg-id a+d)))
+  (set! (-> arg0 drw3-frame dword 0) (the-as uint #x80198))
+  (set! (-> arg0 drw3-frame dword 1) (the-as uint 76))
```
water111, line 363: maybe `drw-frame` is a `gs-adcmd`?

goal_src/jak2/engine/gfx/ocean/ocean-texture.gc
```diff
@@ -7,3 +7,1698 @@
...
+      (set! (-> v1-72 4 quad) (-> obj xy88 quad))
+      )
+    (set! (-> arg0 base) (the-as pointer (-> (the-as (inline-array vector4w) (-> arg0 base)) 5)))
+    (reset-display-gs-state *display* arg0 a2-5)
+    )
+  0
+  (none)
+  )
+
+(defmethod ocean-method-80 ocean ((obj ocean) (arg0 (pointer uint)))
```
water111, line 430: `arg0` might be `(pointer rgba)`?

goal_src/jak2/engine/gfx/ocean/ocean.gc
```diff
@@ -5,5 +5,1374 @@
...
+          (.lvf vf28 (&-> v1-0 camera-temp trans quad))
+          (.mov a0-5 vf28)
+          )
+        )
+      (.lvf vf26 (&-> v1-0 inv-hmge-scale quad))
+      (.lvf vf14 (&-> v1-0 hmge-scale quad))
+      (.lvf vf25 (&-> v1-0 hvdf-off quad))
+      )
+    (.lvf vf13 (&-> *sky-work* fog quad))
+    (.lvf vf27 (&-> *sky-work* giftag-ocean quad))
```
water111, line 140: This function stores stuff in vf13 and vf27, to be used by the large polygon stuff. In OpenGOAL, we don't save the registers (x86 doesn't have enough), so we have to do some rewriting.

You can look at jak 1's version of this as an example. You can convert this to mips2c, then put
```
ocean_regs_vfs.copy_vfs_from_other(c);
```
at the very bottom.

Then, at the top of `render_ocean_quad::execute`, you put
```
c->copy_vfs_from_other(&ocean_regs_vfs);
```
and this will make sure that the large polygon drawing code has the values for vf13 and vf27 set here.

goal_src/jak2/engine/gfx/ocean/ocean.gc
```diff
@@ -5,5 +5,1374 @@
...
+      (set! (-> arg0 z) (fmax 0.0 (fmin 255.0 (-> s5-0 z))))
+      )
+    )
+  0
+  (none)
+  )
+
+(defmethod ocean-method-60 ocean ((obj ocean) (arg0 dma-buffer))
+  (local-vars (sv-48 float) (sv-52 vector) (sv-56 vector))
+  (let ((vertices (the-as (inline-array ocean-vertex) (+ #x70000000 0))))
```
water111, line 185: this line should be patched to
```
(let ((vertices (scratchpad-object (inline-array ocean-vertex))))
```
The `scratchpad-object` macro should replace any uses of `#x7000000` as an address - this was the hard-coded scratchpad on ps2.

goal_src/jak2/engine/gfx/ocean/ocean.gc
```diff
@@ -5,5 +5,1374 @@
...
+      )
+    )
+  (label cfg-18)
+  0
+  (none)
+  )
+
+(defmethod ocean-method-61 ocean ((obj ocean) (arg0 dma-buffer))
+  (local-vars (sv-48 float) (sv-52 vector) (sv-56 vector))
+  (let ((vertices (the-as (inline-array ocean-vertex) (+ #x70000000 0))))
```
water111, line 284: same here

goal_src/jak2/engine/gfx/ocean/ocean.gc
```diff
@@ -5,5 +5,1374 @@
...
+      )
+    )
+  (label cfg-18)
+  0
+  (none)
+  )
+
+(defmethod ocean-method-62 ocean ((obj ocean) (arg0 dma-buffer))
+  (local-vars (sv-48 float) (sv-52 vector) (sv-56 vector))
+  (let ((vertices (the-as (inline-array ocean-vertex) (+ #x70000000 0))))
```
water111, line 383: and here

test/decompiler/vu_reference/jak2/ocean-vu0.txt
```diff
@@ -0,0 +1,132 @@
+    .word 0x40
+    .word 0x0
+    .word 0x20
+    .word 0x0
```
water111, line 4: I think these first 4 words are actually the header, and not part of the program.

water111 (commented):
This is awesome!
I had a few small comments on some of the mips2c stuff, otherwise it looks good

goal_src/jak2/engine/gfx/ocean/ocean-mid.gc
Hat-Kid: It was actually `matrix-world->local` that changed

Issue #1440, opened 2022-06-10
[sound] some sounds stay paused after unpausing until volume is updated


ManDude:
Pretty much any looping sound will have this issue. The volume is updated when the ear position (i.e. the camera) changes.

PR #2002, opened 2022-11-03
Properly update VAG volume

Ziemas:


ManDude:
Fixes #1440

ManDude:
(nvm it doesn't)

PR #2023, opened 2022-11-16
d/jak2: finish decompiling `enemy-h` | `enemy` | `grunt` and `flitter`

xTVaser:


water111 (approved):
looks good, only issue is return type of `grunt-method-185` but I'll fix this

PR #2028, opened 2022-11-20
Jak2 decomp/hover formation

chillypepper:
This is a WIP while I'm learning the ins and outs of decompilation, but putting up what I have for 2 reasons:

- Hoping someone can double check I'm on the right path (all functions have signatures, all reasonably safe guesses for types have been put in, using "object" for type where they're not)
- Might be blocked by not being able to run the offline-tests as a PAL scrub

I'm going to look at what might be involved in making tests work for PAL, but wouldn't be surprised if I have to wait to get a black label version and come back to this after :(


chillypepper:
Still a number of errors in this branch since the last update, working my way through issues while I learn more about the decomp process. Updating in case hover-formation becomes a blocker for anything, at least there's more to work from now if someone needs to pick it up before I finish

xTVaser:
I took a brief look at the `hover-formation` progress just out of curiosity.  It's looking good! A hint on one thing:
When you see code like the following (loading a 64bit number / `sllv` involved) it's usually a `handle->process` conversion
image

If you change the type of these two fields to `handle`, that'll sort out some of the issues related to this 
image
It's a little magical, as the decompiler just detects this pattern and knows what to do with it (most of the time, sometimes it needs some coaxing).

chillypepper:
Thanks heaps for your last comment @xTVaser , that ended up progressing almost every issue I had in one go. There were a couple of other handle/process fields that I found throughout after that first change too that took a fair few other issues out which is great

I've still got some errors left, most of which fall into either `failed type prop at X: Type Error: The method with id Y of type basic could not be found.` or `Failed store: (s.f! ...)`/`Failed load: (set! ...)` if you see this and happen to have any ancient secret techniques that could fix them all in 1 go hahaha. I'm sure I'll find something to fix them soon, and then hopefully not too many issues away from a first passing build!

Hat-Kid:
The `basic` method error means that the code was trying to call a method of some ID that the type doesn't have. In this case, this likely happened because one of your fields is set to be of type `basic`, but it's actually a child of `basic`.

Judging by Vaser's screenshot, you have the field `entity` set to `basic`, but it's most likely either an `entity` or an `entity-actor`.

The failed store/load error happens when the decompiler was unable to turn a basic load/store form into something fancier, such as accessing a type's fields or array access. While the function itself decompiles fine, you will see stuff like `(s.w! ...)`, `(l.f! ...)`, `(s.d! ...)`, etc. in the output (depending on the type of load/store), such forms will later be rejected when you try to compile the file, so the decompiler needs to be able to correctly resolve these loads/stores.
The reason for it happening can be a missing cast, using the wrong type, etc.

xTVaser:
@chillypepper I took the liberty of finishing off the last little bit of `hover-formation` -- I didn't have to do much.  Just filled in a few spots that you have to manually cast (ie. they pass in a `process` but then explicitly type check it's a `process-focusable` -- this is something the decompiler doesn't automatically detect and you have to manually cast it).  And fixed up a few minor mistakes causing things to not decompile/compile successfully.

xTVaser:
ok to test

chillypepper:
@xTVaser That's awesome, thank you! My last little while trying to get this branch to work was just blindly casting everything to arrays or pointers to get around the load/store issues and see if I could get more info to work from, but very little worthwhile progress was made after the last push with guesses. Next time I'm going to see if I can find one that I can just do first before pushing to avoid holding up a section again :P

I've pulled and tested this locally and everything passes that I can see, but there are some warnings about return type mismatches - should I fix them all up and push or are those intentional? Thanks again for all your hard work!

@Hat-Kid Belated response here, sorry I thought I'd replied but apparently not - I did have the entity updated locally but your advice on looking for arrays/types in those store/load calls definitely helped me make some good progress afterwards, so thank you too!

xTVaser:
> I've pulled and tested this locally and everything passes that I can see, but there are some warnings about return type mismatches - should I fix them all up and push or are those intentional? Thanks again for all your hard work!

Most of the remaining warnings can be ignored or are false positives. I didn't check super-closely but it looked like a few of your original mistakes were probably from religiously following those warnings. I think everything is mostly fine now, but if there are some minor mistakes like that they will get ironed out pretty quickly once the code that uses `hover-formation` is decompiled.

PR #2029, opened 2022-11-20
d/jak2: cleanup more of `editable` and `editable-player`

xTVaser:
- Rough start of the SQLite integration to facilitate the SQL queries
- Cleanup and disable a little bit of code so the game no longer crashes when entering the editor
- Implement some of the mouse data stuff

image

https://user-images.githubusercontent.com/13153231/202881484-399747e7-dcdb-4e09-93e9-b561a45c8a18.mp4

This is a very old branch so best to get it merged now that it's at a decent point so it can be iterated on.


water111 (approved):
looks good.
If the c++ sql stuff gets more complicated, you might want to move stuff to a new file instead of putting it all in `kmachine.cpp`, which is already a bit of a mess of different things.

PR #2030, opened 2022-11-20
repl: Add a few new quality of life improvements

xTVaser:
- You can define a `startup.gc` in your user folder, each line will be executed on startup (deprecates the usefulness of some cli flags)
- You can define a `repl-config.json` file to override REPL settings.  Long-term this is a better approach than a bunch of CLI flags as well
  - Via this, you can override the amount of time the repl will attempt to listen for the target
  - At the same time, I think i may have found why on Windows it can sometimes take forever to timeout when the game dies, will dig into this later
- Added some keybinds for common operations, shown here

https://user-images.githubusercontent.com/13153231/202890278-1ff2bb06-dddf-4bde-9178-aa0883799167.mp4
> builds the game, connects to it, attaches a debugger and continues, launches it, gets the backtrace, stops the target -- all with only keybinds.

If you want these keybinds to work inside VSCode's integrated terminal, you need to add the following to your settings file
```json
"terminal.integrated.commandsToSkipShell": [
    "-workbench.action.quickOpen",
    "-workbench.action.quickOpenView"
  ]
```


goalc/compiler/Compiler.h
```diff
@@ -21,6 +21,7 @@
 
 #include "third-party/fmt/color.h"
 #include "third-party/fmt/core.h"
+#include "third-party/json.hpp"
```
water111, line 24: can we avoid having the `json` include in this file? It will make all the compiler stuff slow to compile because the json library is a huge mess that slows everything down.

water111 (commented):
looks good, just one minor comment about includes. Maybe we could pass just a string or a path to a file, instead of the json object.

PR #2034, opened 2022-11-20
target code bugfixes

water111:


water111:
- Support for new `daddiu`/`daddu` sequence for accessing a static array without putting it in a temporary register first
- Bugfix for `check-grab-for-collisions` return problem (was buggy in original game but got lucky)
- Decompile `effect-control`
- Decompile `collide-debug` (collision renderer and collide cache debug draw), hook up "debug engine"
- Setup for `CIB.DGO` (needed to load city slum a from menu)
- Setup L3 buckets
- Finish up `powerups.gc` (needed to set the `focus-state` flag for being in the air)
- scratchpad stuff in `water.gc`
- uncomment a bunch of stuff that works now. Very close to having the full `level-update`, just one `gui-control` method left.
- fix missing argument in `do-joint-math` (it's kind of a hack in the original game too)
- temporary workaround for `defskelgroup` issue in target/sidekick.

PR #2038, opened 2022-11-23
Added "it" voice to game_text

XedoIT:


xTVaser:
ok to test

xTVaser:
Has to merge after https://github.com/open-goal/jak-project/pull/2037 (likely the reason for the failure) ideally both these changes would be in a single PR.

PR #2039, opened 2022-11-27
[jak 2] merc

water111:


water111:
- fix handling in decompiler of weird floats. (unclear if this will work right on x86)
- add ctb, lwideb, lwidec, fea, feb, ldjakbrn dgos
- fix whatever caused decompiler crash on `ctywide-scenes` (don't remember at this point)
- get art group extraction (merc) working on jak 2
- fix issue with type pass label guessing setting a label to pointer that has no automatic or manual hint
- allow label types in the json file to work with labels followed through static pair decompilation
- mips2c for `foreground` (merc dma) and `bones` (skinning matrix calculation), fixes for jak 2 symbols
- support `uv` register in DirectRenderer (not well tested, seems to be using junk texture)
- add a bunch of buckets
- Merc2 C++ changes for jak 2
- improve a few error messages
- add missing change to jak 2 linker to move link block before starting linking to avoid writing over it with level types
- fix bug in `matrix-4x4-inverse!` causing ledge grab problems (subtracting from self to get 0). Also fixed in jak 1
- fix missing arguments in `add-fg-prim-using-line-sphere` mips2c (fixed some glitchy collision)
- various scratchpad patches
- set up the `def-art-elt` stuff so `defskelgroup`s work (and set it to error when stuff is missing, instead of silently messing things up)
- fix a very old bug in the OpenGOAL compiler's register allocator (caused should push away test to return bogus results)
- decompile: `joint-exploder`, `bones`, `ctywide-scenes` `warp-gate`, `forexita-obs`, `forexita-part`, `forexitb-part`, `prison-obs`, `prison-part`
- partial decompile: `drawable.gc`, `foreground.gc`

new stuff that runs:
- `regions`
- `dma-add-process-drawable`, the foreground engine, merc buffer stuff
- bone calculation
- c++ merc renderer

known issues:
- no emerc/generic merc/ripple/blerc/texcroll/eyes
- various blending problems in both merc and background
- sprite problems
- texture problems

PR #2045, opened 2022-12-04
offline-test: Partition by DGO and colorize/condense output

xTVaser:
This solves two main problems:
- the looming threat of running out of memory since every thread would consume duplicate (and probably not needed) resources
  - though I will point out, jak 2's offline tests seem to hardly use any memory even with 400+ files, duplicated across many threads.  Where as jak 1 does indeed use tons more memory.  So I think there is something going on besides just the source files
- condense the output so it's much easier to see what is happening / how close the test is to completing.
  - one annoying thing about the multiple thread change was errors were typically buried far in the middle of the output, this fixes that
- refactors the offline test code in general to be a lot more modular

The pretty printing is not enabled by default, run with `-p` or `--pretty-print` if you want to use it

https://user-images.githubusercontent.com/13153231/205513212-a65c20d4-ce36-44f6-826a-cd475505dbf9.mp4


test/offline/framework/orchestration.cpp
```diff
@@ -0,0 +1,291 @@
...
+                                       const OfflineTestConfig& offline_config) {
+  // TODO - pull out extractor logic to determine release into common and use here
+  OfflineTestDecompiler dc;
+  dc.config = std::make_unique<decompiler::Config>(
+      decompiler::read_config_file((file_util::get_jak_project_dir() / "decompiler" / "config" /
+                                    game_name_to_config[offline_config.game_name])
+                                       .string()));
+
+  // TODO - do I need to limit the `inputs.jsonc` as well, or is the decompiler smart enough
+  // to lazily load the DGOs as needed based on the allowed objects?
```
xTVaser, line 36: An outstanding question I had / didn't look into.  Wondering if limiting the `allowed_objects` is sufficient or if I also need to limit the DGO's the decompiler is allowed to use?

test/offline/framework/orchestration.cpp
```diff
@@ -0,0 +1,289 @@
...
+    }
+    work_colls[file.containing_dgo].art_files.push_back(file);
+  }
+
+  // Now partition by DGO so that threads do not consume unnecessary or duplicate resources
+  // this is a half-decent approximation of a greedy-knapsack approach (where the knapsack can be
+  // overstuffed) Repeatedly just add to the thread with the current least amount of work
+  //
+  // TODO - if it ends up being that it would be advantageous to split up a massive dgo into
+  // multiple threads (ie. lots in engine) then this should be improved to accomodate that.
```
xTVaser, line 119: another thing to call out, there's likely a good reason to duplicate `ENGINE.DGO` across a few threads, as it is a third of all files.

common/dma/gs.cpp
```diff
@@ -352,15 +353,17 @@ std::string GsTexa::print() const {
...
 }
 
 std::string GsPrim::print() const {
-  return fmt::format("0x{:x}, kind {}\n", data, kind());
+  return fmt::format("0x{:x}, kind {}\n", data, fmt::underlying(kind()));
 }
 
 std::string GsFrame::print() const {
-  return fmt::format("fbp: {} fbw: {} psm: {} fbmsk: {:x}\n", fbp(), fbw(), psm(), fbmsk());
+  return fmt::format("fbp: {} fbw: {} psm: {} fbmsk: {:x}\n", fbp(), fbw(), fmt::underlying(psm()),
```
water111, line 365: I'm a little confused by how this code ever worked... I thought `libfmt` would complain if you tried to print an enum directly.

common/goos/Interpreter.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include "common/util/FileUtil.h"
 #include "common/util/unicode_util.h"
+#include <common/log/log.h>
```
water111, line 14: `#include "common/log/log.h"`

common/type_system/TypeSystem.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "TypeSystem.h"
 
+#include <algorithm>
```
water111, line 10: is this change needed?

decompiler/ObjectFile/ObjectFileDB.h
```diff
@@ -185,6 +185,8 @@ class ObjectFileDB {
   void analyze_functions_ir2(
       const fs::path& output_dir,
       const Config& config,
+      const std::optional<std::function<void(std::string)>> prefile_callback,
+      const std::optional<std::function<void()>> postfile_callback,
```
water111, line 189: `std::function` already has the ability to be set or unset, so the `optional` isn't really needed. By default it is unset, and you can just check with something like:
```cpp
if (postfile_callback) {
  postfile_callback();
}
  ```

File: test/offline/config/config.h [hunk 7]
```diff
@@ -0,0 +1,32 @@
...
+  u32 num_threads;
+  bool dump_mode;
+  bool fail_on_cmp = false;
+  bool fail_on_compile = false;
+  bool pretty_print = false;
+  std::vector<std::string> dgos;
+  std::unordered_set<std::string> skip_compile_files;
+  std::unordered_set<std::string> skip_compile_functions;
+  std::unordered_map<std::string, std::unordered_set<std::string>> skip_compile_states;
+};
```
water111, line 32: thanks for splitting this up, it was getting a bit too big.

test/offline/framework/execution.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include "common/util/StringUtil.h"
+#include "common/util/diff.h"
+
+#include "goalc/compiler/Compiler.h"
+#include "test/offline/config/config.h"
+
+#include "third-party/fmt/ranges.h"
+
+// TODO - i think these should be partitioned by game name instead of it being in the filename
+// (and the names not being consistent)
```
water111, line 12: one day :P

test/offline/framework/orchestration.cpp
water111: I think it loads the DGO - it can't do this lazily because it needs to open the dgo to see what's inside.  In theory we could have it trust the `all_objs.json` file to skip dgos that don't contain needed objects, but it's not set up like that right now. It's mostly a leftover for when ObjectFileDB was a tool to unpack what was in the game.

One potential way to fix this is to unpack the DGOs as a pre-processing step (this is fast), then have ObjectFileDB just load object files it needs.

test/offline/framework/orchestration.cpp
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  dc.config->allowed_objects = object_files;
+  // don't try to do this because we can't write the file
+  dc.config->generate_symbol_definition_map = false;
+
+  std::vector<fs::path> dgo_paths;
+  for (auto& x : offline_config.dgos) {
+    dgo_paths.push_back(iso_data_path / x);
```
water111, line 55: I think you could trim down this list here to only included dgos that are in the work group.

water111 (approved):
Thanks!

I think there is possibly some more work needed to prevent duplicate opening all the DGOs, but that will be easier with this improved framework. I can take a look into this.

I'm also curious about your comment with jak 1 - I also noticed that it uses a lot more memory when decompiling in multi-threaded mode. Originally I thought each thread was creating a copy of all the object files (from DGOs and STRs).  But - this doesn't make sense now that I think about it. The allowed object list should block those unneeded objects from being processed/stored. There's code that ignores objects that aren't on the allowed list.

PR #2053, opened 2022-12-09
d/jak2: finish `nav-control` and almost finish `nav-mesh`

xTVaser:
Lots of stuff mips2c'd in both files, not very happy about it but I guess it's fine for the short-term.

I did not manually patch the DMAC stuff in `nav-mesh`'s methods yet because the existing examples I looked at were really complicated and I wasn't sure where the patch began/ended.  nav-mesh's usages of DMAC is thankfully very localized to very small methods by contrast.

Other than that, the only other method not decompiled yet in `nav-mesh` is `(method 35 nav-mesh)` which creates a structure on the stack that I just havn't been able to figure out yet.

nav-control is fully "done" and added to ref tests.


water111 (changes requested):

xTVaser:
superseeded by https://github.com/open-goal/jak-project/pull/2077

PR #2054, opened 2022-12-09
decomp: output mips2c method/function declarations to the `_disasm.gc` file

xTVaser:
When working with mips2c recently, I noticed adding the `defmethod-mips2c` or `def-mips2c` code was a manual step.  This is a bit tedious to have to go and do yourself, but more importantly you have to manually go and find the right spot in the source file else you might be declaring it too early or too late.

This will automatically output the declaration for methods, and a half-finished comment for the functions.  I wasn't able to fully output the function one because it seems the signature info from `all-types` doesn't make it all the way through -- but maybe I'm wrong or this is an easy fix?

### DIFF
--- a/decompiler/analysis/final_output.cpp
+++ b/decompiler/analysis/final_output.cpp
@@ -218,7 +218,25 @@ std::string careful_function_to_string(
   }
 
   if (!func->ir2.top_form) {
-    return ";; ERROR: function was not converted to expressions. Cannot decompile.\n\n";
+    if (func->mips2c_output) {
+      std::string output = ";; INFO: function output is handled by mips2c\n";
+      // Attempt to automatically generate the OpenGOAL code for calling the mips2c varient
+      // For methods this is - (defmethod-mips2c "(method <METHOD_ID> <TYPE>)" <METHOD_ID>)
+      if (func->guessed_name.kind == FunctionName::FunctionKind::METHOD) {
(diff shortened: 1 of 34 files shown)


decompiler/analysis/final_output.cpp
```diff
@@ -218,7 +218,25 @@ std::string careful_function_to_string(
...
+      // For methods this is - (defmethod-mips2c "(method <METHOD_ID> <TYPE>)" <METHOD_ID>)
+      if (func->guessed_name.kind == FunctionName::FunctionKind::METHOD) {
+        output += fmt::format("(defmethod-mips2c \"(method {} {})\" {} {})\n",
+                              func->guessed_name.method_id, func->guessed_name.type_name,
+                              func->guessed_name.method_id, func->guessed_name.type_name);
+      } else if (func->guessed_name.kind == FunctionName::FunctionKind::GLOBAL) {
+        // For functions it is - (def-mips2c <FUNC_NAME> (function <ARGS> <RET>))
+        // TODO - is it possible to easily get the args from `all-types` to this point?
+        output += fmt::format(";; (def-mips2c {} (function ARGS RET))\n",
+                              func->guessed_name.function_name);
```
water111, line 233: The full type should be `func->type.print()`. The name stuff is just the name used to identify the function and doesn't have the full type information.

water111 (approved):
Looks good, thanks for adding.

I left a comment about the function arguments, I think it'll be an easy fix.

PR #2074, opened 2022-12-31
add `ja` `zero` func

### LINKED ISSUE #2071: macros: `ja num! identity` does not function properly
It seems like this is the first instance of `identity` anywhere in the project, so I'm assuming it's just an untested path.

This happens in `gungame-obs::(code close gungame-door)`

```clj
      (suspend)
      (when (>= 0.0 (ja-frame-num 0))
        (ja-no-eval :num! identity :frame-num 0.0)
        (goto cfg-12)
        )
```

```
[GOOS] Evaluation error on (car num-args): Argument 0 has type [empty list] but [pair] was expected
Argument is: ()
test/decompiler/reference/jak2/decompiler-macros.gc:893
        ((eq? num! 'identity) `(set! (-> ja-ch frame-num) ,(car num-args)))
```

ManDude:
fixes #2071


xTVaser:
This seems to solve my situation, but seems to be a huge regression for existing code.

changes about 20 files like so:
image

in gungame-obs though, it's looks correct
image

ManDude:
grrrr

PR #2076, opened 2023-01-01
d/jak2: finish decompiling gungame files and `tomb-scenes`

xTVaser:
This can't be well tested yet as it's tightly coupled with the hud code, but soon hopefully.

Also noticed i missed a scenes file.


decompiler/config/jak2/label_types.jsonc
```diff
@@ -1,744 +1,3373 @@
 {
-  "profile": [["L14", "profile-work"]],
+  "profile": [
```
water111, line 2: I think this file needs formatting

water111 (approved):

PR #2087, opened 2023-01-02
Ame fixes

Ziemas:


xTVaser (approved):
Solved the issue in both spots i found:
- that region in ruins
- when killing both guards at the very end of escape

PR #2092, opened 2023-01-03
d/jak2 - initial `crimson-guard-level` decomp and also the `dig` missions

xTVaser:
Finishes a bunch of files related to crimson guards / early missions:
- `crimson-guard-level`
- `dig-digger`
- `fordumpa-obs`
- `fort-robotank`
- `fort-robotank-turret`
- `fort-turret`
- `transport-level`
- `fordumpb-obs`
- `fort-floor-spike`
- `fordumpc-obs`


goal_src/jak2/levels/common/enemy/guards/crimson-guard-level.gc
```diff
@@ -7,3 +7,4144 @@
...
+                    ((< 409.6 (vector-vector-distance s3-0 s4-0))
+                     #f
+                     )
+                    ((let ((a1-4 (new 'stack-no-clear 'vector)))
+                       (set! (-> a1-4 quad) (-> s4-0 quad))
+                       (set! (-> a1-4 w) (-> arg0 root-override2 nav-radius))
+                       (add-root-sphere-to-hash! (-> arg0 nav) a1-4 #x8046e)
+                       )
+                     #f
+                     )
```
xTVaser, line 2816: I am reasonable certain that this function does not return a `symbol`, but it seems because it is the final form in a `cond` case it needs to return a symbol else this is a compiler error:

```
Cannot use none-typed variable in a condition.
Form:
(let ((a1-4 (new (quote stack-no-clear) (quote vector)))) (set! (-> a1-4 quad) (-> s4-0 quad)) (set! (-> a1-4 w) (-> arg0 root-override2 nav-radius)) (add-root-sphere-to-hash! (-> arg0 nav) a1-4 #x8046e))
```

Something seems wrong here

goal_src/jak2/levels/dig/dig-obs.gc
```diff
@@ -7,3 +7,906 @@
...
+      )
+    (go-virtual idle)
+    (none)
+    )
+  :post (the-as (function none :behavior dig-log) transform-post)
+  )
+
+;; WARN: Return type mismatch object vs none.
+(defmethod init-from-entity! dig-log ((obj dig-log) (arg0 entity-actor))
+  (local-vars (r0-0 uint128) (a0-11 object) (a1-8 uint128) (sv-16 int))
```
water111, line 654: sv-16 needs to be a res-tag here, otherwise it'll corrupt stack memory. I'll merge and fix it

PR #2095, opened 2023-01-03
offline-tests: fail on comparison in offline-tests

xTVaser:
Two main problems fixed here:
- offline tests will fail on a comparison failure (a mistake from the re-write)
- art-group-info is committed to the repo and shared with every thread (running the tests with 1 thread, for example on the CI, and locally were producing different results)

art files are still not provided to the jak2 offline tests:
- `*-ag` files are not being output
- `art-elts.gc` is not complete, as a handful of files claim to be missing stuff

lastly, in jak1's offline tests we were also running `tpage` and `*-vis` files through the decompiler.  This omits that (they came from the `all_objs.json` file) -- is this an issue?


water111:
Thanks, this is perfect. If the only reason we have it off for jak 2 is the `art-elts.gc` missing stuff, then I think I can fix that easily.

> lastly, in jak1's offline tests we were also running tpage and *-vis files through the decompiler. This omits that (they came from the all_objs.json file) -- is this an issue?

Nope. They are needed for FR3 generation, but not offline tests. Maybe in the future we could use `-tex` files automatic make constants for texture id's like we do with art, but we can always add them back.

PR #2104, opened 2023-01-07
goalc/repl: cleanup of goalc/REPL code and some QoL improvements

xTVaser:
- lets you split up your `startup.gc` file into two sections
  - one that runs on initial startup / reloads
  - the other that runs when you listen to a target
- allows for customization of the keybinds added a month or so ago
- removes a useless flag (--startup-cmd) and marks others for deprecation.
- added another help prompt that lists all the keybinds and what they do


goalc/main.cpp
```diff
@@ -201,6 +176,8 @@ int main(int argc, char** argv) {
     status = ReplStatus::WANT_EXIT;
   }
 
+  // TODO - investigate why there is such a delay when exitting
```
water111, line 179: If the delay happens only after doing an `(mi)`, it's likely that the destructor for `Compiler` is slow - it has to basically delete/free everything. At least on linux it's only like a second or so, and I've never looked into it.  But if you wanted, you could just `std::abort` to exit really quickly.

water111 (approved):

PR #2111, opened 2023-01-09
d/jak2: finish strip mine files

xTVaser:
The level crashes similarly to drill-platform though, unclear what is the root cause of it.


goal_src/jak2/levels/strip/strip-obs.gc
```diff
@@ -7,3 +7,1783 @@
...
+      (suspend)
+      )
+    #f
+    (none)
+    )
+  :post (the-as (function none :behavior strip-hazard) transform-post)
+  )
+
+(defmethod init-from-entity! strip-hazard ((obj strip-hazard) (arg0 entity-actor))
+  (local-vars (sv-64 int))
```
water111, line 165: should be a `res-tag`

water111 (changes requested):

PR #2116, opened 2023-01-11
d/jak2: some minor fixes and workarounds to make debugging easier

xTVaser:
- worked around audio code for gungame, the tutorials now function
- auto-exit cutscenes upon entry.  This doesn't always work nicely (i didnt want to cause any side-effects due to messing with load states) but it atleast stops jak from being stuck forever waiting for the cutscene to load


goal_src/jak2/engine/scene/scene.gc
```diff
@@ -870,6 +870,10 @@
 (defstate wait (scene-player)
   :virtual #t
   :enter (behavior ((arg0 symbol))
+    ;; HACK TO NOT GET STUCK WAITING FOR SCENE TO LOAD
+    ;; progress from cutscenes however is handled by the command-list on the load-state cleanup
+    (go-virtual release)
```
water111, line 875: maybe add a print here so we don't forget?

goal_src/jak2/engine/gfx/ocean/ocean.gc
```diff
@@ -1160,10 +1160,11 @@
     (let ((f0-5 1.0)
           (f30-0 1.0)
           )
-      (when (= (-> *bigmap* bigmap-index) 13)
-        (set! f0-5 0.333)
-        (set! f30-0 0.75)
-        )
+      ;; TODO - need bigmap
+      ;; (when (= (-> *bigmap* bigmap-index) 13)
```
water111, line 1164: can we leave this code the same, and just wrap it in `(when (nonzero? *bigmap*)`?
This way if we forget to restore it once we finish bigmap, it will work

game/mips2c/jak2_functions/ocean_vu0.cpp
```diff
@@ -608,6 +609,7 @@ u64 execute(void* ctxt) {
 
 block_20:
   // daddiu v1, fp, L18                                // daddiu v1, fp, L18
+  c->load_symbol2(v1, cache.ocean_generate_verts_vector); // HACK
```
water111, line 612: good catch

goal_src/jak2/levels/gungame/gungame-obs.gc
```diff
@@ -1835,9 +1845,10 @@
     (none)
     )
   :code (behavior ()
-    (while (nonzero? (get-status *gui-control* (-> self last-sound-id)))
-      (suspend)
-      )
+    ;; TODO - disable until sound is ready
+    ;; (while (nonzero? (get-status *gui-control* (-> self last-sound-id)))
```
water111, line 1849: this is fine as long as we don't forget to bring these back.

PR #2127, opened 2023-01-14
decomp: `ai-task-h`, `whack`, `bot`, `sig`, `ash` files

Hat-Kid:
`whack` is missing a single function (`birth-func-whack-score`) that I couldn't figure out the right args for.

I was not able to test any of this code because I cannot get `atoll` to extract.


water111 (approved):
Code looks good, just merge conflicts.

PR #2130, opened 2023-01-15
docs - first chunk of work documenting the files I glossed over

xTVaser:
- started documenting the files I glossed over, some are totally done, others are just partially done
- I changed the decompiler to automatically initialize the art-group-info from the json file.  This makes updating gsrc, even a single file at a time, have consistent naming
  - Though I disabled this functionality for jak 1, as I have no idea if using the ntsc art groups will cause a regression for different versions
- fix indentation for docstrings -- it still doesn't look great, but this is now a formatting concern, rather than the docstring having a bunch of happen-stance leading whitespace.


decompiler/config.cpp
```diff
@@ -52,6 +52,14 @@ Config read_config_file(const fs::path& path_to_config_file, const std::string&
   config.streamed_audio_file_names =
       inputs_json.at("streamed_audio_file_names").get<std::vector<std::string>>();
 
+  if (cfg.contains("art_group_dump_file")) {
```
water111, line 55: thanks for adding this. I don't remember why I didn't do this in the first place.

water111 (approved):
awesome!

PR #2140, opened 2023-01-19
goalc: support static arrays of `type`s

### LINKED ISSUE #2060: Arrays of `type`s are not supported by the compiler
`target-h`

```asm
L1:
    .type array
L2:
    .word 0x5
    .word 0x5
    .type type
    .type pilot-info
    .type mech-info
    .type turret-info
    .type indax-info
    .type tube-info
```

xTVaser:
draft because using the array is a little weird still, don't feel like dealing with window's slow debugging builds today.

I get the following weird error:
```clj
(define test-array (new 'static 'boxed-array :type type vector))
gr> (-> test-array 0)
1538004        #x1777d4              0.0000        vector
gr> (type? (-> test-array 0) type)
1342757        #x147d25              0.0000        #t
gr> (new 'static (-> test-array 0))
-- Compilation Error! --
Got 3 arguments, but expected 2
Form:
(-> test-array 0)
Location:
Program string:1
(new 'static (-> test-array 0))
^
Code:
(new 'static (-> test-array 0))
```

Maybe this is expected though and the `new` method wants a symbol, not a type?

Fixes #2060


water111:
> Maybe this is expected though and the new method wants a symbol, not a type?

Yeah, `(new 'static` is entirely done at compile time - the entire expression is replaced by a pointer to data that the compiler built at compile-time.

PR #2141, opened 2023-01-19
Create Hungarian translation of game text for Jak 1

LuminarLight:
This will probably take a while, since we also have to translate all the text of the base game - Naughty Dog never translated this game to Hungarian. This PR will stay a draft until it is complete.

We realized that every letter in our alphabet was already working, apart from two: Ő and Ű (they are unique sounds, so leaving their marks wouldn't be okay).
Since I did not find that double accent thing in the jak font, I decided to use ~ (see my change in FontUtils.cpp). It is good enough, and my memory tells me that I already saw this exact same "solution" (workaround) somewhere in the past. If anyone knows a better solution, please let us know.

We chose ID 14 for the Hungarian language, as it was the lowest free ID.

**Progress tracker**
A tick here means that everything was translated. It does not mean that everything is perfect yet. We will review everything multiple times, to have the best translations possible.

Game text:
- [x] base game text
- [x] pc port text
- [ ] credits text ?

[Sziloyoo](https://github.com/Sziloyoo) helped with reviewing my changes, and gave advice/suggestions for some complicated translations.

Subtitles will be done in the future, not in this PR.


LuminarLight:
Subtitles are too much for now, so this PR will just focus on game text. We will do the subtitles at some point in the future.

LuminarLight:
It should be ready now...

I copied all of the original game text (it was in one of the output folders), and translated all to Hungarian. Then I grabbed the PC port-specific English text, and translated those as well. A lot of effort was put into making sure that the translations were grammatically correct (though I can not guarantee that it is 100% correct, more like 99%). Hungarian grammar is complicated, especially when it has to use foreign words/names.


As far as I know this is the first fully new language that was (will be) added to the game, so I could not see how others handled the credits, but I ended up not translating the credits. It felt weird. It would be better if the credits language was determined by the audio language, but that is outside the scope of this PR (and my abilities). So for now I just copied whatever is displayed in credits if original game is in English - English is the most-spoken foreign language among the age groups in Hungary that play Jak, so people will be using English audio anyway. And a Hungarian dub is not coming anytime soon.


Regarding my changes in FontUtils.cpp file: As I said, I could not find a way to perfectly display the Ő and Ű characters, but putting a ~ on top of them is close enough and I have seen this trick elsewhere. I tried adding two "acute accents" but I could not figure out how this thing works. Would that even be possible? Adding two acute accents on top of O and U to get Ő and Ű?


Sorry about the mess with the commits, I thought rebase would be nice but I could not use it properly. But I hope this is not a problem, because I assume the PR will be squash-merged, so the project should not get polluted?


I've been following the project since the beginning, checking every single commit on the main branch, so I will be able to add Hungarian translations for any new text that is added to the PC port in the future. And once the other games are more playable, I will translate text for those as well. Meanwhile [Sziloyoo](https://github.com/Sziloyoo) already started working on Jak 1 subtitles, but it will take a while. A PR will be opened when it makes enough progress.

xTVaser:
Regarding translating credits, yeah it's tricky.  Annoyingly the credits text comes from a different source -- if I recall correctly it comes from the actual level file instead of the normal text files.  Jak 2 is the same and I think I have some ideas to support translation / additions to it, but at this time there isn't an supported path.

ManDude:
> Would that even be possible? Adding two acute accents on top of O and U to get Ő and Ű?

Actually, yes! That should be possible. The format used to put accents on letter is literally just "stop advancing text automatically, offset this character by this amount, and then resume" which you can totally abuse to stack multiple characters on top of each other.

@xTVaser the credits are stored in the same file as the rest of the normal text in Jak 1.

xTVaser:
Ah my mistake then, been a while since I looked at Jak 1's text file and I just assumed it was the same.  So it's definitely possible to currently translate in Jak 1, but annoying since ideally you'd only want to partially translate the list of strings

ManDude:
There is a special `credits` form for the text file that makes it less annoying to edit, especially for single languages.

And, yes, I agree it should match the audio language, but we'll worry about that some other time.

LuminarLight:
> > Would that even be possible? Adding two acute accents on top of O and U to get Ő and Ű?
> 
> Actually, yes! That should be possible. The format used to put accents on letter is literally just "stop advancing text automatically, offset this character by this amount, and then resume" which you can totally abuse to stack multiple characters on top of each other.


Thanks for the replies everyone, and thanks for confirming that this is possible. I will try to do double acute accents then, please don't merge this until that is done.

LuminarLight:
jak1_2023_01_31_18_12_11

After a lot of experimenting, I managed to make them nice.

PR can be merged, if there are no requests/problems.

xTVaser:
ok to test

PR #2147, opened 2023-01-21
[merc2] Support emerc

water111:
This adds environment mapping support to `Merc2`, and turns it on for Jak 1 and Jak 2.

- The performance is much better
- Jak 1 can be toggled back to the old behavior with `(set! *emerc-hack* #f)`. The new environment mapping is identical to the old one everywhere I checked.
- Jak 1 still falls back to generic for ripple/texscroll/blerc/eyes - there's still no dynamic texture or vertex updating support. The eye detection stuff will sometimes flag stuff as eyes which is not eyes, which is fine, but means that generic will be used in some places where emerc could be used. For example, the shiny plates on jak's arm will be drawn with generic because jak has eyes.
- Jak 2 hasn't been checked super carefully against PCSX2 yet.
- Jak 2 still isn't technically using emerc, but instead putting emerc models in the merc bucket.
- The interface to merc is a lot different now and totally custom OpenGOAL DMA code. The original merc drawing asm doesn't run anymore.
- The FR3 format changed
- Something funky going on with foreground lighting in escape, but doesn't seem to be related to this change?

Performance comparison, jak 1, in likely the most generic-merc heavy spot:
image
image


ManDude:
Working well in Jak 1.

game/graphics/opengl_renderer/foreground/Merc2.h
```diff
@@ -79,17 +67,17 @@ class Merc2 : public BucketRenderer {
 
   static constexpr int MAX_LEVELS = 3;
   static constexpr int MAX_DRAWS_PER_LEVEL = 1024;
+  static constexpr int MAX_EVNMAP_DRAWS_PER_LEVEL = 512;
```
ManDude, line 70: Typo. Also it's probably fine to set this to `MAX_DRAWS_PER_LEVEL`?

PR #2149, opened 2023-01-21
d/jak2: decompile all `*-texture` files except `castle-texture`

### LINKED ISSUE #2051: Subtypes of `array` do not compile (or potentially decompile) as expected
This may be a bug with the decompiler output, or it might just be the compiler.

I found this issue in `oracle-texture`, where we have the following types:
```clj
(deftype texture-anim-array (array)
  ((array-data texture-anim :dynamic :offset-assert 16)
   )
...
(define-extern *darkjak-hires-texture-anim-array* texture-anim-array)
```

When decompiled, you get array definitions that look like so, which seem reasonable at a glance:
```clj
(set! *darkjak-hires-texture-anim-array*
        (new 'static 'boxed-array :type texture-anim
          (new 'static 'texture-anim
            :num-layers #x3
            :func #f
...
```

But the compiler fails for the following reason
> Typecheck failed. For define on existing symbol *darkjak-hires-texture-anim-array*, got a "(array texture-anim)" when expecting a "texture-anim-array"

Which seems to suggest that the compiler isn't aware of the fact that a `texture-anim-array` _is_ just an `(array texture-anim)`

xTVaser:
`castle-texture` is unique because it does something with the CLUT for the slime.

Fixes #2051


goal_src/jak2/levels/fortress/ammo_dump/fordumpa-texture.gc
```diff
@@ -5,9 +5,76 @@
...
+      :test (new 'static 'gs-test :ate #x1 :afail #x1 :zte #x1 :ztst (gs-ztest always))
+      :alpha (new 'static 'gs-alpha :b #x1 :d #x1)
+      )
+    )
+  )
+
+(defun set-roboscreen-alpha! ((arg0 float))
+  (format 0 "TODO - roboscreen disabled!~%")
+  ;; (set! (-> *blit-displays-work* line-color)
+  ;;       (logior (logand (-> *blit-displays-work* line-color) (the-as uint #xffffffff00ffffff))
```
xTVaser, line 57: calling this out

goal_src/jak2/levels/fortress/ammo_dump/fort-robotank-turret.gc
```diff
@@ -1069,7 +1069,8 @@
        (set! (-> v1-22 prim-core collide-with) (collide-spec jak player-list))
        )
      (logior! (-> self flags) 128)
-     (send-event (handle->process (-> self screen)) 'trigger)
+     (format 0 "TODO - disabling roboscreen!~%")
+     (send-event (handle->process (-> self screen)) 'untrigger)
```
xTVaser, line 1073: and this

goalc/compiler/compilation/Type.cpp
```diff
@@ -1051,7 +1051,8 @@ Val* Compiler::compile_heap_new(const goos::Object& form,
     auto new_method = compile_get_method_of_type(form, main_type, "new", env);
     auto new_obj = compile_real_function_call(form, new_method, args, env);
     if (making_boxed_array) {
-      new_obj->set_type(m_ts.make_array_typespec(m_ts.make_typespec(content_type)));
+      // TODO - handle array subtypes here as well?
+      new_obj->set_type(m_ts.make_array_typespec("array", m_ts.make_typespec(content_type)));
```
xTVaser, line 1055: didn't seem to cause a problem skipping this, but the question remains

PR #2157, opened 2023-01-23
Hook Jak 1 into ImGUI actor filter

Zedb0T:
This adds a Vaser's ImGUI menu from #2085 to help filter out the clutter on screen to jak 1.

https://cdn.discordapp.com/attachments/877636980912046150/1066922792924225646/2023-01-22_22-29-25.mp4


xTVaser (approved):
I've tested this out, it works pretty well, and I've ran similar code in jak 2 for weeks now with no problems.  The changes are also scoped to a `defun-debug` so this is probably as low-risk as it gets for effecting normal users.

PR #2158, opened 2023-01-23
Disable cheats with speedrun mode

Zedb0T:
Disables active cheats when `(-> *pc-settings* speedrunner-mode?)` is true, also draws speedrun verification text anytime that (movie?) is #t

### DIFF
--- a/goal_src/jak1/pc/features/speedruns.gc
+++ b/goal_src/jak1/pc/features/speedruns.gc
@@ -7,16 +7,18 @@
     ;; Update auto-splitter
     (update-autosplit-info-jak1)
     ;; Draw info to the screen
-    (speedrun-draw-settings))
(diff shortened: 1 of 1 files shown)


ManDude:
Wouldn't you have to disable this if you wanted to do ng+ or something?

Zedb0T:
Ng+ doesn't currently rely on `(-> *pc-settings* cheats)` 
image

Zedb0T:
Might be a better idea to just display any cheats are on in the verification string??
`(-> *pc-settings* cheats)`

xTVaser:
Closing this on the assumption it will be done in https://github.com/open-goal/jak-project/pull/2162 before it is merged -- if it's not, it can be reopened with the subsequent conflicts resolved

PR #2169, opened 2023-01-28
d/jak2: get `portrun` working and decompile a bunch of miscellaneous files

xTVaser:


goal_src/jak2/levels/city/common/vehicle-physics.gc
```diff
@@ -6,814 +6,3 @@
 ;; dgos: CWI
 
 ;; DECOMP BEGINS
-
-(defmethod vehicle-method-99 vehicle ((obj vehicle) (arg0 float))
-  (let ((s4-0 (new 'stack-no-clear 'vehicle-grab-rail-info))
```
xTVaser, line 11: had to revert this file, one of the methods isn't correct -- it uses `rigid-body-vehicle-constants` in a strange way that wasn't properly figured out

PR #2173, opened 2023-01-29
custom levels: add initial support for ambients

Hat-Kid:
Allows mappers to place ambient hints in their custom levels.

The syntax is as follows in the custom level JSON:
```jsonc
  "ambients": [
    {
      "trans": [-27.34, 92.7, 22.24, 10.0],
      "bsphere": [-27.34, 92.7, 22.24, 10.0], // required for the game to know when you are in range of the ambient
      "lump": {
        "name": "crystalc-ambient-1", // ambient name
        "type": "'hint",              // ambient type
        "text-id": ["int32", 4297],   // game-text-id to display (if this is >4095, it will display multiple times)
        "play-mode": "'notice"        // not sure if this needed/used with level hints
      }
    }
  ]
```

This has only been tested with ambient hints so far.

One thing that could probably be improved is the way the ambients are added to the `bsp-header`'s `drawable-tree` (in `LevelFile.cpp`), I haven't been able to figure out a nicer way to do that.

### DIFF
--- a/custom_levels/test-zone/test-zone.jsonc
+++ b/custom_levels/test-zone/test-zone.jsonc
@@ -33,7 +33,9 @@
 
    // The base actor id for your custom level. If you have multiple levels this should be unique!
    "base_id": 100,
-   
+
+  "ambients": [],
+
   "actors" : [
     {
       "trans": [-21.6238, 20.0496, 17.1191], // translation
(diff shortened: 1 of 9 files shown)


xTVaser (approved):
Only touches custom level building code

PR #2185, opened 2023-02-01
d/config: re-organize `decompiler/config` and eliminate most of the duplication

xTVaser:
Reasons for doing so include:
1. This should stop the confusion around editing the wrong config file's flags -- when for example, extracting a level.  Common settings can be in one central place, with bespoke overrides being provided for each version
2. Less verbose way of supporting multiple game versions.  You don't have to duplicate the entire `type_casts` file for example, just add or override the json objects required.
3. Makes the folder structure consistent, Jak 1's `all-types` is now in a `jak1` folder, etc.


xTVaser:
### Testing Methodology

- I wrote some  quick throwaway code that dumps the relevant config sections to a minified json file (main.cpp / config.cpp) https://github.com/xTVaser/jak-project/commit/d70884093128e29c9e21d2de77e1e7e75dd9094e
- I used this to dump all configs for ntscv1/v2/pal/jp on `master` as well as from this branch https://github.com/xTVaser/jak-project/tree/v/config-refactor-test/fixtures
- Wrote some more throwaway code to test that the json files match before and after -- thankfully nlohmann dumps in a predictable order so I didn't have to do anything more than a straight file comparison - https://github.com/xTVaser/jak-project/blob/v/config-refactor-test/fixtures/diff.py

This verifies that the configurations are exactly the same before and after for a given version.  The only thing left to check is that the extractor handles them appropriately.  So this should be ready to merge next month once that's done.

xTVaser:
Tested with ntscv1 and pal with a locally built extractor + using the latest release.  I think this is good to go.

decompiler/config/jak1/jak1_config.jsonc
```diff
@@ -153,6 +153,35 @@
       "in": "decompiler_out/jak1_pal/raw_obj/6COMMON.go",
       "out": "game/assets/jak1/patches/6common.xd3"
     }
+  },
+  "version_overrides": {
```
water111, line 157: this is awesome!

water111 (approved):
much cleaner, thank you!

PR #2189, opened 2023-02-03
game: cleanup `gk`'s CLI documentation

xTVaser:
An attempt to cleanup the last CLI interface we have left to cleanup. 
- `gk` args now follow the typical convention ie. `--proj-path` instead of `-proj-path`.
- args that are passed through to the rest of the application / the game's runtime use the typical convention of following a `--`
  - I'm thinking some args shouldn't be handled at this level ie (`-nodisplay`, `-vm`, `-novm` or `-jak2`)  These could be better documented as legitimate flags and passed in via a nice struct.  They don't seem to be used in `InitParams` but I'll triple check.

There's a temporary shim here so there is no coupled release with the launcher (right now it executes `gk` with a few args).  So I just change the old args into the new format.  After one release cycle, I can change it in the launcher and delete it here.

I am unsure if this will break the bash shellscript usages -- not sure which args were usually passed into `$@`

image


xTVaser:
Tested `gk` with the existing launcher release (which uses the following args). 

https://github.com/open-goal/launcher/blob/271a50c3f120f9049630a37f513b11d8574ee412/src/lib/launch.ts#L19-L24

Boots as expected.
image

PR #2195, opened 2023-02-07
Add Speedrun Category Extensions

dallmeyer:
https://cdn.discordapp.com/attachments/1013163148179673149/1076637246905655346/allcutscenes.mp4


xTVaser (approved):
Tried all the new menu options, they work. This code appears to be well isolated.

Something that might want to be improved in the future -- can't open the speedrunner menu during a cutscene (might be annoying for an all cutscenes run?)

dallmeyer:
> Something that might want to be improved in the future -- can't open the speedrunner menu during a cutscene (might be annoying for an all cutscenes run?)

I took a look into this - right now the speedrun menu shares a fair bit of code with the regular start menu and just forks right before the normal start menu is shown. So to support it during cutscenes, we'd probably need to move it further upstream so that it doesn't get blocked by the `(progress-allowed?)` call. Or I suppose we could return true in `progress-allowed?` if speedrunner mode is enabled and L1+R1+X are being held.

https://github.com/open-goal/jak-project/blob/9b80bca5898b883a8933ba45674aed5fdc65727d/goal_src/jak1/engine/game/main.gc#L220-L226

PR #2214, opened 2023-02-12
docs: Automatically generate documentation from goal_src code

### LINKED ISSUE #2215: A failing `--cmd` in the REPL will not terminate, but instead hang on the prompt
Example
image

xTVaser:
This automatically generates documentation from goal_src docstrings, think doxygen/java-docs/rust docs/etc.  It mostly supports everything already, but here are the following things that aren't yet complete:
- file descriptions
- high-level documentation to go along with this (think pure markdown docs describing overall systems that would be co-located in goal_src for organizational purposes)
- enums
- states
- std-lib functions (all have empty strings right now for docs anyway)

The job of the new `gen-docs` function is solely to generate a bunch of JSON data which should give you everything you need to generate some decent documentation (outputting markdown/html/pdf/etc).  It is not it's responsibility to do that nice formatting -- this is by design to intentionally delegate that responsibility elsewhere.  Side-note, this is about 12-15MB of minified json for jak 2 so far :)

In our normal "goal_src has changed" action -- we will generate this data, and the website can download it -- use the information to generate the documentation at build time -- and it will be included in the site.  Likewise, if we wanted to include docs along with releases for offline viewing, we could do so in a similar fashion (just write a formatting script to generate said documentation).

Lastly this work somewhat paves the way for doing more interesting things in the LSP like:
- whats the docstring for this symbol?
- autocompleting function arguments
- type checking function arguments
- where is this symbol defined?
- etc

Fixes #2215


goalc/compiler/compilation/CompilerControl.cpp
```diff
@@ -616,3 +628,185 @@ Val* Compiler::compile_print_debug_compiler_stats(const goos::Object& form,
...
+  auto args = get_va(form, rest);
+  va_check(form, args, {goos::ObjectType::STRING}, {});
+
+  const auto& doc_path = fs::path(args.unnamed.at(0).as_string()->data);
+  lg::info("Saving docs to: {}", doc_path.string());
+
+  const auto symbols = m_symbol_info.get_all_symbols();
+
+  std::unordered_map<std::string, Docs::SymbolDocumentation> all_symbols = {};
+  std::unordered_map<std::string, Docs::FileDocumentation> file_docs = {};
```
water111, line 642: minor nit: in C++ stuff like `std::vector`, string, map, optional, will all default being empty, so you don't need the ` = {}`.

goalc/compiler/SymbolInfo.h
```diff
@@ -117,38 +153,54 @@ class SymbolInfo {
...
   }
 
   void add_fwd_dec(const std::string& name, const goos::Object& defining_form) {
     m_map[name]->push_back(SymbolInfo::make_fwd_declared_sym(name, defining_form));
   }
 
-  void add_function(const std::string& name, const goos::Object& defining_form) {
-    m_map[name]->push_back(SymbolInfo::make_function(name, defining_form));
+  // The m_symbol_types container stores TypeSpecs -- this does have argument information but not
+  // the names! The actual `LambdaVal` however does have all the types and names, so thats why its
```
water111, line 167: I think this comment is just out of date, but we should make sure to not store `LambdaVal*`s - these are supposed to be temporary just while a function is being compiled. Once the compiler moves on to the next file, they will be destroyed.

goalc/compiler/SymbolInfo.h
```diff
@@ -48,66 +59,91 @@ class SymbolInfo {
...
     return info;
   }
 
   const std::string& name() const { return m_name; }
-  const std::string& type() const { return m_type_of_method; }
+  const MethodInfo& method_info() const { return m_method_info; }
   Kind kind() const { return m_kind; }
   const goos::Object& src_form() const { return m_def_form; }
+  const Metadata& meta() const { return m_meta; }
+  const std::vector<GoalArg> args() const { return m_args; }
```
water111, line 136: missing `&`

goalc/compiler/SymbolInfo.h
```diff
@@ -31,11 +39,14 @@ class SymbolInfo {
...
 
-  static SymbolInfo make_global(const std::string& name, const goos::Object& defining_form) {
+  static SymbolInfo make_global(const std::string& name,
+                                const goos::Object& defining_form,
+                                const std::optional<Metadata> meta = {}) {
     SymbolInfo info;
     info.m_kind = Kind::GLOBAL_VAR;
     info.m_name = name;
     info.m_def_form = defining_form;
+    info.m_meta = meta ? meta.value() : Metadata();
```
water111, line 49: minor, but I think this would more normally be written as
```
if (meta) {
  info.m_meta = *meta;
}
// no need to set it to Metadata(), the default constructor of SymbolInfo does this for us.
```
or
```
info.m_meta = meta.value_or({});
```
which sets it to `meta` if it exists, or `{}` otherwise.

common/goos/ParseHelpers.cpp
```diff
@@ -72,11 +77,15 @@ bool va_check(
     } else {
       // argument given.
       if (kv.second.second.has_value() && kv.second.second != kv2->second.type) {
-        // but is wrong type
-        *err_string = "Argument \"" + kv.first + "\" has type " +
-                      object_type_to_string(kv2->second.type) + " but " +
-                      object_type_to_string(kv.second.second.value()) + " was expected";
-        return false;
+        // special case -- an empty list is a valid pair
+        if (!(kv.second.second == goos::ObjectType::PAIR &&
```
water111, line 81: I think this will break error messages for a lot of things. An `EMPTY_LIST` in GOOS isn't actually a pair. So `foo.as_pair()` will fail on an empty list.

There's a lot of existing code that assumes that it's safe to call `.as_pair` if the `va_check` passes with the `::PAIR` type, and this change would break that assumption. I think it's better if functions that want either a pair or an empty string to manually check this.

goalc/compiler/compilation/Macro.cpp
```diff
@@ -203,7 +206,8 @@ Val* Compiler::compile_define_constant(const goos::Object& form,
     m_goos.global_environment.as_env()->vars[sym] = value;
   }
 
-  m_symbol_info.add_constant(sym->name, form);
+  m_global_constant_types.emplace(sym->name, value);
```
water111, line 209: It looks like this isn't really types, but just the value of the constant. I'd recommend just using `m_gloal_constants` instead.

goalc/compiler/compilation/CompilerControl.cpp
```diff
@@ -616,3 +628,185 @@ Val* Compiler::compile_print_debug_compiler_stats(const goos::Object& form,
...
+    }
+
+    // TODO - states / enums / built-ins
+    if (sym_info.kind() == SymbolInfo::Kind::GLOBAL_VAR ||
+        sym_info.kind() == SymbolInfo::Kind::CONSTANT) {
+      Docs::VariableDocumentation var;
+      var.name = sym_info.name();
+      var.description = sym_info.meta().docstring;
+      if (sym_info.kind() == SymbolInfo::Kind::CONSTANT) {
+        var.type = object_type_to_type_name(m_global_constant_types.at(var.name).type);
```
water111, line 696: You can't really know the type of a constant - they are just replaced as text directly so it can be context-dependent. For example, the type of `foo` here can be both string or int.
```
g > (defglobalconstant foo bar)
g > (let ((bar "a")) (print-type foo))
[TYPE] string igpr-1
g > (let ((bar 1)) (print-type foo))
[TYPE] int igpr-1
```

that said, I don't think we abuse constants like this in any existing code, so maybe it's ok if these are wrong.

common/goos/Object.cpp
```diff
@@ -83,6 +83,31 @@ std::string object_type_to_string(ObjectType type) {
   }
 }
 
+// converts an object type to it's symbol name to be converted into a typespec
+// returns 'object' otherwise
+std::string object_type_to_type_name(ObjectType kind) {
```
water111, line 88: do you mind moving this function to somewhere like CompilerControl.cpp? I think this is confusing to have in the general goos folder because this is sort of converting between goos and goal types.

common/goos/Object.cpp
```diff
@@ -83,6 +83,31 @@ std::string object_type_to_string(ObjectType type) {
...
+    case ObjectType::FLOAT:
+      return "float";
+    case ObjectType::CHAR:
+      return "uint";
+    case ObjectType::SYMBOL:
+      return "symbol";
+    case ObjectType::STRING:
+      return "string";
+    case ObjectType::PAIR:
+      return "pair";
```
water111, line 101: I think this isn't going to do the right thing for some constants. Eg:
```
(defglobalconstant foo (+ 1 2))
```
will be `pair` according to this function. I'd recommend just leaving off types for constants for now. GOAL doesn't really support typed constants.

water111 (commented):
Looks good! Just a few minor comments.

PR #2229, opened 2023-02-20
Speedrun verification text cleanup

dallmeyer:
We only draw this text to the screen if we're in speedrunner mode anyway, so the `#t` flag on this line doesn't really add any value imo
https://github.com/open-goal/jak-project/blob/b130b2eac27966c43f6816ee604a3a93a67e82fa/goal_src/jak1/pc/features/speedruns.gc#L248-L255

image


Zedb0T:
It adds value, it shows how it displays to users who watch a speedrun that do not know why it displays for the runner but not for them.

xTVaser:
> It adds value, it shows how it displays to users who watch a speedrun that do not know why it displays for the runner but not for them and also just kind of explains why the menu is there. Without it the text looks extremely out of place imo

True, perhaps make it more general then, like
```
Speedrunner Mode Details:
...
...
```

Instead of the current symbol display which will always be #t -- which quite frankly I suspect most users won't make the connection that it means true/enabled.  Could also cut down `speedrun category` to just `category`

xTVaser (approved):

PR #2230, opened 2023-02-20
[jak2] fix rare civilian flee crash

VodBox:
Fixes a rare crash that can sometimes happen when civilians are picking a direction to flee in, where a branch count may be 0, causing a divide by 0 MATH_EXCEPTION.

### DIFF
--- a/goal_src/jak2/levels/city/common/civilian.gc
+++ b/goal_src/jak2/levels/city/common/civilian.gc
@@ -403,7 +403,8 @@
         (dotimes (s0-0 s2-0)
           (+! f30-0 (civilian-method-214 obj (-> s3-0 branch-array s0-0) (+ arg1 -1) arg2 s1-0))
           )
-        (set! arg3 (+ s1-0 (the float (/ (the int f30-0) s2-0))))
+        ;; changed to fix a divide by zero crash
+        (set! arg3 (+ s1-0 (the float (/-0-guard (the int f30-0) s2-0))))
         )
       )
     )


Hat-Kid (changes requested):

goal_src/jak2/levels/city/common/civilian.gc
```diff
@@ -403,7 +403,12 @@
         (dotimes (s0-0 s2-0)
           (+! f30-0 (civilian-method-214 obj (-> s3-0 branch-array s0-0) (+ arg1 -1) arg2 s1-0))
           )
-        (set! arg3 (+ s1-0 (the float (/ (the int f30-0) s2-0))))
+        (set! arg3
```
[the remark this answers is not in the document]
VodBox: Is there a preferred format for indicating a manual patch?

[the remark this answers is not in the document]
Hat-Kid: Not really, just a small comment saying something like e.g. `;; changed to fix civilian crash` is sufficient just so we know that this section was modified by hand in case we go back and modify the decompiled output of this file.

Hat-Kid:
ok to test

goal_src/jak2/levels/city/common/civilian.gc
```diff
@@ -403,7 +403,13 @@
         (dotimes (s0-0 s2-0)
           (+! f30-0 (civilian-method-214 obj (-> s3-0 branch-array s0-0) (+ arg1 -1) arg2 s1-0))
           )
-        (set! arg3 (+ s1-0 (the float (/ (the int f30-0) s2-0))))
+        ;; changed to fix a divide by zero crash
+        (set! arg3
+          (if (nonzero? s2-0)
+            (+ s1-0 (the float (/ (the int f30-0) s2-0)))
+            s1-0
+            )
```
ManDude, line 411: Don't do this actually, just replace the `/` with `/-0-guard`. An integer division by zero should return -1 and the macro takes care of that.

ManDude (changes requested):

goal_src/jak2/levels/city/common/civilian.gc
VodBox: Changing the ``/`` to a ``/-0-guard`` would definitely be producing the wrong answer for what this function is intending.

It's supposed to be determining the average distance away from a particular point a civilian can take if it moves along a certain branch. ``s1-0`` contains the distance of the current destination node, and the division of ``f30-0`` by ``s2-0`` is supposed to produce the average of all the branches onwards from that node. If branch count is zero, then it shouldn't be adding anything, while ``/-0-guard`` would result in subtracting 1.

ManDude: If that wrong answer isn't causing a game crash or nasty bug, it should still be used.

VodBox: I'm confused why it should still be used if it'll produce the wrong answer. Is the returning of -1 the behavior you'd get on the original PS2 when dividing by 0, and so this would then be bug-for-bug compatible? Or is it just the principle of making the smallest changes possible?

xTVaser: As the macro suggests, it imitates how the PS2 would divide by zero https://github.com/open-goal/jak-project/blob/a76f55853779118efe98a3115c7d78546ae40b6c/goal_src/goal-lib.gc#L653-L655 which is allowed.  The PS2 doesn't use IEEE floating point

ManDude: Actually, because this is a `div` and not a `divu` it would only be -1 if the dividend is non-negative, otherwise it would be one. <https://github.com/PCSX2/pcsx2/blob/eec0984dbef8e3bbf8a9f0a9f9cf050f7c88c393/pcsx2/R5900OpcodeImpl.cpp#L412>

ManDude: (that said, it doesn't appear like the branch count could ever be negative, so it should always be -1. but feel free to make an even safer macro if you want, I doubt this will cause any performance issues ever)

ManDude: (sorry, I'm dumb and tired and misread the code - I'm not actually sure if this number can or cannot be negative so just use the safest option)

PR #2239, opened 2023-02-22
Initial Jak 2 Autosplit Support

Zedb0T:
#|Notes here
TODO:

Misc:
No need to worry about playhints, maybe worry about speech volume instead? 
or we can make speech volume not delay prompts whether its on or off? 
Playing with no voices can get lonely

Currently no need to worry about all actos
Currently does not turn on auto-save - this is probably worth changing later but would need to allow restart mission/pause menu while autosave icon is up.
Currently does not enforce a fps, Jak 2 has dynamic FPS support so this probably isnt an issue
setup-speedrun-post-blackout needs work
Cant turn off pc-cheats that dont exist
End Notes|#

Related to https://github.com/open-goal/speedrunning/pull/12


Hat-Kid:
Just a note on the FPS: I've been running the game at 165 FPS for quite a while now and one crucial thing I noticed is that this uppercut to get on the ledge above is not possible, so this and any other gameplay altering behavior should be looked into before allowing variable framerate in speedruns.
image

goal_src/jak2/pc/features/autosplit-h.gc
```diff
@@ -0,0 +1,148 @@
...
+;;
+;; DO NOT change the order, appending to the end is safe!
+;;Hacks here
+
+
+;;End Hacks
+
+(deftype autosplit-info-jak2 (structure)
+  (;; General stats
+   (num-power-cells uint32)
```
xTVaser, line 16: something i should have did in jak 1 from the get-go -- let's append 2 version numbers (major and minor) to the beginning of the struct, two `uint16`s.  This will let us make breaking changes / roll out changes conditionally instead of having coupled releases between jak-project and the autosplitter script.

PR #2240, opened 2023-02-22
jak2: add back some pc port debugging tools and add `palace-scenes` to gsrc

Hat-Kid:
Ports some of the custom PC port debugging code from Jak 1, such as the entity debugger, Anim Tester X and part tester, to Jak 2 and gives the PC menu code the same treatment Jak 1 received recently in #2216.
Some adjustments have been made to make things work in Jak 2 and some stuff is still incomplete.
The majority of the default PC menu code has been left commented out in preparation for @ManDude's plan to port the rest of the Jak 1 `pckernel` features to Jak 2.

Also adds the code for `palace-scenes` to gsrc as that seems to have been missed.


goal_src/jak2/engine/debug/part-tester.gc
```diff
@@ -13,7 +17,7 @@
 (declare-file (debug))
 
 (defpartgroup group-part-tester
-  :id 127
+  :id 96
```
ManDude, line 20: Why was this changed?

goal_src/jak2/engine/game/main.gc
```diff
@@ -1276,6 +1276,48 @@
...
+  `(let* ((heap ,heap) (pool-total (memory-total heap)))
+      (draw-memory-bar-generic ,buf
+              :remain (- pool-total (memory-used heap))
+              :total  pool-total
+              :name ,(if name name (symbol->string heap))
+              :idx ,idx
+              :color ,color)
+      )
+  )
+
```
ManDude, line 1320: Pretty happy all of the new stuff I wrote still works fine even in a completely different game!

ManDude (changes requested):
Aside from the thing in `part-tester` it all looks fine to me. I'd like to have this merged before going forward with the `pckernel` porting.

goal_src/jak2/engine/debug/part-tester.gc
Hat-Kid: Main reason was that the `let` on line 71 was commented out and replaced with a `let` that uses id 96 instead for whatever reason. I don't think it matters much which id is used since neither seem to be used by the game.

ManDude: I think that was just them testing whatever 96 was at the time. IMO just keep it at 127, there's checks in place to make sure invalid part groups aren't tested anyway.

ManDude (approved):

PR #2242, opened 2023-02-23
Translate new Jak 1 speedrunning text to Hungarian, plus some minor adjustments

LuminarLight:
I saw that some new speedrunning-related text was added to the English language, so I decided to translate them. Meanwhile I also made some minor adjustments/fixes to existing translations.

BUT I have a question...
Do we really need all this text?
image
They are displayed in a menu that is called 'NEW INDIVIDUAL LEVEL RUN'. So adding IL after the level names is kind of redundant in my opinion - the existing level name text entries could be used instead. And the reason why I care so much is because it is hard to come up with a "translation" for IL. But if I am forced to, I will come up with something.

I will keep this PR a draft until I know what the main jak-project people think about this "IL" issue. And I will adjust my branch accordingly.


LuminarLight:
I see that the "level name + IL" text will stay in the project. 

In that case this PR is no longer a draft. It can be merged if everything is okay.

xTVaser:
> I see that the "level name + IL" text will stay in the project.

Feel free to change it

LuminarLight:
Would something like that be accepted?
My idea is that the code could be wired up to the already existing level names. And then this "level name + IL" thing could be removed from game_text files.
It would be clearer to do that in a separate PR though. Once this one is merged I will open such a PR and maybe a discussion can be done in there if necessary.

xTVaser:
Yeah that sounds like exactly what has to be done, there's already a few spots where that pattern already exists.

I imagine you'd just have a single translated string for "IL" and then use the level names that don't change between languages (iirc?)

LuminarLight:
I would get rid of "IL" as well, since the menu name already says "Individual Level run". But your suggestion would also be better than the current situation.

LuminarLight:
Yeah but as I said that should be in a separate PR I believe.

So this PR is okay by me, can be merged if there are no problems.

water111 (approved):
Thanks!

PR #2244, opened 2023-02-24
Fix Jak 2 warp gates 

dallmeyer:
Blatant plagiarism from Jak 1.
From what I could tell, nothing else is sending `change-state` to `target`, so the only thing I tested was the warp gates

https://user-images.githubusercontent.com/2515356/221090385-531845e2-3688-4c9e-9773-6b12dc7016b8.mp4


goal_src/jak2/engine/target/target-handler.gc
```diff
@@ -1028,12 +1028,13 @@
...
-       (-> arg3 param 3)
-       (-> arg3 param 4)
-       (go (the-as (state target) (-> arg3 param 0)))
+       (go 
+        (the-as (state object object object object target) (-> arg3 param 0))
+        (-> arg3 param 1)
+        (-> arg3 param 2)
+        (-> arg3 param 3)
+        (-> arg3 param 4)
+        )
```
dallmeyer, line 1037: thx Jak 1

https://github.com/open-goal/jak-project/blob/2d8ad39fea38a68db99fb07aca2d5866f8087602/goal_src/jak1/engine/target/target-handler.gc#L300-L308

Zedb0T:
Tested both `warp-gate` and `air-train` both seem to function now nice! 👍

Hat-Kid (changes requested):
Instead of a manual patch, change
https://github.com/open-goal/jak-project/blob/2d8ad39fea38a68db99fb07aca2d5866f8087602/decompiler/config/jak2/type_casts.jsonc#L3395

to `[866, "v1", "(state object object object object target)"],`, run the decompiler, do `task update-ref-tests` and commit the updated ref file along with the type cast change. The gsrc output should be the same as your patch so you shouldn't need to mess with it.

Hat-Kid (approved):

PR #2252, opened 2023-02-25
[jak 2] fix boot, increase level heap sizes

water111:
Increase level heaps and borrow heaps. The level heap increase was likely not needed, but better safe than sorry. We allocate the 128 MB main heap anyway so there's no harm.

Also fix the crash when using `-boot`. As I thought it was just a one-line typo in the kernel.

### DIFF
--- a/common/type_system/TypeSystem.cpp
+++ b/common/type_system/TypeSystem.cpp
@@ -648,12 +648,12 @@ MethodInfo TypeSystem::define_method(Type* type,
     int bad_arg_idx = -99;
     // make sure we aren't changing anything that isn't the return type.
     if (!existing_info.type.is_compatible_child_method(ts, type->get_name(), &bad_arg_idx) &&
-        bad_arg_idx != ts.arg_count() - 1) {
+        bad_arg_idx != (int)ts.arg_count() - 1) {
       throw_typesystem_error(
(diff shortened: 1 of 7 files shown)


Zedb0T:
Note sure why, but when I merge this into my branch it won't boot. I can just skip it its not an issue. Just documenting it incase others experience the issue too.

Steps were task extract -> gc2 -> gcnoboot

<https://github.com/open-goal/jak-project/commit/671a5658f8f4fd2519f93682b78907e5da5eb2ba> Merge Does not work

<https://github.com/open-goal/jak-project/commit/fee86413faf3e3e3424d2a64ff39bf503e2dfa1f> Revert Works again

Again not looking for diagnostics here just purely documenting in case it helps find an issue down the road.

xTVaser:
What is `gc2` and `gcnoboot` exactly?

Zedb0T:
The batch scripts on a windows 11 environment, also I just noticed I forgot to mention that I rebuilt the cpp in visual studio.

water111:
Looking at the merge, it kinda seems like that was merging a different PR?
https://github.com/open-goal/jak-project/commit/671a5658f8f4fd2519f93682b78907e5da5eb2ba

But let's try to figure this out. If you run those commands manually, which one crashes? What is printed when it happens?

PR #2267, opened 2023-02-26
[windows] make the games start up in the user's preferred system UI language

### LINKED ISSUE #1734: [runtime] Game always boots up in English
Even the Japanese version. I think.

ManDude:
Where applicable, of course.

My system language is set to English so I actually can't test this. If anyone has their Windows language (NOT LOCALE) set to Spanish, German, French, Italian or Japanese please test this.

Fixes #1734 

Also fixes the opengoal debugger on Windows and fixes the decomp for `menu` which was causing some crashes related to input handling.

### DIFF
--- a/common/cross_os_debug/xdbg.cpp
+++ b/common/cross_os_debug/xdbg.cpp
@@ -392,7 +392,7 @@ bool cont_now(const ThreadID& tid) {
 DEBUG_EVENT debugEvent;
 void ignore_debug_event() {
   if (!ContinueDebugEvent(debugEvent.dwProcessId, debugEvent.dwThreadId, DBG_CONTINUE)) {
-    win_print_last_error("ContinueDebugEvent");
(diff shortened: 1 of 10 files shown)


common/cross_os_debug/xdbg.cpp
```diff
@@ -471,14 +471,12 @@ bool check_stopped(const ThreadID& tid, SignalInfo* out) {
...
-          //   out->kind = SignalInfo::BREAK;
-          // } else {
-          //   // ignore exceptions outside goal thread
-          //   ignore_debug_event();
-          // }
+          if (exc == EXCEPTION_BREAKPOINT) {
+            out->kind = SignalInfo::BREAK;
+          } else {
+            // ignore exceptions outside goal thread
+            ignore_debug_event();
```
water111, line 478: I think the original reason to change this was because there were uncaught exceptions being thrown in the graphics thread, and this somehow caused them to be ignored, but make the game run extremely slowly.

PR #2280, opened 2023-02-26
[jak2] WIP minimap

water111:
Work in progress minimap. Known issues:
- "path finding" doesn't appear to work - it gets stuck forever in many cases
- some nasty patches around timer-based code
- jak arrow blending issues
- would be nice to make it higher resolution

if the search is forced to terminate due to iteration/time limits, the icon is not in the right place
image


water111:
One observation is that our "trail graph" looks clearly incorrect compared to pcsx2.

Just paying attention to orange here: PCSX2 has some reasonable looking graph:
image


we have this crazy stuff. And no text drawn above it:
image

I think this could also explain some strange civilian/guard behavior.

water111:
Looking at one specific bad node:
image

Some of the connections from this one are good, but some aren't. For example:
```
conn 18
  head 9
  tail 12
 (0, 1091) -> (1069, 188)
```
The x-coordinate of the "head" node is 0...
But it does seem to match the `trail-graph.gc` data for node 9.
```
(new 'static 'trail-node :next-id #x500 :z #x443 :first-conn #xb2 :cost-from-start #x1c)
```

PR #2291, opened 2023-02-28
[jak2] minor decomp cleanup + add force actors and sprite pc hacks

ManDude:
I didn't actually visually notice much of a difference with these hacks unlike in Jak 1, but I also avoided checking the missions thoroughly since the game crashes very often right now.


Hat-Kid:
Looks like you forgot to run the formatter for `var_names.json`.

PR #2298, opened 2023-02-28
fix `read_binary_file`

### LINKED ISSUE #2297: log rotation code throws an exception if the log file is empty
https://github.com/open-goal/jak-project/blob/d6bd437ea1b6ee16bb2a4d03b479fadc5af412f9/common/util/FileUtil.cpp#L281-L284

This code doesn't handle an empty file properly

It would probably be better to use the std::filesystem API to copy a file instead of reading/writing it manually. https://en.cppreference.com/w/cpp/filesystem/copy

ManDude:
Fixes #2297

### DIFF
--- a/common/util/FileUtil.cpp
+++ b/common/util/FileUtil.cpp
@@ -278,7 +278,7 @@ std::vector<uint8_t> read_binary_file(const fs::path& path) {
   std::vector<uint8_t> data;
   data.resize(len);
 
-  if (fread(data.data(), len, 1, fp) != 1) {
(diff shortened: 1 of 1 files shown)


xTVaser:
This doesn't fix the issue because it then fails on the write side.

See here https://github.com/open-goal/jak-project/pull/2299/files#diff-d95486b626ddc9c659054dc66ecd4be5f6047728d891507bea57fc75a746a700R216

ManDude:
oops.

PR #2299, opened 2023-02-28
util/file: cleanup log initialization and some file-util functions

xTVaser:
Fixes both issues mentioned in #2297

### DIFF
--- a/common/util/FileUtil.cpp
+++ b/common/util/FileUtil.cpp
@@ -213,6 +213,11 @@ void write_binary_file(const fs::path& name, const void* data, size_t size) {
     throw std::runtime_error("couldn't open file " + name.string());
   }
 
+  if (size == 0) {
+    // nothing to write, just 'touch' the file
+    return;
+  }
+
   if (fwrite(data, size, 1, fp) != 1) {
     fclose(fp);
     throw std::runtime_error("couldn't write file " + name.string());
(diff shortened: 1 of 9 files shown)


common/util/FileUtil.cpp
```diff
@@ -213,6 +213,11 @@ void write_binary_file(const fs::path& name, const void* data, size_t size) {
     throw std::runtime_error("couldn't open file " + name.string());
   }
 
+  if (size == 0) {
+    // nothing to write, just 'touch' the file
+    return;
+  }
+
```
ManDude, line 220: No `fclose`?

common/util/FileUtil.cpp
```diff
@@ -273,6 +278,9 @@ std::vector<uint8_t> read_binary_file(const fs::path& path) {
                              " cannot be opened: " + std::string(strerror(errno)));
   fseek(fp, 0, SEEK_END);
   auto len = ftell(fp);
+  if (len == 0) {
+    return {};
+  }
```
ManDude, line 283: same here?

goalc/main.cpp
```diff
@@ -79,7 +79,13 @@ int main(int argc, char** argv) {
     return 1;
   }
 
-  setup_logging();
+  try {
+    setup_logging();
```
water111, line 83: minor, but in the future I'd prefer the pattern of returning a boolean indicating success or failure, rather than throwing and requiring the user to try/catch. I know it's annoying here because a lot of the file-related functions throw (which was my original mistake).

PR #2303, opened 2023-03-09
[merc2] Support texscroll, use in jak 1 in more places, fix envmap bug

water111:
Three main changes:

- Adds support for the texture scrolling effect used on conveyor belts, and turn it on for jak 2.
- Use merc instead of generic in jak 1 for ripple/water/texscroll stuff (non-ocean water, lava, dark eco, etc). This is a pretty big speedup in a lot of places.
- Fix a really old bug with blending mode used to draw environment maps. The effect is that envmaps were half as bright as they should have been.

As usual, there's a flag to go back to the old behavior on jak 1. Set these to `#t` to use generic like we used to.
```
*texscroll-force-generic*
*ripple-force-generic*
```

The format has changed, and everything must be rebuilt (C++, FR3's, GOAL code)


ManDude:
merc doesn't render dark eco properly:
image

generic:
image

PR #2310, opened 2023-03-10
cleanup gk CLI and fix issue that caused revert

xTVaser:
My mistake -- testing focused too much on preserving the existing behaviour I clearly forgot to make sure the new stuff worked properly.

Just had to early out and not modify the args if they were in the new format.


xTVaser:
### Tested the following
Jak 2
```bash
task: [run-game] ./out/build/Release/bin/gk -v --game jak2 -- -fakeiso -debug
```

Jak 1
```bash
task: [run-game] ./out/build/Release/bin/gk -v --game jak1 -- -fakeiso -debug
```

Also works in the launcher

PR #2317, opened 2023-03-11
Add a `autosplit-flag-task-complete!` macro

Zedb0T:
Cuts out a little bit of repetition in the autosplit code. Also adds some labels to the VS configurations.

https://github.com/Zedb0T/jak-project/blob/090c4977993dd41546615d3da9dc4937d89ca243/goal_src/jak1/pc/features/autosplit-h.gc#L167-L169


Zedb0T:
Currently we take in a field-name and a task-name `(field-name task-name)`.
Even tho the task name is just `field-name' without the `res-` prefix. This way we have more control over naming future fields. 
It would be easy to refactor this to only take in a `field-name`  and enforce a standard practice if people think its better.
Also might be nice to just loop through `(defenum game-task` with a `dotimes` loop.

If anyone thinks either of those are good ideas/worth it let me know and I can adjust the PR.

.vs/launch.vs.json
```diff
@@ -135,7 +142,7 @@
       "type": "default",
       "project": "CMakeLists.txt",
       "projectTarget": "decompiler.exe (bin\\decompiler.exe)",
-      "name": "Decompiler - Jak 1",
+      "name": "Decompiler - Jak 1 - Extract",
```
ManDude, line 145: It's not an extract command.

ManDude (changes requested):

ManDude (approved):

PR #2326, opened 2023-03-17
[jak 2] ETIE

water111:
Definitely needs a clean up pass, but I think the functionality is very close.

There's a few "hacks" still:
- I am using the emerc logic for environment mapping, which doesn't care about the length of the normals. I can't figure out how the normal scaling worked in etie. I want to do a little bit more experimentation with this before merging.
- There is some part about adgifs for TIE and ETIE that I don't understand. The clearly correct behavior of TIE/ETIE is that the alpha settings from the adgif shader are overwritten by the settings from the renderer. But I can't figure out how this happens in all cases.
- Fade out is completely disabled. I think this is fine because the performance difference isn't bad. But if you are comparing screenshots with PCSX2, it will make things look a tiny bit different.


water111:
Good news on the normal length thing:
- emerc and etie logic with length 1 normals are idential! Both seem to match PCSX2.
- for etie, making the normal length smaller or larger clearly makes it different from PCSX2.
- the original etie does some trick with scaling normals that I do not fully understand. I believe they wanted to avoid a divide/sqrt operation in the VU1 code, and the effect of the trick is that the normals have length 1 in the end.

OpenGOAL
image

Issue #1650, opened 2022-07-15
[graphics] Figure out why the resolution settings don't work well on my computer


water111:
Using the menu, I can't put the game in 4k. I can go fullscreen or windowed at a lower resolution. If I force resize the window to full resolution, the internal resolution stays low.  There should be an intuitive way to go fullscreen at your monitor's resolution, and to make nearly-full-screen large windows.

water111:
was fixed by sdl2 PR

PR #2397, opened 2023-03-23
game: Migrate from GLFW to SDL2 & attempt to rewrite / simplify display and input code

### LINKED ISSUE #1488: [runtime] swap between input devices
Right now the runtime just picks the first input device it finds and then hangs on to it until it's disconnected, but a user may want to use something else that's plugged in as an input device.

### LINKED ISSUE #1712: Hide mouse pointer automatically

Hide the mouse pointer automatically when playing the game with a controller 

Using the controller should hide the mouse pointer (when pressing one of its buttons for example). Moving the mouse would enable it again.

### Additional context

_No response_

### LINKED ISSUE #1723: Alt+tabbing disconnects controller
### Acknowledgements

### Describe the Bug

I'm using a DS4 at the moment, without DS4Windows. Alt+tabbing disconnects the controller, and reconnecting requires completely turning it off (holding the PS button) and then reconnecting it. It's nothing game-breaking, but it's a hassle.

### How To Reproduce

Step 1: Use a controller
Step 2: Alt+tab while the controller is connected
Step 3: ???
Step 4: Profit.

Yes, it's unique to OpenGOAL

### Expected Behavior

The controller should stay connected while alt+tabbed, even if the game isn't receiving input while not focused, and alt+tabbing back to the game should be all it takes to restore functionality.

### Operating System

Windows 10

### OpenGOAL Version

APP Version 1.0.0

### Game Version

NTSC Greatest Hits version (red label)

Yes

### LINKED ISSUE #1785: Don't copy vmodes
We look through all vmodes from glfw and find the largest resolution, and also copy the first 128 to a place where they can be accessed from the game thread. I was lazy and made this a fixed size of 128.

Instead we should make this a vector, or just skip copying them because the game always asks for -1 anyway.

### LINKED ISSUE #1879: [runtime] investigate blocking code when window is minimized on windows

### LINKED ISSUE #1890: [linux] Runtime locks up for 30 seconds after input when a controller is plugged in with Steam running
### Acknowledgements

### Describe the Bug

on SteamOS, Game starts up immediately without loading when run through steam as a non-steam game (gk executable after having extracted the ISO with extractor at least once). As a result, the first user input provided puts the game into a glitched state that has to be killed through a task manager (I just used kill $(pgrep gk))

### How To Reproduce

Add gk executable as non-steam game (gk executable) on steam os and run it through steam (doesn't matter whether in desktop mode or game mode, still same issue). I have found that running it out of steam, after having it run through steam this way once makes it skip loading permanently as well. The only workaround I have found is to re-run extractor and then run gk (but not through steam, so it has to be in desktop mode unfortunately).

Yes, it's unique to OpenGOAL

### Expected Behavior

Waits a while before starting the game and then works as expected.

### Operating System

SteamOS

### OpenGOAL Version

1.5.0

### Game Version

NTSC 1.0 (black label)

Yes

### LINKED ISSUE #2003: add an easier way to bind keyboard and controls?
### Is your feature request related to a problem?

the issue is as for now you can't easily bind the keyboard/controller to different keys or to a mouse
(you can do it only via the controller-settings.json, but it's not documented and very hard to understand/configure for pepole who just want to play)

create an option in the launcher UI to select and bind your game keys for keyboard/mouse/controller, have an option to select modes (keyboard only, keyboard + mouse, or controller) ect

### Additional context

_No response_

### LINKED ISSUE #2124: Black screen after switching displays
### Acknowledgements

### Describe the Bug

Game stops displaying when switching to a secondary monitor. Audio continues to work and menus can still be navigated, but nothing is visual on the screen. Issue persists after restarting/uninstalling and reinstalling.

### How To Reproduce

After pressing start on the title screen, proceed to Options->Graphic options->Display. Then select Display 2.

Yes, it's unique to OpenGOAL

### Expected Behavior

When Display 2 is selected, the game should then be able to be played on a secondary monitor.

### Operating System

Windows 11

### OpenGOAL Version

Release v0.1.31

### Game Version

NTSC 1.0 (black label)

Yes

### LINKED ISSUE #2190: [pckernel] resolution query should only pick resolutions for the current display

### LINKED ISSUE #2290: Swap the bumpers with the triggers as a default

Some platformer games on the PS2 use the L1 button to crouch, and most gun games use the R1 button to shoot.

I'd like bumpers to be swapped with the triggers as a default because it feels modern.

### Additional context

_No response_

xTVaser:
Mostly putting this up so I have a diff to reference and look over what I have left to cleanup.

This PR does some major things, and I have a lot of testing to do.  The highlights are:
- SDL2 instead of GLFW for better controller support for the most part
- Since this basically meant rewriting the GLFW code, it was a good opportunity to clean the code up so it's much easier to work with and extend
- Display and input handling are abstracted behind separate managers, who handle the changes via SDL events, rather than polling every frame.  This may or may not have a performance benefit, but it definitely reduces the code complexity by a lot
- More flexible system to bind the game controls to gamepads/keyboards/mice -- but also binding arbitrary C++ functions via the same manager.
- Controller selection, options and remapping via the pause menu, I chose to break the format that controller settings were currently being saved as for a few reasons -- so I figured if I provide the user a way to set those via the UI it would be less of an issue.
- Some various cleanup in areas I was forced to touch (pc-settings / kmachine)

### Things TODO

- [x] Controller remapping #1737 
- [x] Imgui and display settings
- [x] ~~Potentially rewrite the `test_newpad.cpp`~~
  - don't really feel like doing this right now.  Though it'll probably be a lot easier now (mock an SDL event and pipe it through)
- [x] Get the tag values for imgui/sdl2 that i used

### Things to test

- [ ] Are things correct with different scaling?  I couldn't tell myself when looking.
  - We apparently don't use the scaling at all, just were tracking it incase
- [ ] DEFERRED (likely not a lot that can be done about it) The Jak font causes some strings to be printed wrong in the menu for keybindings.
- [ ] DEFERRED Remember window position/display before entering fullscreen/borderless (master does not do this either)
- [x] Make sure SDL builds when statically linked
- [x] Jak 2 still works
- [x] Test old settings.gc still loads successfully
- [x] All new settings work and are saved/restored appropriately:
  - [x] Controller options
  - [x] Mouse Options
  - [x] Enabling/Disabling Keyboard and Mouse
  - [x] Assigning binds
- [x] Controller switching
  - [x] Remembers which controller was selected, even after a device loses/regains connection
- [x] Saving/Restoring binds
  - [x] Repeatedly binding the same input will stop visually working
  - [x] Keyboard mapping can make jak move when coming out of menu
  - [x] Behaves properly when assigning to the same bind
  - [x] Bindings per unique controller device works
  - [x] Translations for unset
- [x] Restoring defaults
- [x] Mouse player/camera movement
- [x] Can switch monitors when in fullscreen/borderless
- [x] Can adjust game resolution
  - [x] Existing problem from master, the resolution list doesn't properly match the monitor's/aspect ratio
- [x] New ImGUI styling works and is saved

Fixes #1785
Fixes #1712
Fixes #1723
Fixes #2003
Fixes #2290
Fixes #1488
Fixes #2124
Fixes #1890
Fixes #2190
Fixes #1879 
 - this is apparently only a problem for me now, and a problem is a huge over-statement.  You get a slow dispatch for 1 frame when minimizing via the window (never via the windows taskbar)

Stuff was improved around #1650 but may or may not be completely fixed, so I'm not going to close it -- but atleast some of these issues have been confirmed to be resolved.


game/graphics/gfx.h
```diff
@@ -143,40 +130,49 @@ u32 Init(GameVersion version);
...
+void set_fullscreen_display(u64 display_id);
+void set_window_size(u64 width, u64 hheight);
 void get_window_scale(float* x, float* y);
-std::tuple<double, double> get_mouse_pos();
-GfxDisplayMode get_fullscreen();
-int get_screen_vmode_count();
-int get_screen_rate(s64 vmode_idx);
-int get_monitor_count();
-void get_screen_size(s64 vmode_idx, s32* w, s32* h);
+int get_connected_display_count();
```
xTVaser, line 158: test

goal_src/jak1/pc/pckernel-h.gc
```diff
@@ -426,7 +470,8 @@
   (set! (-> obj vsync?) #t)
   (set! (-> obj letterbox?) #t)
 
-  (pc-get-screen-size -1 (&-> obj width) (&-> obj height))
+  ;; TODO - why do we change the windowed mode twice
```
ManDude, line 473: `set-size!` performs a different action depending on whether we are in windowed mode or not. It's much simpler than having 2 separate functions everywhere we want to set game resolution.

VodBox:
Would the changes potentially expose the controller type/name to GOAL code? Thinking of adding handling for symbol or texture swaps when using an Xbox style controller instead of a PlayStation style controller, rather than relying on mods that change things out wholesale.

xTVaser:
@VodBox Yes it will (already does actually) however, it might be a bit fragile to depend on (basically string matching on the controller name).  A better option might be an explicit toggle for the button style people want to use, or atleast a way to override the potentially wrong automatic detection.

You are still going to need to depend on mods for most of the controller button swaps though, namely the ones that are textures, or baked into the font itself

VodBox:
@xTVaser Unfortunately the Onin game also requires new textures to be added or swapped out, since the basic button shapes while in the air are a sprite particle. That however is not relevant to this PR, so will discuss that elsewhere.

CMakeLists.txt
```diff
@@ -198,17 +196,15 @@ add_subdirectory(lsp)
...
-include_directories(third-party/SDL/include)
-
-# build glfw library
-add_subdirectory(third-party/glfw EXCLUDE_FROM_ALL)
-add_subdirectory(third-party/zstd EXCLUDE_FROM_ALL)
 
 # build imgui
 include_directories(third-party/glad/include)
-include_directories(third-party/glfw/include)
+include_directories(third-party/SDL/include)
```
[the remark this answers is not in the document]
water111: after looking a bit, I think we might be better off building it.

game/graphics/gfx.h
```diff
@@ -41,38 +41,22 @@ struct GfxRendererModule {
   std::function<void(const void*, u32)> send_chain;
   std::function<void(const u8*, int, u32)> texture_upload_now;
   std::function<void(u32, u32, u32)> texture_relocate;
+  /// NOTE - this is called _no where_ right now
+  /// it's call was commented out in `libpad.cpp` a long time ago
```
water111, line 45: feel free to remove

game/graphics/gfx.h
```diff
@@ -122,61 +106,59 @@ struct GfxGlobalSettings {
...
+std::string get_connected_display_name(int id);
+int get_active_display_refresh_rate();
+void get_active_display_size(s32* w, s32* h);
+void set_window_resizable(bool resizable);
+WindowDisplayMode get_window_display_mode();
+void set_window_display_mode(WindowDisplayMode mode);
+
+// Global Settings Related
+// TODO - how many of these actually need to be globals and instead should just be virtual functions
+// in the gfx pipeline
```
water111, line 156: Some have to be globals so they can be called from GOAL.
That said, I think there's still a lot of opportunity to clean this up to be less confusing (there's the globals, the struct of function pointers, and the virtual `Gfx` thing - I think we don't need both)

game/graphics/opengl_renderer/OpenGLRenderer.h
```diff
@@ -22,7 +22,7 @@ struct RenderOptions {
   bool draw_filters_window = false;
 
   // internal rendering settings - The OpenGLRenderer will internally use this resolution/format.
-  int msaa_samples = 4;
+  int msaa_samples = 2;
```
water111, line 25: :+1:

game/graphics/pipelines/opengl.cpp
```diff
@@ -890,7 +462,9 @@ void GLDisplay::render() {
   // switch vsync modes, if requested
   if (Gfx::g_global_settings.vsync != Gfx::g_global_settings.old_vsync) {
     Gfx::g_global_settings.old_vsync = Gfx::g_global_settings.vsync;
-    glfwSwapInterval(Gfx::g_global_settings.vsync);
+    // NOTE - -1 can be used for adaptive vsync, maybe useful for Jak 2+?
```
water111, line 465: yeah, that would be cool. Right now jak 2 is hard-coded for 60 fps.

game/graphics/pipelines/opengl.cpp
```diff
@@ -1014,7 +588,8 @@ void gl_texture_relocate(u32 destination, u32 source, u32 format) {
 }
 
 void gl_poll_events() {
-  glfwPollEvents();
+  // TODO - should we process the SDL events here
+  // does this run before `render`?
```
water111, line 592: I would remove `poll_events` from everywhere.

game/kernel/common/kmachine.cpp
```diff
@@ -348,25 +350,91 @@ void DecodeTime(u32 ptr) {
...
 
-/*!
- * Do a fast memory copy.
- */
+/// <summary>
+/// Do a fast memory copy.
+/// </summary>
+/// <param name="dst"></param>
+/// <param name="src"></param>
+/// <param name="size"></param>
```
water111, line 382: I'd rather stick with the Doxygen-style (`/*! ...`) comments if possible. I've only seen this `<summary> </summary>` style used in C#.

game/runtime.cpp
```diff
@@ -357,9 +357,9 @@ RuntimeExitStatus exec_runtime(GameLaunchOptions game_options, int argc, const c
     try {
       Gfx::Loop([]() { return MasterExit == RuntimeExitStatus::RUNNING; });
     } catch (std::exception& e) {
-      fmt::print("Exception thrown from graphics loop: {}\n", e.what());
-      fmt::print("Everything will crash now. good luck\n");
-      throw;
+      lg::error("Exception thrown from graphics loop: {}\n", e.what());
+      lg::error("Everything will crash now. good luck\n");
+      throw e;
```
water111, line 362: this should be just `throw;`, not `throw e;`
(see https://stackoverflow.com/questions/1095225/exception-slicing-is-this-due-to-generated-copy-constructor)

If you do `throw e`, it will copy the exception using the copy-constructor for `std::exception` (bad, causing slicing). But `throw` will rethrow the exception without copying.

game/system/hid/input_manager.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+bool GameController::has_led() {
+  // TODO
+  return false;
+}
+
+void GameController::set_led(const u8 red, const u8 green, const u8 blue) {
+  // TODO - handle return
+  SDL_GameControllerSetLED(m_device_handle, red, green, blue);
```
water111, line 129: it would be very cool to have a blue light come on when you pick up blue eco.

game/system/hid/input_manager.h
```diff
@@ -0,0 +1,171 @@
...
+  /// <summary>
+  /// Collection of arbitrary commands to run on user actions
+  /// </summary>
+  CommandBindingGroups m_command_binds;
+
+  bool m_keyboard_enabled = true;
+  bool m_mouse_enabled = false;
+
+  bool m_ignore_background_controller_events = false;
+};
```
water111, line 171: I like the design of the input manager/display manager and how they handle sdl events.

game/system/hid/sdl_util.cpp
```diff
@@ -0,0 +1,23 @@
...
+bool is_any_event_type(uint32_t event_type, std::vector<uint32_t> allowed_types) {
+  for (const auto& allowed_type : allowed_types) {
+    if (allowed_type == event_type) {
+      return true;
+    }
+  }
+  return false;
+}
+SDL_bool sdl_bool(const bool val) {
+  return val ? SDL_TRUE : SDL_FALSE;
```
water111, line 21: why does SDL need to do this.......

water111 (commented):
Sorry this took me so long to get to.

Overall I think this switch to SDL is a good idea. The original reason to avoid it was because GLFW is a lot simpler, and I hoped would have less weird problems. But it didn't seem to be the case in the end.

I only had minor comments. Thanks for doing a bunch of cleanup as part of this.

CMakeLists.txt
```diff
@@ -38,13 +38,12 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
         -Wall \
         -Wno-c++11-narrowing \
         -Wno-c++98-compat \
-        -O3")
+        -O3 -stdlib=libc++")
```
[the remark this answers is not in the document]
xTVaser: no this shouldnt be required iirc, I thought i removed this.

game/system/hid/sdl_util.cpp
xTVaser: from reading, it's because it's fundamentally a C library which doesn't have a proper boolean type

CMakeLists.txt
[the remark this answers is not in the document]
water111: oops, yeah this comment was from a few days ago.

CMakeLists.txt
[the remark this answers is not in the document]
water111: let's leave it like you have it for now and see how things go

game/system/hid/sdl_util.cpp
water111: yeah, this isn't an uncommon thing in C code... it just drives me crazy that people feel the need to make up stuff like `SDL_TRUE`, as if the concept of TRUE is somehow specific to SDL and not programming in general.

C has has `#include <stdbool.h>` to define `true`/`false` since c99, but I guess SDL has to be compatible with truly cursed platforms where `stdbool.h` doesn't exist, and it's not possible to define`true` as `1` for some crazy reason.

CMakeLists.txt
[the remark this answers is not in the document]
xTVaser: for sure, my understanding for why they encourage dynamic linking is so that a newer version of SDL can be used on an old game.  Steam apparently provides their own SDL, so if your game hasn't received an update in a million years, it can still use the latest SDL.  Atleast that's one of the benefits, which isn't really an issue for us.

Manuel-K:
Sorry for yelling from the sideline, but I wouldn't add SDL's code to the repo. Using git submodule would severely reduce bloat.

xTVaser:
We started with submodules for vendoring our dependencies, they cause more problems than they solve.  Either way you have to clone the repo (via a submodule or directly) so there is practically no reduction in bloat.

Manuel-K:
It's not part of the history forever. That's the main issue with git and external data. And since there should at least be an option to dynamically link SDL you'd typically use the system version on Linux, so checking the submodules out would not be required all the time. Another disadvantage with some library as big as SDL in the main tree is that it makes reviewing changes harder.

But submodules certainly have downsides, there's no argument against that.

xTVaser:
It makes reviewing changes on the initial PR that adds the dependency harder.  After that you basically never see it again.

Our project is entirely statically linked when distributed, and our CMake is not setup to support using system packages, if a reason to dynamically link becomes obvious in the future either we can adjust things, or the open-source community can fork and make adjustments accordingly.

water111:
I spent a while testing jak 1 and jak 2, and overall it seems great. I didn't see any suspicious frame time spikes like I did with glfw, and my controller worked perfectly. Building on linux was good too - it's a little slower on the initial cmake generation, but I don't think this is a big deal.


A few things I found so far:
- on start, it looks like the initialize resize of the window is missed:
image
- the startup is slower by about 1s. We've put a lot of effort into making the startup fast, so if possible it would be great to get closer to where we were before. For comparison, loading from disk, linking, and initializing the entire engine takes about 0.7s.
- on linux, the "fullscreen" mode just moves the window up toward the top corner of my screen. Once in this mode, the only resolutions I can pick are 512 x 224 and 512 x 448. Once selecting one of those sizes, there is no way to go back to the resolution of the window.
- borderless works great. This is a great improvement and the first time any sort of fullscreen has shown up at the right resolution.
- Setting the resolution to my monitor's resolution (4k), then disabling MSAA makes everything dark, as if the "shadow" is applied to the whole screen. This also happens in windowed mode
- Selecting "windowed mode" when already in windowed mode resizes the window
- Vibration feels different (linux. ds4). I can't really say which is better/worse, so I don't think it really matters.

startup comparison
https://user-images.githubusercontent.com/48171810/232352217-c84fdbd8-af50-487c-9021-2154eebeb911.mp4

xTVaser:
> on linux, the "fullscreen" mode just moves the window up toward the top corner of my screen. Once in this mode, the only resolutions I can pick are 512 x 224 and 512 x 448. Once selecting one of those sizes, there is no way to go back to the resolution of the window.

I've seen this issue as well when it wasn't able to properly get the width/height of the monitor.  Might be useful to see what is going on around here https://github.com/xTVaser/jak-project/blob/v/sdl2/game/kernel/common/kmachine.cpp#L524 as well as in logs when fullscreening for any errors https://github.com/xTVaser/jak-project/blob/v/sdl2/game/system/hid/display_manager.cpp#L153 the fact that borderless works fine for you by comparison is....unexpected.

> Setting the resolution to my monitor's resolution (4k), then disabling MSAA makes everything dark, as if the "shadow" is applied to the whole screen. This also happens in windowed mode

This is out of my element, do you have any ideas on this one?  It sounds similar to when i tried to run the game with HDR enabled, but that was apparently a well-known issue.

> on start, it looks like the initialize resize of the window is missed

I assume the initial resize is from i3 automatically tiling the window?  I bet this is due to the order in which things are created in our current code -- there seems to be a race between the window being created, and the events associated with that window starting to be processed (not until `GLDisplay::GLDisplay` is called).  I could probably do a manual call on initialization to fix this.

The startup time being so different is...surprising to me. There is way less overhead in general (event processing per frame is single-digit nanosecond) and the startup is fairly generic/boilerplatey.  I'll add some more profiling metrics around the startup to see if we can narrow it down.

Hat-Kid:
> @Hat-Kid do you still have the 30second hang issue with this?

Yep, it's gone. I do still get a small hitch when reconnecting a controller that felt slightly worse with Steam open, but it's less than a second.

xTVaser:
Nice, I also get a small hitch that I need to look into after this PR merges (existing issue) -- for whatever reason the kernel dispatch time when doing so is quite slow for a few frames, and I don't immediately believe polling for the controllers is to blame.  But i have to profile it.

xTVaser:
I added a `--profile` flag which will turn the profiler on immediately, I'm interested to see how this compares on your environment.

The below are my results, both on debug builds.  Branch for GLFW timings - https://github.com/xTVaser/jak-project/tree/glfw-profiling

Seems like a lot of the time is spent on GLAD initializing, and the window creating.  GLFW is slightly faster to create the window, but then the rest of the initialization code takes longer.

image

image

So for me, the startup time seems to within margin of error, and it makes sense that SDL would take a little longer on the first init, as it's starting up all the relevant APIs at the same time.  GLAD seems to be the real opportunity here.

---

Also since im already doing this, a comparison of the SDL2 per-frame usage vs GLFW.  Hundreds of micro-seconds vs 10+ milliseconds

image

image

game/main.cpp
```diff
@@ -14,6 +14,7 @@
 #include "common/util/os.h"
 #include "common/util/unicode_util.h"
 #include "common/versions.h"
+#include <common/global_profiler/GlobalProfiler.h>
```
ManDude, line 17: "" instead of <>

game/graphics/opengl_renderer/debug_gui.cpp
```diff
@@ -3,6 +3,8 @@
 
 #include <algorithm>
 
+#include <common/global_profiler/GlobalProfiler.h>
```
ManDude, line 6: "" instead of <>

ManDude (changes requested):

xTVaser:
Correction on the profiling, noticed that the GLAD measurement was also capturing our creation of `GraphicsData`, which is actually the culprit.  Not sure which of the 4 things that are created by `GraphicsData` is causing all the time -- it's hard to profile it without making a bunch of annoying changes since it's all done in the initializer list.

I was also missing the SDL initialization in `gl_init`.  The following is a much more comprehensive overview of the startup sequence.

image

xTVaser:
Hoping we can get this merged this week.  The issues I'm concerned about looking into / resolving before that happens are the following:
- In rare situations (or apparently consistently when water tries to go fullscreen) the resolution menu either has very limited options or none at all.  Consistent repro steps for this should help figure out what's going on

And the following as well, but they require debugging from a linux environment:
- MSAA off causes everything to be dark, similar to the issue I had but "fixed" it by increasing the bit depth.
- Why isn't "fullscreen" working on linux but borderless is.
- Why isn't the initial window size set properly, is there an additional event i have to listen to?

ManDude:
Biggest issues are:
- switching from fullscreen to borderless breaks things. When you go back to windowed the display will be set to the wrong size.
- switching from borderless to windowed sometimes puts the window at the wrong size.

ManDude:
> I removed the 512x224 / 512x448 resolution options, revert?

Keep these because why not, but make them borderless-only since exclusive fullscreen might not actually allow those properly.

ManDude (approved):

game/graphics/pipelines/opengl.cpp
```diff
@@ -233,200 +148,163 @@ static std::shared_ptr<GfxDisplay> gl_make_display(int width,
   g_gfx_data->imgui_filename = file_util::get_file_path({"imgui.ini"});
   g_gfx_data->imgui_log_filename = file_util::get_file_path({"imgui_log.txt"});
   ImGuiIO& io = ImGui::GetIO();
+  io.ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange;  // We manage the mouse cursor!
+  if (!Gfx::g_debug_settings.monospaced_font) {
+    std::string font_path =
+        (file_util::get_jak_project_dir() / "game" / "assets" / "fonts" / "Roboto-Medium.ttf")
+            .string();
+    if (file_util::file_exists(font_path)) {
+      io.Fonts->AddFontFromFileTTF(font_path.c_str(), Gfx::g_debug_settings.imgui_font_size);
```
ManDude, line 157: Can you add the following before this line?
```cpp

    static const ImWchar ranges[] = {
        0x0020, 0x00FF,  // Basic Latin + Latin Supplement
        0x2000, 0x206F,  // General Punctuation
        0x3000, 0x30FF,  // CJK Symbols and Punctuations, Hiragana, Katakana
        0x3131, 0x3163,  // Korean alphabets
        0x31F0, 0x31FF,  // Katakana Phonetic Extensions
        0xAC00, 0xD7A3,  // Korean characters
        0xFF00, 0xFFEF,  // Half-width characters
        0xFFFD, 0xFFFD,  // Invalid
        0x4e00, 0x9FAF,  // CJK Ideograms
        0,
    };
```
This will make it so more than just the latin alphanumeric characters get loaded.

xTVaser: Surely `ranges` has to be used somewhere, right?

Also `Wchar` makes me nervous -- is this a windows only thing / what happens on linux?

ManDude: Oops yeah I forgot to tell you where it's used. You pass it to the `AddFontFromFileTTF` call, it's the fourth argument I believe.

ManDude: As for the type... these are just unicode code points, so they normally cap out at `0xFFFF`. I don't even know if there's a way to get imgui to render characters above that code point.

PR #2406, opened 2023-03-25
[jak2] fix flickering and depth writes

water111:
Fixes some mistakes with merc draw modes. The glass in the palace level no longer writes to the depth buffer (it's "water"):
image

Also fixes the one-frame flickers when level draw orders change. We might be able to make this more efficient in the future, but this will at least fix the frame with nothing drawn.


water111:
Also, fix a mistake in texture offsets that was causing green eyes here:
image

PR #2412, opened 2023-03-25
[jak2] Fix texture page dir offset for jak 2

water111:
Accidentally used Jak 1's tpage dir size...


water111:
Fixes the invisible sign issues:
image
(appears, but is rotated 90 degrees)
image

PR #2424, opened 2023-03-26
[jak2] a few small graphics fixes

water111:
- better handling of the `disable-fog` settings for merc, should fix the spotlights. There's a setting in the merc effect, and also a runtime flag for the draw-control. I'm not actually sure what reads these, but the draw-control one is definitely used to disable fog on the spotlights.
- increase merc draw limit to try to fix the issue about partially drawn citizens in the city
- remove useless debug prints (it's okay to die in init, and the medium load buffer size mode is understood now)


water111:
Fixed trees in farm. They were accidentally using the TIE + wind stuff, which is not set up at all in jak 2 yet:
image

water111:
Spotlights fixed by the fog changes:
image

PR #2434, opened 2023-03-27
[jak2] add pal decomp support

VodBox:
Unsure what hacks or type casts might eventually be needed for the PAL version, but these changes are enough to get things to extract and run the game (though the extract phase requires iso_data to be in a jak2_pal folder, but the build still looks for jak2 in iso_data and decompiler_out).


decompiler/config/jak2/pal/label_types.jsonc
xTVaser, line 1: does the PAL version actually need this much?  The size of this file is suspiciously close to the ntsc file, you should only have to add the files that differ (i havn't run a diff myself)

VodBox: They do not differ, but while shorter, the Jak 1 pal file of this contains a lot of duplicates from the Jak 1 ntsc_v1, and I was unsure what should be carried over or what can be removed. With that uncertainty, copying felt safest.

xTVaser: The files are merged at the top level, if there are any true complete duplicates (where the contents of the key is the same for both) then that was just a mistake from having to do it retroactively.  So it would be nice to start out properly this time with only the minimal changes that need to be made.

VodBox: Pushed a smaller commit

xTVaser (approved):
Kinda surprised it requires no changes at all, I thought I remember someone having issues decompiling it with a PAL ISO because some of the files were different.  In any case I differ to you since you have an ISO and I do not.

VodBox:
> Kinda surprised it requires no changes at all, I thought I remember someone having issues decompiling it with a PAL ISO because some of the files were different. In any case I differ to you since you have an ISO and I do not.

It might be different when it comes to running the full decompiler, but just for the purposes of extracting assets and building the game, this is enough, likely because it's using the already decompiled goal_src.

xTVaser:
add to whitelist

PR #2494, opened 2023-04-16
Making some room for custom level

himham-jak:
OpenMaya will soon export new progress screens, load boundaries, and missions to custom levels. I figured having somewhere standardized to export these things to would be good.

### DIFF
--- a/game/assets/jak1/text/game_text_en.gs
+++ b/game/assets/jak1/text/game_text_en.gs
@@ -437,6 +437,8 @@
 (#x1524 "ALL CUTSCENES"
         "ALL CUTSCENES")
 
+;; Custom Level Text: 1600-1800
+
 ;; -----------------
 ;; test (DO NOT TRANSLATE)
 
--- a/goal_src/jak1/engine/game/task/game-task-h.gc
+++ b/goal_src/jak1/engine/game/task/game-task-h.gc
@@ -127,5 +127,10 @@
   (plunger-lurker-hit 113)
   (leaving-misty 114)
   (assistant-village3 115)
-  (max 116)
+  ;;Do not remove dummy
+  (dummy 116)
+
+  ;; Custom Missions: 150-224
+
+  (max 225)
   )
--- a/goal_src/jak1/engine/game/task/task-control.gc
+++ b/goal_src/jak1/engine/game/task/task-control.gc
@@ -4090,7 +4090,7 @@
 (defun get-task-control ((arg0 game-task))
   "Get the task control for a given game-task"
   (cond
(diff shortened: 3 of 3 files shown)


game/assets/jak1/text/game_text_en.gs
```diff
@@ -437,6 +437,8 @@
 (#x1524 "ALL CUTSCENES"
         "ALL CUTSCENES")
 
+;; Custom Level Text: 1600-1800
+
```
xTVaser, line 441: this is a fairly fragile thing to depend on. for example this PR adds things in the `1600` range https://github.com/open-goal/jak-project/pull/2397 

If you want to reserve a range, you should actually add the entries to the enum so we don't touch them.  I struggle to see how this will work end-to-end consistency but that's a separate issue.

goal_src/jak1/engine/game/task/game-task-h.gc
```diff
@@ -127,5 +127,10 @@
   (plunger-lurker-hit 113)
   (leaving-misty 114)
   (assistant-village3 115)
-  (max 116)
+  ;;Do not remove dummy
+  (dummy 116)
```
xTVaser, line 131: why is this entry needed, is it used somewhere?

xTVaser (commented):
the errors are probably because you changed the `game-task` enum but didn't update it in `all-types.gc`

goal_src/jak1/engine/game/task/game-task-h.gc
himham-jak: according to Hat Kid, removing 116 breaks the enum. If this is not the case, then it can be removed

goal_src/jak1/engine/game/task/game-task-h.gc
```diff
@@ -127,5 +127,10 @@
...
+
+  ;; Custom Missions: 150-224
+
+  (max 225)
```
ManDude, line 135: Why 225 specifically?

game/assets/jak1/text/game_text_en.gs
himham-jak: I'm not trying to add things here manually. Blender will be doing it automatically. I can't add them ahead of time unless they had obnoxious placeholders which would be... worse than nothing

xTVaser: You can add an entry to the actual enum at 1600 and 1800 so people know not to add something in that range.  I think you still have a ton of issues to solve related to that but atleast then when we go to add something in `text-h.gc` it's very obvious we should avoid that range.

I'm just saying a comment in the english text file isn't going to guarantee anything for you.

xTVaser:
This PR seems abandoned, feel free to re-open if work resumes

PR #2526, opened 2023-04-18
decompiler: some hacks to allow running decompiler on jak 3 v5 code files, improve `all-types` generation

Hat-Kid:


goal_src/jak2/engine/gfx/hw/display.gc
```diff
@@ -34,6 +34,9 @@
     (('pal)
      (set! (-> obj time-factor) 6.0)
      )
+    (('ntsc)
+     (set! (-> obj time-factor) 1.81818182) ;; (300 / 165fps)
+     )
```
ManDude, line 39: Oops?

ManDude (changes requested):

goal_src/jak2/engine/gfx/hw/display.gc
Hat-Kid: Whoops lol

water111 (approved):

Hat-Kid:
With the latest push, the `all-types` generation for Jak 3 has been vastly improved.

It should now pass most types without failing miserably except for two edge cases, `level-group`, which is so big that it stores its offsets for the load operations in separate registers and `load-dir` which is doing stack spill shenanigans. If it fails to read a field in the middle of an inspect, it will now just generate a dummy field instead of completely giving up on the type.

For far label types, it will now get the fields, but will still fail to get the flags and, sometimes, the parent type, which is still a vast improvement over nothing.

goal_src/jak3/build/all_objs.json
```diff
@@ -0,0 +1,2764 @@
...
+["temple-part",                                     "temple-part",                                     5, ["TEMA"], ""],
+["temple-obs",                                      "temple-obs",                                      5, ["TEMA"], ""],
+["temple-scenes",                                   "temple-scenes",                                   5, ["TEMA"], ""],
+["temple-obs2",                                     "temple-obs2",                                     5, ["TEMA"], ""],
+["temple-mood",                                     "temple-mood",                                     5, ["TEMA"], ""],
+["templea-texture",                                 "templea-texture",                                 5, ["TEMA"], ""],
+["hover-training",                                  "hover-training",                                  5, ["TEMA"], ""],
+["tomb-baby-spider-TEMA",                           "tomb-baby-spider",                                5, ["TEMA"], ""],
+["tomb-baby-spider-TEMA",                           "tomb-baby-spider",                                5, ["TEMA"], ""],
```
ManDude, line 579: We still need to do something about this - can you detect if the type of a data-only object file is a child of (or exactly) `art-group`?

ManDude (dismissed):

PR #2528, opened 2023-04-18
[jak2] bring other version fixes to source

### LINKED ISSUE #2530: If you do the reverse races early, you get softlocked out of the defend stadium mission
This happens on the black label version of the game, however it is patched on the greatest hits version

ManDude:
- [x] compare NTSC-K
- [x] compare NTSC-J
- [x] compare PAL
- [x] figure out version order
- [x] ~~write delta patch for spanish text~~ no need for now

Fixes #2530


decompiler/config/jak2/jak2_config.jsonc
```diff
@@ -139,6 +139,15 @@
...
+    },
+    "ntsc_v2": {
+      "game_name": "jak2_us2",
+      "type_casts_merge_file": "decompiler/config/jak2/pal/type_casts.jsonc",
+      "label_types_merge_file": "decompiler/config/jak2/pal/label_types.jsonc",
+      "hacks_merge_file": "decompiler/config/jak2/pal/hacks.jsonc",
+      "anonymous_function_types_merge_file": "decompiler/config/jak2/pal/anonymous_function_types.jsonc",
+      "obj_file_name_map_file": "goal_src/jak2/build/all_objs.json",
+      "object_patches": {}
     }
```
xTVaser, line 151: so i was mistaken, and original PAL doesn't have the fix.  Apparently someone reproduced the problem on original PAL (which makes sense, unlike Jak 1, Jak 2 released in EU mere days later).

You mentioned this is the only difference across all versions, and it isn't PAL specific.  If it's shared for both ntsc_v2 and PAL_v2, perhaps it should go in a better named folder like `ntsc_pal_v2`.

ManDude: Ah ok. I read at first that PAL had the fix so I just assumed.

Either way I'll be comparing all 5 version now.

Issue #2531, opened 2023-04-21
mac support


Helias:
### What feature is your idea related to?

I was trying to build jak-project on Mac, but I am encountering an issue that I cannot solve

```
jak-project/./third-party/cubeb/cubeb/include/cubeb/cubeb.h:10:10: fatal error: 'cubeb_export.h' file not found
#include "cubeb_export.h"
         ^~~~~~~~~~~~~~~~
1 error generated.
```

bypassing this issue I reached the 98% of the compiling build, we should be close to support (non-M1) Mac build I hope.

### Describe the solution you'd like.

successful build

### Additional context

clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin21.6.0

PR #2541, opened 2023-04-24
feat: mac support

Helias:
still working on it.

I opened this PR to let any other devs to see the progress and help me with this issue https://github.com/open-goal/jak-project/issues/2531


xTVaser:
- The debugger needs to be implemented for MacOS, so it makes sense you're seeing so many undefined symbols, right now it's either linux or windows https://github.com/open-goal/jak-project/blob/d84859e43160b3b06199934cccd5e97eda38234b/common/cross_os_debug/xdbg.cpp#L36 hopefully the linux implementation is fairly close.
- There are a few more things that you will have to do before getting something to actually work on macOS.  Namely OpenGL 4.3, as well as some file-path handling stuff.  Just as a heads up.

Helias:
Thanks for your answer.
I am not really keen on solving this incompatibility but I could give it a try.

So, do you know how can I solve that issue or similar ones? Do you have more tips, guides or hints for the next steps?

Thank you in advance.

(Thanks all, I check monthly the report and sporadically the PRs and commit history)

water111:
Some specific issues you might hit:
- For the debugger: this is tricky to implement, and you can probably just skip it. It's only needed to debug OpenGOAL code itself, and it's not the most useful tool. You can build and run the game without the debugger.
- For graphics: this is probably the biggest project. Currently we require OpenGL 4.3. This is over 10 years old, yet Apple doesn't support it. All of the 4.3-only features we use aren't really required, and with some editing of the renderers I think you could get things running on an older version. But it would require some OpenGL experience, or some time to learn about OpenGL.
- For everything else: for the most part, the linux implementation should work. A decent portion of our tools actually work on mac already.

Helias:
thank you both, from my understanding, it's not a task that I can do with my knowledge, I think that I will make another try and in case of failure I will abandon this closing issue and PR, I will still follow the project and maybe try again another time in the future.

Helias:
I saw that in someway Mac can support OpenGL 4.3, I do not know exactly how and if it's easy to do but I could try to make it work.

My main issue for now is this error compiling the source with `cmake --build build -j 8`

```
/Users/4009093/projects/OPENGOAL/jak-project/./third-party/cubeb/cubeb/include/cubeb/cubeb.h:10:10: fatal error: 'cubeb_export.h' file not found
#include "cubeb_export.h"
         ^~~~~~~~~~~~~~~~
1 error generated.
```

do you have any idea?

PR #2578, opened 2023-04-27
jak2: fix drill platform excitement being maxed everywhere

Hat-Kid:
The Ginsu that ambush you had invalid bones before getting out of the ambush state, so Jak was trying to look at a NaN vector, maxing out the excitement value.

Fixes #2577

PR #2580, opened 2023-04-29
fix many instances of bad bone data being used

### LINKED ISSUE #2577: Excitement level is maxed out almost everywhere

### LINKED ISSUE #2579: sneak-in -- cutscene plays before it should
Adding additional Early cutscenes here Zed -

atoll-2-intro sig after red gun course (should play when passing the carpet texture)
sewer-1-intro

ManDude:
Fixes cutscenes starting way too early and various cutscene softlocks.

Fixes #2579
Fixes #2577 

Supersedes #2578


goal_src/jak2/engine/ai/enemy.gc
```diff
@@ -157,7 +157,8 @@
   "@returns the `trans` [[vector]] from the process's `root` (typically either a [[trsqv]] or a [[collide-shape]])"
   (let ((s4-0 (-> obj root-override2)))
     (cond
-      ((zero? arg0)
+      ;; PC PORT NOTE : added lod-set check so we don't use invalid bones
+      ((or (not (logtest? (-> obj draw status) (draw-control-status lod-set))) (zero? arg0))
```
water111, line 161: I'm a little afraid of this change because it potentially changes a bunch of stuff for things that aren't being drawn. For example, enemies that aren't being drawn and call this function with `arg0` set to 1 used to get `gspot-pos`, and now they will get `trans`.

ManDude:
result of `get-trans` with two different parameters on the first frame of `ashelin-npc-1`:
```
0: [0024ee30] vector
        [ 546695.3750] [1414090.7500] [2204495.7500] [      1.0000]

2: [0024ef00] vector
        [         NaN] [         NaN] [         NaN] [   7372.7998]
```

goal_src/jak2/engine/ai/enemy.gc
ManDude: Alright, changed it so the hack is a lot less aggressive and only patches when it would actually give a bad result. I don't think it's even necessary in `enemy`.

goal_src/jak2/engine/draw/drawable.gc
```diff
@@ -1205,9 +1205,10 @@
                     ;; PC port note : we ALWAYS disable the envmap hack when a process-drawable has warp effect enabled
                     (when (or (= lod-idx (-> dc cur-lod)) (logtest? (-> dc status) (draw-control-status lod-set)))
                         (protect ((-> *pc-settings* force-envmap?))
-                          (dotimes (eff-i (-> dc mgeo header effect-count))
-                            (if (logtest? (effect-bits cross-fade) (-> dc mgeo effect eff-i effect-bits))
-                              (false! (-> *pc-settings* force-envmap?))))
+                          (when (not (movie?))
```
water111, line 1208: is this for the baron throne room cutscene, when you can see the metal head hologram from far away?

goal_src/jak2/engine/process-drawable/process-focusable.gc
```diff
@@ -76,7 +76,10 @@
                       (-> (the-as collide-shape-moving gp-0) gspot-pos)
                       )
                      ((and (or (= arg0 2) (= arg0 3)) (type? gp-0 collide-shape))
-                      (-> gp-0 root-prim prim-core)
+                      ;; PC PORT NOTE : added lod-set check so we don't use invalid bones
+                      (if (not (logtest? (-> obj draw status) (draw-control-status lod-set)))
+                          (-> gp-0 trans)
+                          (-> gp-0 root-prim prim-core))
```
water111, line 82: I'm still not totally convinced this is right... I think there could be another bug somewhere else because I would never expect the `(-> gp-0 root-prim prim-core trans)` to be NaN. I guess let's merge this for now and I will look into it later: https://github.com/open-goal/jak-project/issues/2581 if we find a better fix maybe we can remove this change.

goal_src/jak2/engine/draw/drawable.gc
ManDude: yes

PR #2608, opened 2023-05-01
game: don't exit prematurely when rebooting in debug

xTVaser:
I'm not sure what has changed here...but rebooting in debug mode now hangs.

- It seems to exit the GFX loop prematurely because `MasterExit` is still set to `RESTART_IN_DEBUG` https://github.com/open-goal/jak-project/blob/master/game/runtime.cpp#L431
- And then it gets stuck waiting for the DECI thread to close

This works, but im not sure if it's the right fix / what has changed to require this.

### DIFF
--- a/game/main.cpp
+++ b/game/main.cpp
@@ -192,6 +192,7 @@ int main(int argc, char** argv) {
     // run the runtime in a loop so we can reset the game and have it restart cleanly
     lg::info("OpenGOAL Runtime {}.{}", versions::GOAL_VERSION_MAJOR, versions::GOAL_VERSION_MINOR);
     try {
+      MasterExit = RuntimeExitStatus::RUNNING;
       auto exit_status = exec_runtime(game_options, arg_ptrs.size(), arg_ptrs.data());
       switch (exit_status) {
         case RuntimeExitStatus::EXIT:


game/main.cpp
```diff
@@ -201,6 +201,7 @@ int main(int argc, char** argv) {
           break;
         case RuntimeExitStatus::RESTART_IN_DEBUG:
           force_debug_next_time = true;
+          MasterExit = RuntimeExitStatus::RUNNING;
```
water111, line 204: can you try putting this on line 194, right before the `exec_runtime` call? I think that's the right place to put it.

I think this is potentially a race condition on startup that shows up because I disabled the VM stuff, which makes graphics stuff start up about 200ms faster now, so the graphics thread beats the EE thread in setting this flag. The right thing to do is set MasterExit to running before we start anything.

water111 (commented):
Looks good, thanks for debugging.

game/main.cpp
xTVaser: Line moved.  Ah I saw the VM changes, but I was thinking it was unrelated.

water111: yeah I didn't expect it to make a difference either - I think this was always buggy and we just got really lucky with the timing before.

PR #2620, opened 2023-05-05
fix blit size being wrong most of the time

ManDude:


ManDude:
The blit UVs were wrong whenever the visible area of the game did not match the visible area of the game on the window (e.g. borderless at a custom resolution)

PR #2629, opened 2023-05-06
Support multiline text (e.g. credits) override with JSON format

dallmeyer:
Ported over the logic from parsing credits in GOAL game text file over to the JSON parser.

minor notes:
- didn't bother replicating the logic for skipping empty lines
- didn't need to handle multiple languages in one file

https://github.com/open-goal/jak-project/blob/40b2e93be7ec24a597017e04b117a58162840016/common/serialization/subtitles/subtitles_ser.cpp#L92-L142

Using the sample multiline credits override from my comment in the code:
image


common/serialization/subtitles/subtitles_ser.cpp
```diff
@@ -189,6 +189,34 @@ void parse_text(const goos::Object& data, GameTextDB& db, const GameTextDefiniti
...
+ *      ]
+ *    }
+ * 
+ * The "begin" entry is the ID (x0b00) for the first line of the credits,
+ *  which you can get/verify from the extracted assets in
+ *  jak-project/decompiler_out/jak1/assets/game_text.txt
+ * 
+ * Note that this overwrites the original credits line-by-line, i.e. if
+ *  you provide only one entry in "lines", that will overwrite the first
+ *  line of the credits and everything after will remain unchanged.
```
xTVaser, line 218: ah i see, eventually it might be nice to extend it to you can add additional overrides (if you need to change a line in the middle of the credits for example) as i imagine most translation efforts would need to do such a thing.

dallmeyer: I think the challenge for that might be lining things up with the original credits, the extracted assets just spits out the hex value for the ID of the first line. Contributors would need to figure out the offset from that to know which line to override. It wouldn't be hard to support here though

dallmeyer: I guess another option could be to let users provide a `null` row to indicate that there are no changes to that row - so you could have 10 lines of nulls and those lines aren't changed and then provide an override for the 11th line. A bit ugly, but then you don't need to duplicate any text you're not changing

xTVaser: im imagining that yeah the user does the painful task of counting lines so they can explicit say "replace starting at X with these lines".  That isnt a great solution either, but if you had to change something at line...200, its better than 200 nulls (and in both cases you have to know how many lines anyway)

dallmeyer: In that case, does this even need to special-cased to the credits? What if we just let you provide either a single string or an array of strings as the value for a text ID, and we update line by line starting from that ID, incrementing each time? I don't know if there are any cases outside credits where this is really that useful, but might simplify the code a bit?

xTVaser: well for the normal files via Crowdin, it has to be a simple 1-1 mapped format.  But yeah if that makes things easier for this stuff, it doesnt sound too outlandish.  Though I'm not sure if the `id` (for example `b00`) is actually a text-id and I'd be worried about things overlapping if that isn't the case.

tldr: If it makes the json format less verbose and accomplishes everything required -- i'd say it'd be an improvement.

dallmeyer: > Though I'm not sure if the id (for example b00) is actually a text-id and I'd be worried about things overlapping if that isn't the case.

There's no entry in the `text-id` GOAL enum for the credit lines, but it looks like we update them the same when parsing the game_text files.

ManDude: `#xb00` comes from `credits.gc`

dallmeyer: >_< of course, how could I overlook `(gp-0 2815)`

xTVaser (approved):
I think this is fine to merge assuming it all works?  If during use the format has to be tweaked a bit in hindsight that's fine as the credits stuff won't ever go in the files that Crowdin targets anyway.

PR #2507, opened 2023-04-17
jak2: shadow renderer and blackout

Hat-Kid:
Still a WIP, shadows are a bit broken, use the Jak 1 VU1 code (Jak 2 VU1 breaks them even more) and there are some ugly hacks in the renderer

PR #2632, opened 2023-05-06
[jak2] combined shadow work PR

water111:
This PR is a combination of https://github.com/open-goal/jak-project/pull/2507 and some additional changes to port Shadow VU1 to OpenGL.  As far as I can tell, it's working.


water111:
The fortress tank shadow sometimes looks wrong:
image
which is caused by the shadow volume being too small:
image

I think there's two possibilities:
- the shadow volume should be longer
- the end cap of the volume should be visible (this would be confusing - the end cap definitely shouldn't be visible for jak's shadow, and I haven't found a way for the game to control this behavior yet.)

PR #2634, opened 2023-05-07
[jak2] experimental glow changes

water111:


ManDude:
This is no longer necessary with the updated glow renderer.

PR #2654, opened 2023-05-16
better decomp of `set-subtask-hook!`, `sub-task-list` array and `*-setting!` macros

ManDude:
Improves code readability considerably.


ManDude:
Also fixed the traffic spawn ranges being accidentally 10x larger than normal. Fixes #2658

PR #2662, opened 2023-05-17
jak2: add `current-time` macro for process clocks

### LINKED ISSUE #1872: [decomp] macro for `(-> self clock frame-counter)`
appears very often and would hide the process register use

Hat-Kid:
Closes #1872


goal_src/jak2/engine/debug/editable-player.gc
```diff
@@ -1370,21 +1370,20 @@
...
+            )
+       (while (< s4-23 s5-19)
+         (if a0-76
+             (editable-method-23 a0-76)
+             )
+         (+! s4-23 1)
+         (set! a0-76 (-> obj data s4-23))
+         )
+       )
+     (update-light-hash *light-hash*)
```
water111, line 1386: this changed, but it seems fine if it compiles

goal_src/jak2/engine/target/target-darkjak.gc
```diff
@@ -189,9 +189,9 @@
       (set! (-> a3-7 w) 1.0)
       (t9-14 a0-43 (the-as vector a1-11) (the-as quaternion a2-8) a3-7)
       )
-    ;; (set-darkjak-texture-morph! f30-0)
+    (set-darkjak-texture-morph! f30-0)
```
water111, line 192: this changed (I think it will still crash?)

ManDude: Does not crash.

PR #2663, opened 2023-05-18
[jak 2] Fix possible stereo desync in overlord

water111:
Normally, when they allocate a VagCmd, they do a bunch of stuff to clear all the status bits and reset things
in particular the InitVAGCmd function does a lot

image

but for the stereo command, they do a lot less:
image

Which means that the new_stereo_command can just have random status bits left over from whatever the last user had.
we seem to end up in a state where byte21 is set, and this causes everything else to be wrong and off-by-one dma transfer. My guess is that the original game avoided this bug due to lucky timing that I don't understand.

I think the fix of just clearing byte21 is ok because there's no way that the old value of the byte is useful after the command is repurposed.


game/overlord/jak2/vag.h
```diff
@@ -64,7 +64,7 @@ struct VagCmd {
       u8 byte8;
       u8 byte9;
       u8 byte10;
-      u8 byte11;
+      u8 byte11;  // is stereo secodnary
```
ManDude, line 67: typo

game/overlord/jak2/spustreams.cpp
```diff
@@ -989,31 +963,12 @@ void WakeSpuStreamsUp() {
...
-      uVar5 = uVar4;
-    } while (uVar4 == 0);
-  } while ((uVar4 < uVar7) || (uVar7 + 0x4040 <= uVar4));
-  return uVar4;
+  // this is simplified a lot from the original.
+  // they seem to sanity check if the value is reasonable or not, but the sanity check can fail
+  // if the overlord thread isn't keeping up.
+  // as far as I can tell, it's totally fine to discard these checks because our sceSdGetAddr
+  // works perfectly.
+  return sceSdGetAddr((param_1->voice & 0xffffU) | 0x2240);
```
Ziemas, line 971: There's two things going on here. There's the safety check of calling sceSdGetAddr three times, this is because the SPU address registers are composed of two 16bit registers which means there's risk of tearing when reading. This part can safely be omitted for us.

The other check seems to be that the address is contained within the allocated region (0x4040 bytes), if this is actually happening it means the voice has escaped the allocated area, this might be kinda bad if it actually happens for us.

PR #2665, opened 2023-05-18
Hungarian Custom Text JSON Disappearance fix, and translating new speedrun text

LuminarLight:
When the json format was introduced for text, the custom text translations I did in PR #2242 weren't present somehow. So I fixed this now, and also translated the few new text entries.


LuminarLight:
Also I noticed that text entries 138 and 161 are present in both the base text file and the custom text file. Which seems redundant.

xTVaser:
Sorry about that, I tried my best but not surprised something was lost in the shuffle.

> Also I noticed that text entries 138 and 161 are present in both the base text file and the custom text file. Which seems redundant.

They are redundant, you can remove them from the base file.  The point of the base file is for any string that can't be included the base `en-US` file. https://github.com/open-goal/jak-project/blob/master/game/assets/jak1/text/game_custom_text_en-US.json

fyi, you can also make these changes via Crowdin instead of making a PR.

LuminarLight:
Yeah I guessed it was difficult to keep track of all the changes while the rework to json was in progress.

I removed the redundant entries as suggested.

I saw the crowdin thing, but for now a PR feels better/clearer for me. I will use it for launcher translating though since I am fully unfamiliar with the launcher('s structure).


PR can be merged from my perspective.

xTVaser:
ok to test

PR #2669, opened 2023-05-21
overlord: Some cleanup

Ziemas:
Started cleaning up some of the lower hanging fruit.


Ziemas:
UnqueueMessage was inlined in ReleaseMessage, it was still really nasty on its own though but seems to be basically the same thing as Jak 1. A second set of eyes on these might be nice.

water111 (approved):
Thanks! I double checked these and they seem right to me.

PR #2672, opened 2023-05-25
Jak 2 pc subtitle support

ManDude:
Adds support for adding custom subtitles to Jak 2 audio. Comes with a new editor for the new system and format. Compared to the Jak 1 system, this is much simpler to make an editor for.

Comes with a few subtitles already made as an example.
Cutscenes are not officially supported but you can technically subtitle those with editor, so please don't right now.

This new system supports multiple subtitles playing at once (even from a single source!) and will smartly push the subtitles up if there's a message already playing:
image
image

Unlike in Jak 1, it will not hide the bottom HUD when subtitles are active:
image

Sadly this leaves us with not much space for the subtitle region (and the subtitles are shrunk when the minimap is enabled) but when you have guards and citizens talking all the time, hiding the HUD every time anyone spoke would get really frustrating.

The subtitle speaker is also color-coded now, because I thought that would be fun to do.

TODO:
- [x] proper cutscene support.
- [x] merge mode for cutscenes so we don't have to rewrite the script?


ManDude:
I think everything is working correctly now. I'll work on cutscene stuff at a later point.

Oversized PR is oversized.

PR #2677, opened 2023-05-30
Show the git commit version in screenshots/videos [Jak 2]

Zedb0T:
With issues/screenshots becoming more common its probably time to turn this on for debugging reasons.


Zedb0T:
(Can set *display-sha* to false in the startup.gc file to disable it for main devs)

goal_src/jak1/pc/pckernel-h.gc
```diff
@@ -374,7 +374,6 @@
 (define *display-heap-status* #f)
 (define *display-actor-counts* #f)
 (define *display-text-box* #f)
-(define *display-sha* *debug-segment*)
```
ManDude, line 377: Why remove this? `pckernel-h` is loaded for all games.

goal_src/jak1/pc/pckernel-common.gc
```diff
@@ -256,7 +258,9 @@
   `(begin
     (cpad-clear! 0 r1)
     (logxor! ,cheats-var (pc-cheats ,cheat))
-    (cheats-sound-play (logtest? ,cheats-var (pc-cheats ,cheat)))
+    (if (-> *pc-settings* speedrunner-mode?)
+    (sound-play "caught-eel")
```
ManDude, line 262: This sound effect is not in Jak 2! (what's the reason for this change anyway?) Also, fix the indenting please.

ManDude (changes requested):

goal_src/jak1/pc/pckernel-common.gc
Zedb0T: The reasoning for the change is in the commit 

```Add bad sound feedback if a cheat is attempted to toggle with speedrunner mode on

Before this it just didnt turn on but gave zero feedback so it was kind of bad...```

Then there is a video that is supplied as well. Do you have any suggestions for a sound that is present in both games?

ManDude: Oh, well it wasn't apparent that speedrun mode locked out cheats. Why not just *not set* the cheat variable instead of this, then?

ManDude: This code still lets the cheat code go through. If you input the cheat for blue eco, you will still get blue eco even with speedrun mode enabled.
image

PR #2692, opened 2023-06-04
[jak2] fix missing drill crane center part

water111:
We were forcing the drill crane to draw at lod 0, but we should have listened to the game and drawn it at lod 1.


ManDude:
Might be worth making the lod hack NOT override `force-lod`. Here's another instance where it's used: https://sourcegraph.com/github.com/open-goal/jak-project/-/blob/goal_src/jak2/levels/ruins/breakable-wall.gc?L229:25

As far as actual model quality, it only seems to be used to reduce quality in the intro for all the metal heads coming out of the portal (we should patch this in the intro code and not the drawable function).

water111:
Good idea. Using `force-lod` fixes the mech too

goal_src/jak2/engine/draw/drawable.gc
```diff
@@ -1144,8 +1142,11 @@
                       )
                     ;; lod hacks!
                     (with-pc
-                      (if (not (-> *pc-settings* ps2-lod-dist?))
-                        (set! lod-idx (minmax (-> *pc-settings* lod-force-actor) 0 (-> dc lod-set max-lod))))
+                      (when (not (-> *pc-settings* ps2-lod-dist?))
+                        (set! lod-idx (minmax (-> *pc-settings* lod-force-actor) 0 (-> dc lod-set max-lod)))
+                        (when (> (-> dc force-lod) 0))
```
ManDude, line 1147: I think this should be a `-1` instead of `0`, since `-1` means "don't use force lod"

water111: yep, good catch.

PR #2712, opened 2023-06-09
Rename `collide-action` enum values for readability

dallmeyer:
```lisp
(defenum collide-action
  :type uint32
  :bitfield #t
  (solid 0)              ;; used for solid things
  (rider-plat-sticky 1)  ;; used for platforms in rider/platform interactions
  (rider-target 2)       ;; used for target in rider/platform interactions
  (edgegrab-active 3)    ;; set/cleared when entering/exiting edgegrab states
  (rider-plat 4)         ;; used for platforms in rider/platform interactions
  (unused 5)             ;; totally unused?
  (edgegrab-possible 6)  ;; used when edge grab checks should be done
  (edgegrab-cam 7)       ;; set/cleared when entering/exiting edgegrab states
  (swingpole-active 8)   ;; set/cleared when entering/exiting swingpole states
  (racer 9)              ;; set/cleared when entering/exiting racer states
  (attackable 10)        ;; used for something to do with attacking/damaging
  (attackable-unused 11) ;; seems to relate to attacking - set in several places but never tested for?
  (snowball 12)          ;; set/cleared when entering/exiting snowball states
  (tube 13)              ;; set/cleared when entering/exiting tube states
  (flut 14)              ;; set/cleared when entering/exiting flutflut states
  (racer-grounded 15)    ;; set/cleared when entering/exiting certain racer states w/ extra conditions
  (racer-unused 16)      ;; seems to relate to racer - never set, only cleared in one place?
  )
```

### DIFF
--- a/goal_src/jak1/pc/pckernel.gc
+++ b/goal_src/jak1/pc/pckernel.gc
@@ -360,7 +360,7 @@
 
     (when (and  (-> obj controller-heat-led?)
                 (nonzero? (-> *target* racer))
-                (logtest? (-> *target* control root-prim prim-core action) (collide-action ca-9))
+                (logtest? (-> *target* control root-prim prim-core action) (collide-action racer))
                 (or (= (-> *target* current-level name) 'lavatube)
                     (= (-> *target* current-level name) 'firecanyon)
                     (= (-> *target* current-level name) 'citadel)))


dallmeyer:
Not entirely sure about the difference between `ca-1` and `ca-4`... Seems like they're both used on the platform in rider/platform interactions but I can't easily tell what each controls

`ca-11` is also weird, seems to be paired with `ca-10` and relate to being attack-able, but `ca-11` never actually gets checked anywhere that I see

ManDude:
you don't need to include `ca-` at the start.

dallmeyer:
> you don't need to include `ca-` at the start.

Personally I kind of like the ca- prefix to differentiate from e.g. `racer` (the process-drawable) vs `racer` (the collide-action), but I can remove it

ManDude:
It is unnecessary because it's already preceded by `collide-action` since it is an enum. Writing `(collide-action ca-racer)` is redundant.

ManDude:
You need to fix the enum in `pckernel.gc`.

dallmeyer:
> You need to fix the enum in `pckernel.gc`.

derp thanks, I didn't notice you had added another collide-action usage with the controller LED stuff - updated

PR #2719, opened 2023-06-12
revamp controller LED in jak 1 + reorganize some pc-settings things + fix some jak 2 decomp

ManDude:


goal_src/jak1/pc/pckernel-common.gc
```diff
@@ -243,6 +248,18 @@
     (pc-sound-set-fade-hack fade-hack)
     )
 
+  (update-led obj)
```
xTVaser, line 251: this seems like it will overwrite peoples LED to blue even if they don't have any of the options enabled?  Would be annoying if your controller's LED was not blue by default.

goal_src/jak1/pc/pckernel-common.gc
```diff
@@ -902,15 +816,15 @@
...
+        (reset obj #t)))
     (format 0 "[PC] PC Settings not found at '~S'...initializing with defaults!~%" *pc-temp-string-1*))
   0)
 
 (defmethod new pc-settings ((allocation symbol) (type-to-make type))
   "make a new pc-settings"
   (let ((obj (object-new allocation type-to-make (the-as int (-> type-to-make size)))))
     ;; load defaults not covered by the file
-    (set-defaults! obj)
+    (reset obj #f)
```
xTVaser, line 827: non-actionable, just a comment...

the diff is so large so I can't actually tell if this is a simplification or if it just makes the pc-settings more confusing with now an additional boolean....I tried to make things more explicit by separating concerns but I have a feeling that undoes some of that.

goal_src/jak1/pc/pckernel-common.gc
ManDude: well, like you said as soon as you set the led color you can't revert it back to "default" so I'm not sure that's gonna matter too much. Ideally we would have a "default color" setting somewhere.

goal_src/jak1/pc/pckernel-common.gc
ManDude: I think a boolean that determines whether we actually call the OS functions or not is more convenient than duplicating a lot of the settings code. We had a method to reset to defaults, and other methods that just "set" the defaults... it didn't make much sense with how the rest of the system was designed, so I changed it for a boolean. It seems relatively straight-forward to me, at least.

goal_src/jak1/pc/pckernel-common.gc
xTVaser: My point is that if a user has disabled all LED affecting settings, we should not still mess with their controller's LED -- as they may have set it externally / their controller has a different default color.  I'm not talking about us having our own default.

goal_src/jak1/pc/pckernel-common.gc
xTVaser: I agree with that, a lot of it has to be rewritten in my eyes because it requires too much intimate knowledge of the internals to avoid making a mistake -- I was trying to make small steps towards that and reduce the duplication over-time, but this just increases the complexity.

goal_src/jak1/pc/pckernel-common.gc
ManDude: DS4Windows seems to be the only program that sets the LED color and DS4Windows never sets it back to the color you selected unless you reconnect the controller (if it's yellow and you close the game, it will get stuck at yellow). I'm not sure if this is really worth it. I can add an arbitrary timer to prevent the LED from being set when the game boots if you have all settings off, but that's about it. Seems like a design flaw with the DS4 more than anything.

ManDude: Weirdly enough there seems to be some sort of mechanism that remembers what you set the LED to because if I set the color, close the game, reconnect my controller and open the game again it restores that last color.

Anyway, it seems like a fool's errand to handle not wanting to set the LED at all right now.

Issue #2699, opened 2023-06-06
SDL Regression - Wrong Fullscreen Resolution Chosen at High DPI


BreakPoints:
### Acknowledgements

- [X] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+sort%3Aupdated-desc) for duplicate or closed bug reports
- [X] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

Since the SDL changes, swapping to fullscreen mode with a non 100% scale set in Windows results in the wrong resolution being chosen.

From what I can see the desktop is forced into 100% scaling and a set resolution when fullscreening but if scaling was **not** set to 100% before swapping modes an incorrect desktop resolution is set. This means running at native resolution from the games settings still looks blurry.

Notes: Different scaling levels in Windows will lead to different incorrect desktop resolutions set, 2560x1600 125% ends up in a 4:3 resolution.

### How To Reproduce

- Set Windows display scaling to a non 100% value
- Boot game and set display mode to fullscreen
- Dev topbar is a good reference to check resolution as it does not scale

### Does this problem occur on original hardware or PCSX2?

Not needed, bug is obvious

### Expected Behavior

Native display resolution is selected for fullscreen

### Operating System

Windows 11 22H2

### OpenGOAL Version

dev

### Game Version

NTSC 1.0 (black label)

### Are you using a high refresh-rate monitor (>60Hz)

No

xTVaser:
@BreakPoints Let me know if this works better https://github.com/open-goal/jak-project/pull/2713 as for ImGUI being blurry, I think that is expected as you have to do a bunch of other work to get that right, and it's probably not worth it since it's just for debugging - https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-should-i-handle-dpi-in-my-application

But I was able to reproduce the imgui bar not spanning the entire width, as well as the resolution clearly being wrong when you attempt to minimize the game (indication that the fullscreen window size was too small (25% scaled down for example).  Both of those are now fixed on my end.

BreakPoints:
Tried #2713 and it's an improvement, the Windows Desktop is now kept at native resolution.

[...]

PR #2729, opened 2023-06-14
SDL: Hint to Windows that we handle DPI scaling

BreakPoints:
Fixes https://github.com/open-goal/jak-project/issues/2699

Setting a Windows specific SDL hint that disables OS forced scaling. With Windows handling scaling we are locked into a progressively smaller resolution buffer the higher the Windows display scaling option is set. Not sure if this is the best place to put it.

The game itself handles scaling already but ImGUI does not which can lead to it getting quite small on high DPI devices with high scaling. OS scaling can still be forced on the gk.exe as shown bellow to revert this behaviour if a blurry but larger ImGUI is preferable:

image

### DIFF
--- a/game/system/hid/display_manager.cpp
+++ b/game/system/hid/display_manager.cpp
@@ -7,6 +7,8 @@
 
 DisplayManager::DisplayManager(SDL_Window* window)
     : m_window(window), m_selected_fullscreen_display_id(0) {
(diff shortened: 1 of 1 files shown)


xTVaser:
ok to test

xTVaser (approved):

PR #2730, opened 2023-06-15
sdl: Directly use Windows DPI scaling hint, not SDL

BreakPoints:
Turns out I PR'd the wrong change. The SDL hint still required us to handle some changes with SDL_GetWindowSize, etc. Haven't figured out all the changes needed but just using the Windows level hint directly disables OS scaling and SDL scaling.

I tripped up as the SDL scaling still resulted in a sharp image but the pixels were nearest neighbour scaled which lead to an inconsistent blocky image.

Am I fine to just leave an OS specific call here? Linux compile didn't fail for me but maybe it should be hidden behind an OS check?

Fixes https://github.com/open-goal/jak-project/issues/2699
(for real this time)


ManDude (approved):

xTVaser:
ok to test

PR #2740, opened 2023-06-17
[jak2] fge=0 at the start of warp

water111:
The warp giftags don't set PRE and seem to rely on fogging being disabled at the start.


water111:
image

PR #2748, opened 2023-06-19
sdl: only call SDL video functions from the gfx thread

### LINKED ISSUE #2747: `pc_set_display_mode` is a blocking call
It only gets unblocked because of the timeout in the graphics thread, which is set to 50ms.

xTVaser:
Potentially fixes #2747


ManDude (approved):
Works fine for me, and makes the EE calls no longer blocking until timeout.

PR #2763, opened 2023-06-21
Update subtitle_en-US.json

atylmo:
Just a minor fix so the subtitles match the audio :)


ManDude:
He does not say "boy"

xTVaser:
It sounds like "boy" to me as well.  It also makes a lot more sense colloquially -- the voice acting is intentionally out of breath so I guess it's a little ambiguous.  But to my ears, there's a clear "bah" sound

Zedb0T:
Is there any chance we can deduce this from the same line in other languages? Or are these smaller lines typically improvised in other langs?

xTVaser:
In german he says "oh man, <words ill probably butcher, but probably mean "that was close">

ManDude:
There definitely is not "b" sound though, listen carefully. He might be saying "oi" or "oh", but definitely not "boy" - or it is "boy" and the "b" just didn't come through when the actor spoke.

BreakPoints:
This is reminiscent of the [McGurk effect](https://en.wikipedia.org/wiki/McGurk_effect), I can hear either if I prime myself for it.

What I will say is if it's a "oh" it's a strange sounding "oh". If it's "boy" it sounds more natural to me. Could be the b peaked the mic and it dropped or the audio clip starts a split second late?

xTVaser:
In my opinion, I think it's just intentional voice acting -- if you were trying to act being surprised / catching your breath (the context is almost crashing your racer), you would not perfectly enunciate the `b`

BreakPoints:
Either way, after much deliberation
boy

xTVaser:
I think we're gonna go with boy.  Other languages seem to corroborate it, and most importantly that's how the majority seem to hear it with only variation on the onomatopoeia.
image

xTVaser:
ok to test

PR #2766, opened 2023-06-24
[jak2] GOAL side texture animation stuff

water111:
It turns out we didn't decompile any of this stuff yet.


goal_src/jak2/engine/gfx/texture/texture-anim.gc
```diff
@@ -6,37 +6,1325 @@
...
+    (set! (-> (the-as (pointer gs-xy-offset) v1-13) 0) (new 'static 'gs-xy-offset :ofx #x7000 :ofy #x7300))
+    (set! (-> (the-as (pointer gs-reg64) v1-13) 1) (gs-reg64 xyoffset-1))
+    (set! (-> (the-as (pointer gs-frame) v1-13) 2)
+          (new 'static 'gs-frame :fbw (/ (+ width 63) 64) :fbp fbp-to-draw)
+          )
+    (set! (-> (the-as (pointer gs-reg64) v1-13) 3) (gs-reg64 frame-1))
+    (set! (-> dma-buf base) (&+ v1-13 32))
+    )
+    |#
+  0
```
ManDude, line 440: make sure there's NO casts at all. The detection only works on the uncast (broken) output.

water111: I unfortunately couldn't get it to work. The variables are used a little differently, and it wasn't an easy fix to make this one work.

PR #2767, opened 2023-06-24
fix more orb softlocks

ManDude:
Fixes orb softlocks during races and other side missions.

The side mission tasks will now not count as completed until the precursor orb has been picked up.

Races will not let you advance (or even pause the game) until the precursor orb has been picked up.


xTVaser:
Went to test if this causes a regression with something you can do in the original game:
https://youtu.be/B13Sia2r7Pc?t=2994

But hit a crash instead, so I'm not sure.  I would hope not since on original hardware you do get the orbs, so atleast from the description here it should continue to work.
https://github.com/open-goal/jak-project/issues/2768

ManDude:
That crash is not caused by the PR. It happens because an orb tries to spawn and it uses an invalid process to fetch its spawn position which leads to a misaligned qword read which are automatically aligned on the PS2.

Honestly though that really makes me want to reconsider all of this and just trash the orb spawning completely. It's clearly more trouble to make it work properly than it's worth lol

xTVaser:
I think a much safer approach, considering that you cannot _spend_ orbs in jak 2 -- is just to have a check that sees if there are any orbs that _should_ be awarded and if things don't add up - adjust the count.

This approach won't work for jak 3, but that's a problem for later.

ManDude:
How can we tell if an orb was correctly awarded or not, though?

xTVaser:
Assuming the in-level orbs don't effect this approach (there is a way to enumerate them) -- you add up the total that should be awarded from all the side-missions and hi-scores and adjust the count if it doesn't match.

ManDude:
You could use the AID of the entity-placed orbs to check if they have been picked up, so it's definitely doable, but that's also a hardcoded list of ~90 items to keep track of. Then a second hardcoded list for all the tasks that give you orbs as reward.

More importantly though, I have no idea right now where I would put those checks.

ManDude:
Changed it so there is no pickup and the orbs are awarded immediately. This requires way less hacks and does not (seem to) cause any additional bugs. It's also the same behavior as in Jak 3, so clearly Naughty Dog felt similarly.

PR #2780, opened 2023-06-27
Make all project targets compile on Intel MacOS

fabjan:
## Problem

Building the compiler doesn't work on macOS. This means programmers on macOS cannot really help contribute to the project. We cannot run the tests locally, and CI cannot catch regressions on macOS.

## Solution

Make the whole thing build on macOS.

Make the tests pass on macOS.

Make the macOS CI build the whole thing and run the tests!


game/kernel/jak1/kscheme.cpp
```diff
@@ -1712,12 +1727,10 @@ s32 InitHeapAndSymbol() {
...
-      MsgErr("\n");
-      MsgErr(
-          "dkernel: compiled C kernel version is %d.%d but the goal kernel is %d.%d\n\tfrom the "
-          "goal> prompt (:mch) then mkee your kernel in linux.\n",
-          KERNEL_VERSION_MAJOR, KERNEL_VERSION_MINOR, kernel_version >> 0x13,
-          (kernel_version >> 3) & 0xffff);
+      lg::error("Kernel version mismatch! Compiled C kernel version is {}.{} but the goal kernel "
+                "is {}.{}",
+                KERNEL_VERSION_MAJOR, KERNEL_VERSION_MINOR, kernel_version >> 0x13,
+                (kernel_version >> 3) & 0xffff);
```
fabjan, line 1733: I made this change while debugging runtime issues. Not using `lg::error` made a problem harder to see.

I don't know if there is some other important reason to use MsgErr instead.

common/cross_sockets/XSocketServer.cpp
```diff
@@ -42,19 +42,26 @@ bool XSocketServer::init_server() {
...
-  int server_socket_opt = SO_REUSEADDR | SO_REUSEPORT;
+  if (set_socket_option(listening_socket, SOL_SOCKET, SO_REUSEADDR, &yes, sizeof(yes)) < 0) {
+    close_server_socket();
+    return false;
+  }
+  // macOS doesn't support setting multiple options at once, so we have to do this separately.
+  if (set_socket_option(listening_socket, SOL_SOCKET, SO_REUSEPORT, &yes, sizeof(yes)) < 0) {
+    close_server_socket();
+    return false;
+  }
```
fabjan, line 56: I figured Linux can handle setting them separately, so they can just have the same code.

But I can create separate `__APPLE__` and `__linux` sections.

game/kernel/jak1/kscheme.cpp
```diff
@@ -264,7 +264,11 @@ u64 make_string_from_c(const char* c_str) {
 }
 
 extern "C" {
+#ifdef __APPLE__
+void _arg_call_linux() asm ("_arg_call_linux");
+#else
 void _arg_call_linux();
+#endif
```
fabjan, line 271: Without `asm ("_arg_call_linux")`, the object file will reference `__arg_call_linux` (two leading underscores) on macOS. Let me know if there is some way other than this you prefer.

game/kernel/jak1/kscheme.cpp
```diff
@@ -470,6 +479,9 @@ Ptr<Function> make_function_from_c(void* func, bool arg3_is_pp = false) {
   return make_function_from_c_linux(func, arg3_is_pp);
 #elif _WIN32
   return make_function_from_c_win32(func, arg3_is_pp);
+#elif __APPLE__
+  // FIXME - reusing linux for now
+  return make_function_from_c_linux(func, arg3_is_pp);
```
fabjan, line 484: DRY, but also, linux and macOS are different beasts. Is there any reason to believe this two versions should always be the same (for non-arm at least)? Otherwise perhaps it's better to copy-paste the implementation but call it `make_function_from_c_macos`.

game/sound/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ target_link_libraries(sound PRIVATE fmt cubeb common)
 add_executable(sndplay 989snd/sndplay.cpp)
 if(WIN32)
     target_link_libraries(sndplay PRIVATE sound cubeb)
+elseif(APPLE)
+    target_link_libraries(sndplay PRIVATE sound cubeb)
```
fabjan, line 30: `-lstdc++fs` was not found by my linker, and `#include <filesystem>` seems to work fine on macOS anyway.

game/kernel/asm_funcs.asm
```diff
@@ -436,6 +436,90 @@ _call_goal_on_stack_asm_linux:
...
+;; - second goal arg
+;; - third goal arg
+;; - address of function to call
+;; - address of the symbol table
+;; - GOAL memory space offset
+
+global _call_goal_asm_macos_x64
+
+;; FIXME - copied from linux, but not really tested
+_call_goal_asm_macos_x64:
```
Ziemas, line 453: As I understand it both MacOS and Linux use the System V calling convention, maybe rename the Linux variants and share it on both platforms?

fabjan: Aha, yeah it seems so:
https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI
(Programming at this low level is very much not my forte, so I've been guessing a lot for the ASM integration parts.)

I'll go ahead and suggest a rename in this PR then.

game/kernel/jak1/kscheme.cpp
fabjan: @Ziemas sorry for dragging you back, but I thought maybe you know here as well: is the calling convention the reason linux and win32 have different implementations of `make_function_from_c_XXX` and `make_stack_arg_function_from_c_XXX` as well, you think?

If so the linux ones could be renamed to `XXX_from_c_systemv`.

Ziemas: Yes, I think that's right. Those can be shared as well.

fabjan:
Marking this as ready for review now.

Both `./test.sh` and `build/goal-test` succeed locally for me:

```
... test runner output ...
[----------] Global test environment tear-down
[==========] 1467 tests from 54 test suites ran. (765454 ms total)
[  PASSED  ] 1467 tests.
```

common/cross_sockets/XSocketServer.cpp
water111: This seems fine, better to have only 1 implementation I think

game/kernel/jak1/kscheme.cpp
```diff
@@ -1712,10 +1725,9 @@ s32 InitHeapAndSymbol() {
     // check the kernel version!
     auto kernel_version = intern_from_c("*kernel-version*")->value;
     if (!kernel_version || ((kernel_version >> 0x13) != KERNEL_VERSION_MAJOR)) {
-      MsgErr("\n");
-      MsgErr(
-          "dkernel: compiled C kernel version is %d.%d but the goal kernel is %d.%d\n\tfrom the "
-          "goal> prompt (:mch) then mkee your kernel in linux.\n",
+      lg::error(
+          "Kernel version mismatch! Compiled C kernel version is {}.{} but"
+          "the goal kernel is {}.{}",
```
water111, line 1730: Thanks. For a historical note - we used `MsgErr` and this weird error message because that's what was in the original game. At the time this was written, I was more interested in matching the original game, but we've now shifted towards modernizing the C++ code, and using `lg::error` is a lot nicer.

water111 (approved):
Thanks for the cleanup on the assembly function names!

xTVaser:
add to whitelist

fabjan:
I wonder if there is something off in os.cpp with the AVX detection. I get this error trying to run `gk`:
```
-------- CPU Information --------
 Brand: 
 Model: 
 AVX  : false
 AVX2 : false
[08:05] [info] Your CPU does not support AVX, which is required for OpenGOAL.
```

My MacBook Air 2019 has a Core i5, which is supposed to have AVX, and `sysctl hw.optional` says I have AVX:
```
❯ sysctl hw.optional | grep avx
hw.optional.avx1_0: 1
hw.optional.avx2_0: 1
hw.optional.avx512f: 0
hw.optional.avx512cd: 0
hw.optional.avx512dq: 0
hw.optional.avx512bw: 0
hw.optional.avx512vl: 0
hw.optional.avx512ifma: 0
hw.optional.avx512vbmi: 0
```

Google´s `cpu_features` library also says yes to AVX:
```
❯ ./build/list_cpu_features 
arch            : x86
brand           : Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
family          :   6 (0x06)
model           : 142 (0x8E)
stepping        :   9 (0x09)
uarch           : INTEL_KBL
flags           : adx,aes,avx,avx2,bmi1,bmi2,clflushopt,clfsh,cx16,cx8,erms,f16c,fma3,fpu,lzcnt,mmx,movbe,pclmulqdq,popcnt,rdrnd,rdseed,sgx,ss,sse,sse2,sse3,sse4_1,sse4_2,ssse3,tsc
```

Ziemas:
https://github.com/open-goal/jak-project/blob/master/common/util/os.cpp#L26

Looks like the CPU feature check isn't implemented for __APPLE__, it just sets everything to 0.

xTVaser (approved):
Yeah the CPU feature identification was something I noticed as well, but `gk` in general has bigger problems so that can probably be fixed in a followup.

I think the dust has settled for the latest release (had to push out a hot fix for linux) so I think it's a good time to merge now.

PR #2784, opened 2023-06-29
repl: update replxx and some configuration for bracketed paste

### LINKED ISSUE #722: [repl] newlines in input eat previous output
It looks kinda funny when it happens.
image

xTVaser:
Closes #722

Took another stab at this and did way too much reading into the current state of terminals.  The proper way to support pasting multiple lines in a terminal is via bracketed paste - https://en.wikipedia.org/wiki/Bracketed-paste

replxx supports this and looking at the implementation it seems to do it right.  This can be confirmed from basically any linux terminal. (Confirmed on macOS - good enough! https://github.com/open-goal/jak-project/pull/2784#issuecomment-1612725605)

The problem is, PowerShell does not support bracketed paste, and neither does the much older `cmd`.  So the problem isn't our code, we'd have to do a whole bunch of windows specific console code for the `paste` handler (this is what powershell apparently does _instead_ of just supporting bracketed paste like everyone is hoping they would).

However this does allow you to submit multi-line inputs via `ctrl+enter` or `ctrl+j` now.  You just can't paste them in (on windows).

### DIFF
--- a/common/repl/util.cpp
+++ b/common/repl/util.cpp
@@ -150,6 +150,9 @@ void Wrapper::init_settings() {
   // NOTE - a nice popular project that uses replxx
   // - https://github.com/ClickHouse/ClickHouse/blob/master/base/base/ReplxxLineReader.cpp#L366
   repl.set_word_break_characters(" \t");
+  repl.set_complete_on_empty(false);
+  repl.set_indent_multiline(false);
+  repl.enable_bracketed_paste();
   // Setup default keybinds
   for (const auto& bind : repl_config.keybinds) {
     char32_t code;
@@ -167,8 +170,6 @@ void Wrapper::init_settings() {
     repl.bind_key(code, commit_text_action(bind.command));
   }
 }
-
-// TODO - command to print out keybinds
 
 void Wrapper::reload_startup_file() {
   startup_file = load_user_startup_file(username, repl_config.game_version);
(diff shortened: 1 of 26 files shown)


fabjan:
Not sure how many terminals we expect to test, so just throwing in an anecdote here.

The fix works in the VS Code terminal on macOS:

```
g > (
    repl-help
)

REPL Controls:
(:clear)
 - Clear the current screen
(e)
 - Exit the compiler
(lt [ip-address] [port-number])
 - Connect the listener to a running target. The IP address defaults to `127.0.0.1` and the port to `8112`
(r [ip-address] [port-number])
 - Attempt to reset the target and reconnect. After this, the target will have nothing loaded.
(:status)
 - Send a ping-like message to the target. Requires the target to be connected
(shutdown-target)
 - If the target is connected, make it exit
 ```

Without these changes I get read and compilation errors:

```
g > (
Reader error:
failed to find close paren
at Program string:1
(
^
[Warning] REPL Error: Reader error:
failed to find close paren
at Program string:1
(
^

g >     repl-help
-- Compilation Error! --
The symbol repl-help was looked up as a global variable, but it does not exist.
Form:
repl-help
Code:
repl-help
Location:
Program string:1
    repl-help
^
Code:
(top-level repl-help)
[Warning] REPL Error: Compilation Error
```

PR #2789, opened 2023-06-30
[Jak II] Allow Returning to Palace Interior

LuminarLight:
A childhood dream of probably many...
There is no real reason to block the player from returning to the Palace Interior.
So I decided to start working on making it revisitable, after the player completes the game of course.

There were two main obstacles:
- The Palace Interior is supposed to be explored in a specific way/route, and Andy Gavin ensured this using the tasks system. But after we complete the game, we can no longer rely on tasks. So my idea was to just make the interior explorable in any way/order, which seems doable.
- The Palace Rooftops elevator is in the shaft, and normally usable when you are coming down from the rooftops after finishing the game. I needed to find a way to prevent the two elevators from conflicting with each other.

I wanted to keep these changes as clean as possible. I had to add a 'palshaft-activate' function, this is very essential for all the changes I did. But for example I really don't want to add new event handlers to the elevator (even though it would make things a lot easier).

Progress tracker:
- [x] Make the elevator that leads to palace interior appear after the game is done.
- [x] Let the player enter the palace interior.
- [x] Allow moving between all palace interior rooms/levels flawlessly.
- [x] Make the 'palout' level appear when it should be visible.
- [x] Allow leaving the palace interior through the entrance.
- [x] Allow leaving the palace interior through the exit.
- [x] When using the palace rooftops elevator, neutralize the other elevator, to prevent conflicts.
- [x] Restore the elevators when we cross the main entrance.

Keeping PR as a draft for now, since I need to clean things up and there are still some minor weirdnesses, though functionally everything seems to be working.


ManDude:
You want to check https://github.com/open-goal/jak-project/blob/master/goal_src/jak2/levels/common/warp-gate.gc#L773-L783 for an example on replacing an existing entity's data.

Also, the task you want to check is `nest-boss-resolution`. `city-win-resolution` is never closed in the actual game.

xTVaser:
Is this still being actively developed?  No activity in a month, should I close and you can reopen once you've finalized your PR?

LuminarLight:
ManDude expects a different approach, which may work for some things, but for example for the elevator I couldn't get it to work. It may be because the kill mask kills it before a process is created? I never figured it out. But changing things in the activate function worked reliably.

Anyway, I see that there are even conflicts now. And my changes are kind of a hacky mess...

The changes I made kind of show/highlight what needs to be done, but it needs to be done in a much nicer way.

I may return to this one day, but I think I will start over - I can look at my branch for reference if I ever return to this. So I guess this can be closed?

PR #2796, opened 2023-07-04
[decompiler] new features + fixes, + other jak2 fixes

ManDude:
Fixes empty boxed arrays of strings breaking some decomp (`ctywide-speech` and `race-info`).

Adds `decomp-as` tag to decompiler types so that the static data decompiler can use macros like `meters` and `seconds` on fields that aren't of type `meters` or `time-frame`.

Adds `override` tag to decompiler types which overrides the type of field with that name. The type must be a child type of the original field's type (or the same type, but why would you do this?).

Fixes the camera being offset for `drillmtn` after loading `palout` once.

This is a huge refactor sadly.


decompiler/config/jak2/ntsc_v1/var_names.jsonc
xTVaser, line 1: looks like you forgot to re-format this file

ManDude: yeah fixed right as you posted lol

PR #2811, opened 2023-07-07
Make Jak1 playable on macOS (intel)

fabjan:
## Problem

OpenGOAL uses OpenGL 4.3.

Apple stopped upgrading OpenGL after 4.1, so the way OpenGOAL currently works will never be playable on Macs using OpenGL.

## Solution

Luckily, downgrading to OpenGL 4.1 is not a huge change (at least it doesn't seem like it to my untrained eyes).

## Changes

* set hints for OpenGL 4.1 instead of 4.3 for __APPLE__
* skip the OpenGL debugging callback setup for macOS (requires 4.3)
* bump down the version string for all shaders
* stop using the `binding` layout qualifier in shader code
* move the `flat` qualifier first (not sure if this is a 4.1 thing or just Macs being more strict)
* don't mix signed and unsigned ints in shaders (not sure if this is a 4.1 thing or just Macs being more strict)
* add some hacky CPP to the Shader constructor for binding texture units and bones buffers based on variable names in the shader code

## Results

image
Skärmavbild 2023-07-07 kl  13 10 30
Skärmavbild 2023-07-07 kl  13 13 48

### DIFF
--- a/game/graphics/opengl_renderer/OpenGLRenderer.cpp
+++ b/game/graphics/opengl_renderer/OpenGLRenderer.cpp
@@ -74,11 +74,14 @@ OpenGLRenderer::OpenGLRenderer(std::shared_ptr<TexturePool> texture_pool,
       m_version(version) {
   // setup OpenGL errors
   glEnable(GL_DEBUG_OUTPUT);
+  // requires OpenGL 4.3
+#ifndef __APPLE__
   glDebugMessageCallback(opengl_error_callback, nullptr);
   // disable specific errors
   const GLuint gl_error_ignores_api_other[1] = {0x20071};
   glDebugMessageControl(GL_DEBUG_SOURCE_API, GL_DEBUG_TYPE_OTHER, GL_DONT_CARE, 1,
                         &gl_error_ignores_api_other[0], GL_FALSE);
+#endif
 
   lg::debug("OpenGL context information: {}", (const char*)glGetString(GL_VERSION));
 
--- a/game/graphics/opengl_renderer/Shader.cpp
+++ b/game/graphics/opengl_renderer/Shader.cpp
@@ -11,8 +11,6 @@ Shader::Shader(const std::string& shader_name, GameVersion version) : m_name(sha
   const std::string scissor_height = version == GameVersion::Jak1 ? "448.0" : "416.0";
   const std::string scissor_adjust = "512.0 / " + scissor_height;
 
-  lg::debug("Loading shader {}", shader_name);
-
   // read the shader source
   auto vert_src =
       file_util::read_text_file(file_util::get_file_path({shader_folder, shader_name + ".vert"}));
@@ -73,14 +71,12 @@ Shader::Shader(const std::string& shader_name, GameVersion version) : m_name(sha
     std::string uniformName = "tex_T" + std::to_string(i);
     GLint texLoc = glGetUniformLocation(m_program, uniformName.c_str());
     if (texLoc != -1) {
-      lg::debug("Shader {}: setting up texture uniform {}", shader_name, uniformName);
       glUniform1i(texLoc, i);
     }
   }
   // assuming that the bones uniform block is always using binding point 1
   GLint bonesLoc = glGetUniformBlockIndex(m_program, "ub_bones");
   if (bonesLoc != -1) {
-    lg::debug("Shader {}: setting up bones uniform block", shader_name);
     glUniformBlockBinding(m_program, bonesLoc, 1);
   }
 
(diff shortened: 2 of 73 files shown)


game/graphics/opengl_renderer/Shader.cpp
```diff
@@ -10,6 +10,9 @@ Shader::Shader(const std::string& shader_name, GameVersion version) : m_name(sha
   const std::string height_scale = version == GameVersion::Jak1 ? "1.0" : "0.5";
   const std::string scissor_height = version == GameVersion::Jak1 ? "448.0" : "416.0";
   const std::string scissor_adjust = "512.0 / " + scissor_height;
+
+  lg::debug("Loading shader {}", shader_name);
```
xTVaser, line 14: there's some left over debug logs in this file which maybe should go?  Not sure if they will be spammy or not.  If they aren't then ignore this.

fabjan: I don't think they are very spammy, one loading per shader loaded on boot, and for a few of them debug logs are emitted if they get uniforms bound.

The default log levels are
```cpp
    lg::set_file_level(lg::level::debug);
    lg::set_stdout_level(lg::level::warn);
    lg::set_flush_level(lg::level::warn);
```

game/graphics/opengl_renderer/OpenGLRenderer.cpp
```diff
@@ -74,11 +74,14 @@ OpenGLRenderer::OpenGLRenderer(std::shared_ptr<TexturePool> texture_pool,
       m_version(version) {
   // setup OpenGL errors
   glEnable(GL_DEBUG_OUTPUT);
+  // requires OpenGL 4.3
+#ifndef __APPLE__
   glDebugMessageCallback(opengl_error_callback, nullptr);
   // disable specific errors
   const GLuint gl_error_ignores_api_other[1] = {0x20071};
   glDebugMessageControl(GL_DEBUG_SOURCE_API, GL_DEBUG_TYPE_OTHER, GL_DONT_CARE, 1,
                         &gl_error_ignores_api_other[0], GL_FALSE);
```
fabjan, line 83: These are the reason I didn't just go with `SDL_GL_CONTEXT_MINOR_VERSION, 1` for all platforms. Not using two different versions, but only 4.1 for all platforms with OpenGL, I think would make things simpler. I don't think a debugging callback at all is available in 4.1 though.

water111: I agree. This debugging callback is super useful, so it's worth the effort to keep for linux/windows.

game/graphics/opengl_renderer/Shader.cpp
```diff
@@ -64,6 +67,23 @@ Shader::Shader(const std::string& shader_name, GameVersion version) : m_name(sha
...
 
+  // uniform samplers must be named matching the texture unit
+  glUseProgram(m_program);
+  for (int i = 1; i < 30; ++i) {
+    std::string uniformName = "tex_T" + std::to_string(i);
+    GLint texLoc = glGetUniformLocation(m_program, uniformName.c_str());
+    if (texLoc != -1) {
+      lg::debug("Shader {}: setting up texture uniform {}", shader_name, uniformName);
+    }
+    glUniform1i(texLoc, i);
```
water111, line 78: I think this `glUniform1i` should be inside the `if` statement above

game/graphics/opengl_renderer/Shader.cpp
```diff
@@ -64,6 +67,23 @@ Shader::Shader(const std::string& shader_name, GameVersion version) : m_name(sha
...
+  }
+  // assuming that the bones uniform block is always using binding point 1
+  GLint bonesLoc = glGetUniformBlockIndex(m_program, "ub_bones");
+  if (bonesLoc != -1) {
+    lg::debug("Shader {}: setting up bones uniform block", shader_name);
```
water111, line 83: I would probably remove these two `lg::debug` statements, there's going to be a lot of them for all the shaders and they aren't that useful.

water111 (commented):
Thanks for doing this - we've been talking about it for years but never got around to it!

I just had two minor comments, but otherwise this looks good to me.

game/graphics/opengl_renderer/Shader.cpp
fabjan: oh, yeah that was my intention as well! good catch

game/graphics/opengl_renderer/Shader.cpp
fabjan: There are only two shaders that have `ub_bones` uniforms, 23 shaders use some `tex_TXX` uniform, they would log more (49 total). I will remove the extra debug logs though, if they are not useful.

PR #2823, opened 2023-07-11
implement eval msgs

himham-jak:
added functionality beyond pinging, let me know how this looks.

I split off respond() since the functionality inside ping_response() was reusable for eval_response()

I added a reference to the compiler in the .h so we can pass arguments to the REPL, not sure if this is best
then I made a constructor to assign the compiler ref and send a message along it in eval_response()

lastly I updated the logic at the bottom to switch between ping and eval methods


xTVaser:
Going to close as there isn't a ton of activity and it seems to still be in very early stages.  Also from conversations in the discord the overall goal here will require a lot more work to be done properly.  Feel free to re-open when it's in better shape

PR #2827, opened 2023-07-12
Get the project compiling on Apple Silicon macOS natively (arm64)

xTVaser:
I havn't tested it yet, but I can almost guarantee that atleast `goalc` will not work in the slightest!

But the project is atleast fully compiling.  My hope is to start translating some AVX to NEON next / get `goalc` working...eventually.


game/graphics/opengl_renderer/DirectRenderer2.cpp
```diff
@@ -1,6 +1,10 @@
 #include "DirectRenderer2.h"
 
+#ifdef __aarch64__
+#include "third-party/sse2neon/sse2neon.h"
+#else
```
water111, line 5: is this even needed here? I think we can remove `immintrin.h` entirely

game/graphics/opengl_renderer/SkyBlendCPU.cpp
```diff
@@ -1,6 +1,10 @@
 #include "SkyBlendCPU.h"
 
+#ifdef __aarch64__
+#include "third-party/sse2neon/sse2neon.h"
```
water111, line 4: SkyBlendCPU is not used anymore - we can just disable this renderer on ARM.

game/graphics/opengl_renderer/background/background_common.cpp
```diff
@@ -2,7 +2,11 @@
 
 #include "background_common.h"
 
+#ifdef __aarch64__
+#include "third-party/sse2neon/sse2neon.h"
```
water111, line 6: I would fall back to `interp_time_of_day_slow` on arm instead of this.

game/kernel/asm_funcs_x86_64.asm
```diff
@@ -10,12 +6,11 @@ SECTION .text
 global _arg_call_systemv
 _arg_call_systemv:
   pop rax
-  push r10
-  push r11
-  sub rsp, 8
+  push r10 ; arg 6 (arg in GOAL only so must be preserved)
+  push r11 ; arg 7 (arg in GOAL only so must be preserved)
```
water111, line 10: Minor nit - but the reason to preserve this isn't because it's an argument in GOAL only, it's because the OpenGOAL compiler expects this register to be saved but systemv doesn't save it.

goalc/emitter/Register.h
```diff
@@ -60,6 +60,78 @@ enum X86_REG : s8 {
...
+  X16,  // temp, not-saved
+  X17,  // temp, not-saved
+  X18,  // temp, not-saved
+
+  SP,  // stack pointer
+
+  x19,  // saved TODO purpose?, R12
+  x20,  // pp, R13
+  x21,  // st, R14
+  x22,  // offset, TODO purpose?, R15
```
water111, line 94: what about x23 to x29?

goalc/emitter/Register.h
```diff
@@ -60,6 +60,78 @@ enum X86_REG : s8 {
   XMM15,  // saved
 };
 
+// TODO - i think it'll be better to make some sort of abstraction
+// mapping between x86 and arm, btu just using this enum as a place to prototype
+// the registers to use.
```
water111, line 65: I think we should avoid an abstraction over registers, and the existing abstractions within the compiler are enough. I would keep this enum (maybe split it for simd/gprs), and add ArmGpr and ArmSimd to HwRegKind.

For frontend compiler stuff, we can modify `make_gpr()`/`make_fpr()` to return appropriately types for the target architecture. In places where we currently specify x86 register constraints in the front end, we'll need to make ARM specific versions. It won't be possible to abstract these places. (but there are only a few).

For instruction/code generation stuff, registers are treated as just a `Kind`, `int`, where the `Kind` tells you xmm/gpr/(arm_gpr)/(arm_simd), and the second int is the "hardware id" that actually goes in the instruction.

We'll need to eventually write GOAL code that refers to a specific register to implement stuff like switching GOAL threads.  There are also weird ways that we'll need to generate code/constrain registers differently. (for example, x86 variable shifts require that shift amount is in rcx, and division has some weird requirements too).

For GOAL code outside of the kernel, we never specify register mapping. So the existing `rlet`s in code aren't specific to x86.

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  stp q13, q12, [sp, #-32]!
+  stp q11, q10, [sp, #-32]!
+  stp q9, q8, [sp, #-32]!
+
+  blr x8
+
+  ldp x8, x9, [sp], #32
+  ldp x10, x11, [sp], #32
+  ldp x12, x13, [sp], #32
+  ldp x14, x15, [sp], #32
```
water111, line 31: I think these should follow the same order as the stp: `ldp x15, x14...` (but it's been a while, i could be wrong)

goalc/emitter/Register.h
```diff
@@ -60,6 +60,78 @@ enum X86_REG : s8 {
...
+  x19,  // saved TODO purpose?, R12
+  x20,  // pp, R13
+  x21,  // st, R14
+  x22,  // offset, TODO purpose?, R15
+
+  // quadword registers, equivalent to XMMs
+  // the convention in arm64 is the callee preserves all Q values
+  // at the same time though, the caller should not depend on this convention!
+  // TODO - ARM has double the amount available to us
+  // likely some optimizations can be done with that in mind
```
water111, line 100: We shouldn't have to worry about making explicit changes to get this benefit - all xmm assignments for x86 are handled by the register allocator. So if more registers are available, it will just end up spilling less and the code will be better.

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  ret
+
+;; Call c++ code through mips2c.
+;; GOAL will call a dynamically generated trampoline.
+;; The trampoline will have pushed the exec function and stack offset onto the stack
+.global _mips2c_call_arm64
+.align 4
+_mips2c_call_arm64:
+  ;; TODO - this is really weird using half an XMM, this makes the arm assembly
+  ;; more difficult - this probably isn't required for arm?
```
water111, line 91: yep, you're right. I only used `xmm0` because all the x86 GPRs were used and it was convenient.

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  str x3, [sp, #+112] ; arg 2 (RDX in x86) and arg 3 (RCX in x86)
+  str x4, [sp, #+128] ; arg 4 (R8 in x86) and arg 5 (R8 in x86)
+  str x5, [sp, #+144] ; arg 4 (R8 in x86) and arg 5 (R8 in x86)
+  str x6, [sp, #+160] ; arg 6 (R10 in x86) and arg 7 (R11 in x86)
+  str x7, [sp, #+176] ; arg 6 (R10 in x86) and arg 7 (R11 in x86)
+  str x20, [sp, #+352] ;; s6 (pp) (R13 in x86) and s7 (st) (R14 in x86)
+  str x21, [sp, #+368] ;; s6 (pp) (R13 in x86) and s7 (st) (R14 in x86)
+
+  mov x0, sp ; move the stack pointer to arg 0
+  sub x0, x0, x22 ; R15 is a "special" offset TODO - whats special about it?
```
water111, line 122: This is the offset between real pointers, and GOAL pointers. The tricks are:
- when doing `call` or `blr` or load or store, the register needs to be a real pointer.
- Real pointers will be 64-bits
- GOAL pointers can only be 32-bits (they are stored in types as 32 bits only)
- To convert: `goal_pointer + offset = real_pointer` where `offset` is a 64-bit value.
- the stack pointer is a real pointer. This is just convenient so we can push/pop.
- in mips2c code, we need to set the emulated `sp` register's value to a goal pointer (to simulate what it would have been on ps2). So we subtract the offset here. The store below this is what stores `sp` into the simulated MIPS2C register struct.

water111: https://github.com/open-goal/jak-project/blob/master/game/mips2c/mips2c_private.h#L225

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  str x0, [sp, #+464] ;; mip2c code's MIPS stack
+
+  mov x0, sp ;; move the stack pointer to the new position
+
+  sub sp, sp, x8 ;; allocate space on the stack for GOAL fake stack
+  stp x8, x8, [sp, #-16]! ;; and remember this so we can find our way back
+
+  ;; TODO - this used to be a movq rax, xmm0
+  ;; TODO - not sure why an `xmm` was used because that movq only uses the lower 64bits anyway
```
water111, line 131: i ran out of other registers, otherwise I would have used a 64-bit one.

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  ;; x5 - off (goes in x22)
+
+  ;; set GOAL process
+  mov x20, x4
+  ;; symbol table
+  mov x21, x4
+  ;; offset
+  mov x22, x5
+  ;; call GOAL by function pointer
+  blr x3
```
water111, line 183: this ends up writing the `x30` (link register) with the address of the next instruction so the function being called can return to the right place.

water111: You'll need to back up this register and restore it before `ret` in this function, otherwise the `ret` and the end of this function will branch back to right after this line. (equivalent to how `ra` works in mips)

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  ldr x8, [sp, #+32]
+
+  add sp, sp, 1280 ; reset the stackpointer back
+
+  ldp x8, x9, [sp], #32
+  ldp x10, x11, [sp], #32
+  ldp x12, x13, [sp], #32
+  ldp x14, x15, [sp], #32
+
+  add sp, sp, 24 ;; 16 for the stuff pushed by trampoline
```
water111, line 148: x86 also needs 16-byte aligned stacks, but `call`/`ret` on x86 push/pop 8 byte values (but this is not the case on arm). All these offsets and stuff assume the 8-bytes are added and might need to be redone for arm (`blr` stores the return address in a register instead of pushing).

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  mov x21, x4
+  ;; offset
+  mov x22, x5
+  ;; call GOAL by function pointer
+  blr x3
+
+  ;; restore saved registers.
+  ldr x22, [sp], #16
+  ldp x20, x21, [sp], #16
+  ret
```
water111, line 188: it can help to look on godbolt to see what real compilers do:
https://godbolt.org/z/PP97TTbzT

game/kernel/asm_funcs_arm64.s
```diff
@@ -0,0 +1,262 @@
...
+  stp x22, x9, [sp, #-16]!
+
+  ;; switch to new stack
+  mov sp, x0
+
+  mov x20, x4 ;; set GOAL function pointer  
+  mov x21, x4 ;; symbol table
+  mov x22, x5 ;; offset
+  ;; call GOAL by function pointer
+  blr x3
```
water111, line 256: I think all of these assembly functions will need to backup/restore x30

water111 (commented):
This is awesome progress. I left a bunch of random comments/thoughts. I need to review some of the compiler code before I can say much more.

I remember when I was first generating x86 code, something that helped a lot was the `test_CodeTester.cpp` tests.  It's basically a way to test generating and running small sections of code, without most of the compiler infrastructure. We might want to do those for ARM as well.

common/util/crc32.cpp
```diff
@@ -2,26 +2,11 @@
 
 #include <cstring>
 
-#ifdef __arm__
```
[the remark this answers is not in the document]
xTVaser: for me the `__arm__` definition wasn't defined, so I could have fixed that but fixing the intrinsics removed the need to

CMakeLists.txt
```diff
@@ -112,7 +112,6 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
         -Wredundant-decls \
         -Wshadow \
         -Wsign-promo \
-        -O3 \
```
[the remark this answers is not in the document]
xTVaser: Because `-O3` is applied below if it's a Release build, but having it here means it will always have full optimizations, even if it's a debug build

common/custom_data/TFrag3Data.cpp
```diff
@@ -3,7 +3,11 @@
 #include <algorithm>
 #include <functional>
 
+#ifndef __aarch64__
 #include "xmmintrin.h"
+#else
+#include "third-party/sse2neon/sse2neon.h"
```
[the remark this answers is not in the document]
xTVaser: I had a similar worry, but considering that nothing works yet for ARM -- I felt like this was a decent compromise to just get things compiling and they can be replaced / optimized later?  It should be very easy to remove the header and resolve the compiler errors

game/kernel/asm_funcs_arm64.s
xTVaser: Nope you're right, I had to rewrite the stack pushing/popping too many times and got that wrong.

goalc/emitter/Register.h
xTVaser: Yeah my primary concern from doing a few searches was how there's a decent amount of code that refers to explicit registers, atleast by name.  Ranging from the debugger code, the IR code, and obviously the kernels.  But I havn't dug into any of them in detail yet -- I just don't want to make the code a mess of ifdefs.  

It certainly makes sense in some spots like the kernel, but I was imagining nothing more than an abstraction like "give me the register for arg0" which either returns `rdi` or `x0` instead of that being communicated by comments.

But when I look into fixing up the related code, it might not even be relevant or worth it.

game/kernel/asm_funcs_arm64.s
xTVaser: Somewhat related -- what is the reason why `rax` is pushed twice here? https://github.com/open-goal/jak-project/blob/33858cbae6ab2c68958f4ac5f56619a249e3433c/game/kernel/asm_funcs.asm#L151-L160

common/custom_data/TFrag3Data.cpp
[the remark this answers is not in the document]
water111: Sounds good. I'd rather not keep it for long, but let's merge it as-is and then I will go through and rewrite these at some point.

goalc/emitter/Register.h
water111: > , but I was imagining nothing more than an abstraction like "give me the register for arg0" which either returns rdi or x0 instead of that being communicated by comments.

Yeah, that sounds great to me.  I think the key detail is that the `Register` type should still hold actual real hardware registers (`rdi` or `x0`), rather than some generic register representation. The compiler is designed so that we only `Register` when we really care a specific register. 

The specific example you mentioned is mostly already done:
https://github.com/open-goal/jak-project/blob/master/goalc/compiler/compilation/Function.cpp#L185. We'd need to add an ARM case, but we already have a function that returns `vector<Register>` for all the arguments/return value in a function call. 
(due to 128-bit arguments/return values, the actual registers used depend on the type of the function)

>  there's a decent amount of code that refers to explicit registers, atleast by name.

Maybe I'm missing something, but I don't think there's too much here. I _think_ I've done a decent job of only doing this when absolutely needed. Stuff like this:
https://github.com/open-goal/jak-project/blob/master/goalc/compiler/compilation/Math.cpp#L465
is just a requirement of x86 and it will be different on ARM (which needs no constraint here, I think you can use any register for integer division).

game/kernel/asm_funcs_arm64.s
water111: This was just me being dumb. The second is "needed" to align the stack properly to 16-bytes. But this could be replaced with `sub rsp, 8` and `add rsp, 8`

PR #2839, opened 2023-07-23
Change force-actors? to ps2-actor-vis?

### LINKED ISSUE #2815: Jak 1 actor culling option is flipped?
It noticed that the Actor Culling option seems to work in reverse, as actors are reduced when Off instead of when On. I assume this isn't the expected behavior, as Particle Culling reduces particles when On.

Actor Culling off:
jak1_2023-07-08T01-37-19

Actor Culling on:
jak1_2023-07-08T01-37-03

(see the barrels and orbs)

This is on the release v0.1.38

Zedb0T:
Closes #2815

Follows the lead set by `ps2-parts?` and changes `force-actors?` to `ps2-actor-vis?` and inverts all the conditions that use this value, so now it shows all actors when `ACTOR CULLING` is set to `OFF` in the menu

The other solution would I could do would be to leave it how it was, and just change the `ACTOR CULLING` text in the menu to say something like `PC ACTOR DISTANCE` or something... so then it would describe the expected behavior when turned on.

I opt'd for the first option just to be more consistent with the already established `ps2-parts?` and avoid translation issues when possible but can change if needed just let me know.

### DIFF
--- a/goal_src/jak1/pc/debug/default-menu-pc.gc
+++ b/goal_src/jak1/pc/debug/default-menu-pc.gc
@@ -755,6 +755,7 @@
          (menu "PS2 settings"
             ;(flag "PS2 Load speed" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-read-speed?)))
             (flag "PS2 Particles" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-parts?)))
+            (flag "PS2 Actor Vis" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-actor-vis?)))
             ;(flag "PS2 Music" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-music?)))
             ;(flag "PS2 Sound effects" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-se?)))
             ;(flag "PS2 Hints" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-hints?)))
@@ -795,7 +796,6 @@
             (flag "heat" #f ,(dm-lambda-boolean-flag (-> *pc-settings* controller-heat-led?)))
             )
          (flag "V-sync" #f ,(dm-lambda-boolean-flag (-> *pc-settings* vsync?)))
-         (flag "All actors" #f ,(dm-lambda-boolean-flag (-> *pc-settings* force-actors?)))
          (flag "Display actor counts" *display-actor-counts* dm-boolean-toggle-pick-func)
          (flag "Display git commit" *display-sha* dm-boolean-toggle-pick-func)
          (flag "Extra hud elements" #f ,(dm-lambda-boolean-flag (-> *pc-settings* extra-hud?)))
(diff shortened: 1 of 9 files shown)


goal_src/jak1/pc/pckernel-common.gc
```diff
@@ -697,7 +697,7 @@
     (("ps2-hints?") (set! (-> obj ps2-hints?) (file-stream-read-symbol file)))
     (("ps2-lod-dist?") (set! (-> obj ps2-lod-dist?) (file-stream-read-symbol file)))
     (("force-envmap?") (set! (-> obj force-envmap?) (file-stream-read-symbol file)))
-    (("force-actors?") (set! (-> obj force-actors?) (file-stream-read-symbol file)))
+    (("ps2-actor-vis?") (set! (-> obj ps2-actor-vis?) (file-stream-read-symbol file)))
```
ManDude, line 700: Do not remove this string from the case list unless you want everyone's settings to be reset after this is merged!

goal_src/jak1/pc/debug/default-menu-pc.gc
```diff
@@ -755,6 +755,7 @@
          (menu "PS2 settings"
             ;(flag "PS2 Load speed" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-read-speed?)))
             (flag "PS2 Particles" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-parts?)))
+            (flag "PS2 Actor Vis" #f ,(dm-lambda-boolean-flag (-> *pc-settings* ps2-actor-vis?)))
```
ManDude, line 758: Can you keep this where it was before? It was much easier to toggle in the parent menu.

ManDude (changes requested):

goal_src/jak1/pc/debug/default-menu-pc.gc
Zedb0T: Sure thing!

goal_src/jak1/pc/pckernel-common.gc
Zedb0T: I changed both the file IO lines to use `force-actors?` in order to maintain compatibility. Is there anything else I should be aware of?

ManDude: Yes. People's settings files will have a `force-actors?`. If you removed the case, then the settings parser will assume it's an invalid key and nuke everything.

Zedb0T: With the changes in 1c0e5ab this should be accounted for and the parser should expect to find `force-actors?` like it has in the past no?

https://github.com/open-goal/jak-project/blob/1c0e5abb6f2374b27b1ba26a9489f552aa52c53b/goal_src/jak1/pc/pckernel-common.gc#L700

https://github.com/open-goal/jak-project/blob/1c0e5abb6f2374b27b1ba26a9489f552aa52c53b/goal_src/jak1/pc/pckernel-common.gc#L787

ManDude: I think you should:
- remove the `force-actors?` key from the file write function so it stops showing up in settings files.
- in the file read function, keep the old `force-actors?` setting but make it set the opposite value - that way people's settings aren't suddenly inverted when updating.
- add a new `ps2-actor-vis?` case that just works normally.

PR #2848, opened 2023-07-28
fix stad-samos regression from refactor (#2796) + change imgui toggle to left-alt and make it customizable

ManDude:


xTVaser:
Apply this patch please to return the keybind back to left alt, and make it customizable.  You can add the following to your `GAME/misc/debug-settings.json` file if you want it to be F11:
```json
"hide_imgui_key": 1073741892
```
[imgui-keybind-customizable.txt](https://github.com/open-goal/jak-project/files/12196714/imgui-keybind-customizable.txt)

ManDude:
Where does this number come from?

xTVaser:
it's the value that `SDLK_F11` maps to

PR #2876, opened 2023-08-02
cleanup our cmake and build warnings

xTVaser:
There are now zero warnings when doing a full compile on my end (windows + clang).  It revealed a few actual problems too, like an ignored clamp in the sound code / some bad regexes in other places.  

Most of them were just nagging unused variables which I kept around because they seemed intentional for debugging.


water111 (approved):
Thanks!

PR #2934, opened 2023-08-26
Add flag to make merc models common (stuffed into GAME.fr3)

dallmeyer:
Open to better names for that config flag 😅

With flag off, GAME.fr3 is ~800KB. With flag on, GAME.fr3 is ~25MB


dallmeyer:
#2936 seems like a better solution

PR #2936, opened 2023-08-27
custom levels: add support for packing models and textures into the FR3 file

Hat-Kid:
Previously, custom levels could only have non-common merc models in them if you had another level loaded whose FR3 did have those models.

Similar to #2934, this PR aims to solve the issue by spinning up a barebones version of the extractor during the custom level build to find any art groups that the level requires, going through all the official level DGOs (ignoring any duplicates), extracting the models and textures from them and adding them to the custom level's FR3 file.

To define all the models you want in your custom level, the level JSONC has a new entry:
```jsonc
  // All art groups you want to use in your custom level. Will add their models and corresponding textures to the FR3 file.
  // Note: You will still have to add them to your level's .gd file.
  "art_groups": ["plat-ag"]
  ```

I added a `plat` actor to `test-zone` to demonstrate the feature (and a test ambient as there wasn't one previously).

### DIFF
--- a/custom_levels/test-zone/test-zone.jsonc
+++ b/custom_levels/test-zone/test-zone.jsonc
@@ -34,7 +34,24 @@
    // The base actor id for your custom level. If you have multiple levels this should be unique!
    "base_id": 100,
 
-  "ambients": [],
+  // All art groups you want to use in your custom level. Will add their models and corresponding textures to the FR3 file.
+  // Note: You will still have to add them to your level's .gd file.
+  "art_groups": ["plat-ag"],
+
+  // Ambients you want to use in your custom level. Ambients can be used for various things like changing the music variation (flava),
+  // adding ambient sounds, level name hints, etc.
+  "ambients": [
+    {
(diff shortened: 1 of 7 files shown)


LuminarLight:
Hello, I am really happy about this, but wouldn't it be possible to get this information from the custom level's .gd file? I mean, the list of art groups that we want to be included. All art group assets end with '-ag', as far as I know.

Hat-Kid:
> Hello, I am really happy about this, but wouldn't it be possible to get this information from the custom level's .gd file? I mean, the list of art groups that we want to be included. All art group assets end with '-ag', as far as I know.

Yes, technically, this would be possible, but I wanted to avoid adding extra logic just to parse that file when we have the level JSONC already parsed and available during custom level building.

Hat-Kid:
The config settings for the `ObjectFileDB` are hardcoded for now, but after some testing (using black label), it seems like the version that is set does not have an impact here as only the tpages are processed. I made it so the extractor looks for the first folder in `iso_data` that has `jak1` in the name, which should hopefully cover the majority of cases.

PR #2942, opened 2023-08-28
entity-debug: Show path and volume data.

LuminarLight:
I consider this a minor quality-of-life change.

As the (now removed) code said, the paths and volumes can be easily displayed with help of debug menu, but there is no way to see the values there.

The values are quite readable if we just display them in the entity debug thing. It may be useful to people who are researching paths and volumes. I am not sure if it would be beneficial or not to display them as meters... all of these vectors will always have '1' or its negative, and that would look weird when divided by 4096. So I just left it like this.

I attached some images to prove that they look okay (see the 'path', 'vol', and 'visvol' tags):

jak1_2023-08-28T23-44-47
jak1_2023-08-28T23-45-40
jak2_2023-08-28T23-49-55
jak2_2023-08-28T23-49-02


ManDude:
I'm really not sure what it adds other than bloat. The `vol` (and `cutoutvol`) tags just contain a list of normals (but the W component is used as some sort of padding for the dot product, I think). The debug viewer draws a volume out of all the intersections (which is pretty crazy). This also means the volume must be concave.

The `visvol` is just an AB-box. Nothing you can't infer from using the viewer.

The `path` is just a list of coordinates. Nothing special here either.

LuminarLight:
For me personally, seeing examples of paths of actor was somewhat useful. Same goes for water vols. That is why I thought that maybe making the entity debugger show this data by default could potentially be useful for others in the future. But I can also understand that these aren't important/useful in most cases for most people.

Still, it is data that can be displayed. Why shouldn't we display it? The only downside is that there is more text on screen, and even then it can still fit on the screen in most cases. And when not, we can scroll.

Anyway, if the "bloat" is still a problem, would it be more acceptable if it was toggleable with a button?

ManDude:
> Why shouldn't we display it?

It is displayed, in a much more readable fashion with the instructions already given. This is just not useful information:
image

LuminarLight:
I have made the changes now. I feel like this is good enough, but please confirm.

xTVaser:
ok to test

xTVaser (approved):
looks good, new changes are properly behind a keybind

PR #2951, opened 2023-09-01
[jak2] higher resolution cloud textures

ManDude:
Adds a new "hires" sky texture animation that makes the sky use 512x512 textures instead of 128x128. The clouds have higher detail, are slightly more noisy, and are considerably less blocky. There is also a less noticeable "dot crawl" effect.

image
image


goal_src/jak2/engine/gfx/texture/texture-anim.gc
```diff
@@ -774,18 +799,41 @@ struct SlimeInput {
...
                                   )
        (pc-texture-anim-flag start-anim-array dma-buf)
-       (pc-texture-anim-flag clouds-and-fog dma-buf :qwc 4)
+       (pc-texture-anim-flag clouds-and-fog dma-buf :qwc (/ (psize-of sky-input) 16))
        (make-sky-input (the sky-input (-> dma-buf base)))
-       (&+! (-> dma-buf base) 64)
+       (&+! (-> dma-buf base) (psize-of sky-input))
        (pc-texture-anim-flag finish-anim-array dma-buf)
-       (dotimes (i 8) ;; intentially skipping fog here!!
+       (dotimes (i 9) ;; intentially skipping fog here!!
```
water111, line 806: This should actually be 8, and skip the last one (idx 8). The last one is the fog animation.

The code below this sets `anim-idx` to 8, then doesn't return, so the loop below updates the fog animation.

ManDude: oops, definitely didn't mean to change that.

PR #2954, opened 2023-09-02
g/j2: Some more work on the SQL editor - dump and seed `light` data

xTVaser:
image

Also did:
- Cleaned up the imgui menu
  - Moved "reboot in debug" to "Debugging" menu
  - Moved "Filters" from settings to tools
  - Hide the imgui hide keybind if you've enabled the setting to ignore it
- Automatically push content down in `*stdcon*` if the imgui bar is visible
- Automatically swap pad0 to pad1 when opening the editor and vise-versa.  There's a new imgui option to manually change this as well
- Support `append` `file-stream`s end-to-end

This only works for viewing data so far, nothing will save out to the SQL database yet, and with the edits i had to make, there's likely some major issues around editing planes and region boxes.


ManDude:
> there are many examples in the final game that aren't squares

Are there? I can't remember any region being both not a square, not a volume and not a square.

PR #2962, opened 2023-09-04
jak2: add frame rate option to graphics menu

Hat-Kid:
This PR adds a frame rate option to the graphics menu for some of the most common refresh rates.

Jak 2 has much better support for variable frame rates than Jak 1 out of the box, but there are still some edge cases, most prominently the fact that sprites are still limited to the 300 tick system, which is most noticeable on glow sprites. For this, I abused the glow boost debug setting to scale the glow based on the frame rate.

While testing, I noticed two other cases that I have also patched, there's likely to be many more that are yet to be found, but aside from that, the game is playable as normal.

https://github.com/open-goal/jak-project/assets/6624576/ad4db24f-cd27-4237-a155-0db7008160f3


BreakPoints:
I'll create a mega issue for Jak 2 once this is merged and keep things separate from TPL. From the limited testing I've done I've noticed that sound effects are a lot more likely to cut off at >60

goal_src/jak2/kernel/gkernel-h.gc
```diff
@@ -763,3 +763,11 @@
 (defmacro current-time ()
   `(-> PP clock frame-counter)
   )
+
+(defmacro seconds-per-frame ()
+  "Macro for assuming a 16.6 ms frame time at higher frame rates."
+  `(if (= (get-video-mode) 'custom)
+     0.016666668
+     (-> PP clock seconds-per-frame)
```
water111, line 771: This seems kinda weird, why would we set the frame time to 16.6 ms when running at a higher frame rate?

water111 (approved):
Looks good to me, though I don't entirely understand the `seconds-per-frame` macro. Maybe it's just working around a few places where the original developers forgot to scale by frame rate or something?

goal_src/jak2/kernel/gkernel-h.gc
Hat-Kid: I just adapted it from this particular patch: https://github.com/open-goal/jak-project/blob/master/goal_src/jak1/engine/target/target.gc#L813

I wasn't quite sure if this is the right move, but I did find another case where it fixed an issue (the jetboard spin).

PR #3001, opened 2023-09-17
[opengoal] make `none` a child of `object`

### LINKED ISSUE #830: [decompiler] `none` casts on event blocks
It's causing issues with `(event progress-normal)`

### LINKED ISSUE #928: [decomp] Forcing return value to none
- ice-cube
- balloonlurker
- target-handler

both have this issue where a function's return value is being forced to none and it seems non-obvious how to correct it.

### LINKED ISSUE #1703: [opengoal] relax type checking in state handlers
Return type for the `code`, `trans`, `post`, `enter` and `exit` handlers should be `object`. There should also be minimal type checking if the handler passed is not an immediate lambda (`behavior`).

ManDude:
Previously, `object` and `none` were both top-level types. This made decompilation rather messy as they have no LCA and resulted in a lot of variables coming out as type `none` which is very very wrong and additionally there were plenty of casts to `object`. This changes it so `none` becomes a child of `object` (it is still represented by `NullType` which remains unusable in compilation).

This change makes `object` the sole top-level type, and the type that can represent *any* GOAL object. I believe this matches the original GOAL built-in type structure. A function that has a return type of `object` can now return an integer or a `none` at the same time. However, keep in mind that the return value of `(none)` is still undefined, just as before. This also makes a cast to `object` meaningless in 90% of the situations it showed up in (as every single thing is already an `object`) and the decompiler will no longer emit them. Casts to `none` are also reduced. Yay!

Additionally, state handlers also don't get the final `(none)` printed out anymore. The return type of a state handler is completely meaningless outside the event handler (which is return type `object` anyway) so there are no limitations on what the last form needs to be. I did this instead of making them return `object` to trick the decompiler into not trying to output a variable to be used as a return value (internally, in the decompiler they still have return type `none`, but they have `object` elsewhere).

Fixes #1703 
Fixes #830 
Fixes #928


decompiler/analysis/expression_build.cpp
```diff
@@ -91,21 +91,31 @@ bool convert_to_expressions(
...
-            break;
           }
-        }
 
-        if (!found_early_return && f.type.last_arg() != return_type) {
-          needs_cast = true;
+          // the return value of this function is not an exact match
+          // we cast it to avoid complicated issues with polymorphism (e.g. methods)
+          // we don't run this if we find a (return statement because we do not handle
+          // type checking on those at the moment.
```
water111, line 115: I'm pretty sure we do: https://github.com/open-goal/jak-project/blob/697b07abd523f2b3efc793596035f318e390aae4/goalc/compiler/compilation/Block.cpp#L133

water111: that said, if this is passing all tests I wouldn't worry about it. I'm not convinced we actually had this logic right originally and this is not any worse.

water111 (approved):
Thanks for doing this, it's a huge improvement on the decompiled code for states.

I'm a little worried that this will make it easier to have subtle "use of `none`" bugs when decompiling new code. The reason the original thing was so strict and annoying was to try to make it hard to produce decompiled code that was incorrect.  That said, I'm okay with this because:
- we've already decompiled jak 2, and I don't think there's a a way for these changes to add any new bugs. If anything it probably removes possible bugs by getting rid of `(the-as none` casts.
- I'm not even sure we're doing jak 3, and if we do, we can probably improve the decompiler's warnings around use of `none` type values.
- we could detect invalid use of `(none)` by modifying the compiler to insert some bogus value for `(none)` instead.

common/type_system/TypeSystem.cpp
```diff
@@ -1715,10 +1715,6 @@ std::string TypeSystem::lca_base(const std::string& a, const std::string& b) con
     return a;
   }
 
-  if (a == "none" || b == "none") {
-    return "none";
```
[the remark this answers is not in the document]
ManDude: For that second one, I don't think we have ever handled that case correctly anyway. See these two functions:
```lisp
(defun i-want-uint128 ((test symbol))
(declare (print-asm))
(if test
    0
    (-> (new 'static 'qword :dword (new 'static 'array uint64 2 123 456)) quad)))
```

```lisp
(defun i-want-uint128-2 ((test symbol))
(declare (print-asm))
(if test
    (none)
    (-> (new 'static 'qword :dword (new 'static 'array uint64 2 123 456)) quad)))
```

The first one is type `(function symbol int)` and the second one is type `(function symbol none)`. In both functions, the asm for the `uint128` case looks like this:
```
  [0x10014] lea r9, [0x000000000001001B] mov-sva igpr-7, static-structure
  [0x1001B] sub r9, r15
  [0x1001E] vmovaps xmm7, [r15+r9*1]               mov ii128-6, [igpr-7 + 0]
  [0x10024] movq rax, xmm7 mov igpr-3, ii128-6
```
```
  [0x10011] lea r9, [0x0000000000010018] mov-sva igpr-6, static-structure
  [0x10018] sub r9, r15
  [0x1001B] vmovaps xmm7, [r15+r9*1]               mov ii128-5, [igpr-6 + 0]
  [0x10021] movq rax, xmm7 mov igpr-3, ii128-5
```
so the upper 64-bits are being dropped.

Compare to this function that's type `(function uint128)`: 
```
(defun i-want-uint128-3 ()
(declare (print-asm))
(-> (new 'static 'qword :dword (new 'static 'array uint64 2 123 456)) quad))
[i-want-uint128-3]

Program string:3
-> (-> (new 'static 'qword :dword (new 'static 'array uint64 2 123 456)) quad))
    ^
- [0x10000] lea r9, [0x0000000000010007] mov-sva igpr-2, static-structure
  [0x10007] sub r9, r15
  [0x1000A] vmovaps xmm0, [r15+r9*1]               mov ii128-1, [igpr-2 + 0]
  [0x10010] ret
```

PR #3003, opened 2023-09-18
g/j2: Fill out `game options` menu

xTVaser:
https://github.com/open-goal/jak-project/assets/13153231/b103963e-7fad-4578-b1ae-5042dbdccc99


xTVaser:
`preserve-this` removal is intentional, it was within a custom file, something that will never be automatically removed by decompilation.

goal_src/jak2/pc/progress/progress-static-pc.gc
```diff
@@ -55,18 +294,40 @@
...
+          (commit-to-file *pc-settings*))))))
 
 
+;; TODO - migrate this menu as well
+;; graphic options
+;;   resolution
+;;   display mode
+;;   display select (disable if windowed)
+;;   vsync
+;;   aspect ratio
```
water111, line 322: I think some of these are already there (vsync, full screen)

water111 (commented):
Looks good to me!

One minor thing I found during testing is that some of the text is cut off when using 16x9 aspect ratio:
image

goal_src/jak2/pc/progress/progress-static-pc.gc
xTVaser: Yeah, this is just the full jak 1 menu layout for reference.

xTVaser:
You're using the non ps2 aspect ratio I assume?  I havn't done a ton of testing outside the ps2 aspect ratios and I had some issues using `adjust-game-x` for that menu, and the sliders so I'm not surprised.

PR #3013, opened 2023-09-23
g/j2: speed up level-select menu scrolling

xTVaser:


goal_src/jak2/engine/ui/progress/progress.gc
```diff
@@ -2880,6 +2880,42 @@
            )
          )
        )
+      ;; og:preserve-this - Added to make scrolling these menus faster (jak3 has this)
+      ((or (cpad-pressed? 0 left) (and (cpad-hold? 0 left) (>= (- (current-time) (-> obj last-move)) (seconds 0.075))))
```
ManDude, line 2884: Might want `l-analog-left` too? (and `l-analog-right` later)

xTVaser: I avoided those intentionally, analog controls are less precise and I wouldn't want someone to accidentally hit left/right and something unexpected happen.  Though if jak 3's menus allow for it, it would make sense to add it here.

PR #3014, opened 2023-09-23
improve decomp of state handlers and art groups

ManDude:
- state handlers that are not inlined lambdas have smarter type checking, getting rid of 99.9% of the casts emitted (they were not useful)
- art groups were not being properly linked to their "master" groups.
- `max` in `ja` in Jak 2 was not being detected.

Another huge PR...


ManDude:
removed `preserve-this` is intentional.

common/type_system/TypeSystem.cpp
```diff
@@ -1590,7 +1590,7 @@ bool TypeSystem::typecheck_and_throw(const TypeSpec& expected,
       if (!got) {
         success = false;
       } else {
-        if (*got != tag.value) {
+        if (!tc(tag.value, *got)) {
```
water111, line 1593: I really don't remember why it was like this in the first place... I can't think of any reason why. At some point maybe I thought that tag values wouldn't always be types? Either way I think this is a good change.

It also looks like tag stuff is just missing from the `TypeSystem::lowest_common_ancestor` function (it will just strip away tags right now?)

decompiler/IR2/FormExpressionAnalysis.cpp
```diff
@@ -3881,7 +3881,7 @@ void FunctionCallElement::update_from_stack(const Env& env,
         macro.push_back(rate);
       }
 
-      if (env.version > GameVersion::Jak1) {
+      if (env.version != GameVersion::Jak1) {
```
water111, line 3884: but what about jak 3? :)

decompiler/analysis/expression_build.cpp
```diff
@@ -111,8 +111,8 @@ bool convert_to_expressions(
 
           // the return value of this function is not an exact match
           // we cast it to avoid complicated issues with polymorphism (e.g. methods)
-          // we don't run this if we find a (return statement because we do not handle
-          // type checking on those at the moment.
+          // we don't run this if we find a (return statement for unknown reasons
+          // TODO : remove this and check?
```
water111, line 115: yeah I totally forget about this too.

goal_src/jak2/engine/process-drawable/process-drawable.gc
```diff
@@ -603,8 +603,6 @@
       )
     )
   :code (behavior ((arg0 string))
-    ;; og:preserve-this
-    (format 0 "ART ERROR: ~A for ~A~%" arg0 self)
```
water111, line 607: Do you think any of the remaining art errors are worth looking into? I think a lot of them are original game errors at this point.

water111 (approved):
Looks good! Thanks for the cleanup!

goal_src/jak2/engine/process-drawable/process-drawable.gc
ManDude: Nope. They're either missing art data from the original game (unfixable without some new hacks) or extremely rare GOAL bugs.

PR #3041, opened 2023-09-30
fix draw-string-xy-scaled cropping

### LINKED ISSUE #3038: Jak 1 `draw-string-xy-scaled` is buggy
### Acknowledgements

### Describe the Bug

The `draw-string-xy-scaled` function I added in #1756 doesn't quite work properly - it does scale the text, but in doing so it scales the available screen space that the text can be drawn to. So say your scale is 0.5, then any text in the lower and/or right half of the screen would get cut off.

Spoke to Vaser a bit about this and he suggested replacing with `draw-string-adv` instead, setting the scale in the `font-context`, but the issue is that `draw-string-adv` (really `draw-string` itself) doesn't do anything with the `font-context`'s scale, at least in [Jak 1 where we use the GOAL implementation](https://github.com/open-goal/jak-project/blob/1f4044b9ffcd0836e33f2ab68ad1563891ab8016/goal_src/jak1/engine/gfx/font.gc#L610-L615)

In Jak 2, I believe it's using the [mips2c implementation of `draw-string`](https://github.com/open-goal/jak-project/blob/master/game/mips2c/jak2_functions/font.cpp), and apparently this does respect the font context - maybe this is a bug in the Jak 1 GOAL implementation?

### How To Reproduce

`draw-string-xy-scaled` with a scale like 0.5 and you won't be able to draw text to the right/bottom of screen

Yes, it's unique to OpenGOAL

### Expected Behavior

`draw-string` (whether through a helper function like `draw-string-xy-scaled` or otherwise) should support scaling text properly to all areas of the screen

### Environment Information

All environments

### Game Version

NTSC 1.0 (black label)

No

dallmeyer:
fixes #3038 

https://github.com/open-goal/jak-project/assets/2515356/5b7a490f-e1fb-4a12-a3d9-b576a2c1d574


goal_src/jak1/engine/gfx/font.gc
```diff
@@ -1784,36 +1784,31 @@
...
-    (let ((scaled-width (* (-> font-ctxt width) scale))
-          (scaled-height (* (-> font-ctxt height) scale))
-          (orig-mat-x-scale (-> font-ctxt mat vector 0 x))
-          (orig-mat-y-scale (-> font-ctxt mat vector 1 y))
-          (orig-vid-parms-x-scale (-> *video-parms* relative-x-scale))
-          (orig-vid-parms-y-scale (-> *video-parms* relative-y-scale))
-          )
-      (*! (-> font-ctxt scale) scale)
-      (set! (-> font-ctxt width) scaled-width)
-      (set! (-> font-ctxt height) scaled-height)
```
dallmeyer, line 1797: this was causing the cropping

goal_src/jak1/engine/gfx/font.gc
```diff
@@ -1784,36 +1784,31 @@
...
   "Draw a string at the given xy location, with the given scale."
-  (let ((font-ctxt (new 'stack 'font-context *font-default-matrix* x y 0.0 color flags)))
-    (let ((scaled-width (* (-> font-ctxt width) scale))
-          (scaled-height (* (-> font-ctxt height) scale))
-          (orig-mat-x-scale (-> font-ctxt mat vector 0 x))
-          (orig-mat-y-scale (-> font-ctxt mat vector 1 y))
-          (orig-vid-parms-x-scale (-> *video-parms* relative-x-scale))
-          (orig-vid-parms-y-scale (-> *video-parms* relative-y-scale))
-          )
-      (*! (-> font-ctxt scale) scale)
```
dallmeyer, line 1795: `draw-string` doesn't seem to care about the `font-context` scale

water111 (approved):
Looks good to me, but @ManDude is the font expert so I will let him double check.

One thing worth double checking is making the aspect ratio _very_ wide and make sure the text shadow doesn't do something weird.

PR #3042, opened 2023-09-30
[jak2] Set up extractor

water111:
This sets up the extractor for jak 2. I was expecting that I'd have to make some more significant changes to the decompiler/compiler path stuff, but this was not the case!

The only real change is that you can now provide multiple ISO hashes for an entry in `ISOMetadata`. This is needed for the two different NTSC versions, which have the same configs, serials, and ELF hashes, but slightly different contents.

I also didn't add the korean version because I don't have the info for it.


ManDude:
Could you not have also added two entries that are the same except for the hash? Or does the extractor not work like that?

water111:
It doesn't work like that - it only does the lookup based on the ELF name and ELF hash, which are the same for the two versions.

We could definitely do something fancier for the lookup, but it became a mess to implement (`build_level.cpp` looks at this stuff, we'd have to change the build_info metadata stuff, ...).  I _think_ this will work because there is really no difference between ntsc_v1/ntsc_v2 for extraction.

water111:
For more context about the most recent commit - the way that we were building and linking the `extractor_utils.cpp` wasn't quite right - which causes it (at least on linux) to segfault because the same .cpp is built twice and linked together into the final library.

I think it's better to just avoid sharing globals across library boundaries (and avoid globals with non-trivial dtors in the first place). And when you must, it's better to do something like this: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Construct_On_First_Use  (which, conveniently, avoids the need for declspec nonsense on windows, which has different behavior from linux when it comes to exporting symbols from shared libraries)

xTVaser:
Thanks for fixing that, I ran into issues when I split that code apart and I could replicate/figure out the issue on my end.

PR #3057, opened 2023-10-02
Add Yopie input display

Zedb0T:
Need advice on where to put the check for
```
 (if (-> *pc-settings* input-display?)
        (input-display-on)
        (input-display-off)
        )
```

This adds the input display that is in the mod base to the main project, this setting is added to the pc-settings and persisted, it is also added to the debug menu and the pause menu.


Zedb0T:
[https://youtu.be/pxZqqTFjFTY](https://youtu.be/pxZqqTFjFTY)

goal_src/jak2/engine/ui/text-id-h.gc
```diff
@@ -719,6 +719,11 @@
   (progress-hires-sky #x1309)
   (progress-fast-airlock #x130a)
   (progress-fast-elevator #x130b)
+  (progress-toggle-input #x130c)
+  (pad-triangle #x2000)
+  (pad-circle #x2001)
+  (pad-x #x2002)
+  (pad-square #x2003)
```
Hat-Kid, line 726: Add these entries to the enum in `all-types.gc` so the type consistency test passes.

Hat-Kid (changes requested):

Hat-Kid:
> Need advice on where to put the check

I would probably put the check in the `pc-settings-jak2::update` method.

Zedb0T:
Thanks, I've made the requested changes locally.

One thing before a push that I dont fully understand
`(with-dma-buffer-add-bucket ((testbuf (-> (current-frame) global-buf)) (bucket-id subtitle))` 

Would there be a better bucket-id for this? Or is `subtitle` sufficient?

goal_src/jak2/engine/gfx/font.gc
```diff
@@ -209,6 +209,19 @@
   (none)
   )
 
+;; Added for PC port
+(defun draw-string-xy-scaled ((str string) (buf dma-buffer) (x int) (y int) (color font-color) (flags font-flags) (scale float))
+  "Draw a string at the given xy location, with the given scale."
+  (let ((font-ctxt (new 'stack 'font-context *font-default-matrix* x y 0.0 color flags)))
+    (*! (-> font-ctxt scale) scale)
+    (*! (-> font-ctxt width) scale)
+    (*! (-> font-ctxt height) scale)
```
ManDude, line 218: You most certainly do not want to modify the height and width of the text box.

ManDude (changes requested):

goal_src/jak2/pc/pckernel.gc
```diff
@@ -230,6 +230,11 @@
   (when (not (led-enabled? obj))
     (disable *led-fader-state*)
     )
+  
+   (if (-> *pc-settings* input-display?)
+        (input-display-on)
+        (input-display-off)
```
ManDude, line 236: Where are these functions defined?

ManDude (dismissed):

goal_src/jak2/pc/pckernel.gc
Zedb0T: https://github.com/open-goal/jak-project/pull/3057/files#diff-80370092254773d4b5c4f8774df3cd686c9a1dad5528f790c727450d8d58f2d0R324

https://github.com/open-goal/jak-project/pull/3057/files#diff-80370092254773d4b5c4f8774df3cd686c9a1dad5528f790c727450d8d58f2d0R370

Zedb0T: Well that was less helpful than I wanted it to be, but they are in goal_src/jak2/pc/features/input-display.gc

ManDude: Oops they were hidden, thanks github.

xTVaser, line 236: I don't think this should be running once per frame, i'm not really concerned about performance problems but it's just not the idiomatic location for it.  It should only need to run when the user actually changes the setting as far as I can tell.

goal_src/jak2/pc/features/input-display.gc
xTVaser, line 1: doesn't have to be done before merging, but jak 2 adds a lot to the font file in terms of buttons, so this can be improved quite a bit.  Additionally, it unfortunately omits a few buttons (no "start"), but I've had an idea on how to solve that which I'll probably try soon.

xTVaser (changes requested):

goal_src/jak2/pc/pckernel.gc
Zedb0T: tbh that makes so much sense

Zedb0T: So, after sleeping on it, the process still needs to be spun up on initialization no?

xTVaser:

PR #3083, opened 2023-10-11
decompiler: add texture merging feature

xTVaser:
This adds a new feature to the decompiler to complement texture _replacement_ with what I'm calling texture _merging_.

Instead of taking a PNG and using it's data to fully replace an existing texture, we instead only replace the pixels that are non-transparent from the merge image.

This might be useful in a few other situations, **but what I'm interested in using it for is expanding the game's font without having to commit any of the existing texture assets, just merging in individual new characters.**  

For Jak 1, there isn't much empty room available and I'm not sure how easy it would be to support larger font-textures.  But in Jak 2, across all `font.24*` textures, there are over 200 unused character slots.

In some ways, this is an interim workaround because adding brand new textures to the game is non-trivial (ie. adding entire new font textures would allow for unlimited characters!).

### DIFF
--- a/decompiler/main.cpp
+++ b/decompiler/main.cpp
@@ -298,6 +298,13 @@ int main(int argc, char** argv) {
 
   mem_log("After textures: {} MB", get_peak_rss() / (1024 * 1024));
 
+  // Merge textures before replacing them, in other words, replacements take priority
+  auto texture_merge_path = file_util::get_jak_project_dir() / "game" / "assets" /
+                            game_version_names[config.game_version] / "texture_merges";
+  if (fs::exists(texture_merge_path)) {
+    tex_db.merge_textures(texture_merge_path);
+  }
+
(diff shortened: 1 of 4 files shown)


decompiler/main.cpp
```diff
@@ -298,6 +298,13 @@ int main(int argc, char** argv) {
 
   mem_log("After textures: {} MB", get_peak_rss() / (1024 * 1024));
 
+  // Merge textures before replacing them, in other words, replacements take priority
+  auto texture_merge_path = file_util::get_jak_project_dir() / "game" / "assets" /
+                            game_version_names[config.game_version] / "texture_merges";
+  if (fs::exists(texture_merge_path)) {
+    tex_db.merge_textures(texture_merge_path);
+  }
+
```
water111, line 307: you might need to add this to the extractor too

water111 (approved):

PR #3086, opened 2023-10-14
Add label_types file merge for PAL version

jabermony:
This was added as the progress file's L879 has a different type in the PAL
 version of the game: being 'progress-global-state' rather than 'unit64'.
The difference caused the decompiler to die with a mismatched type error.


xTVaser:
ok to test

xTVaser (approved):

PR #3098, opened 2023-10-22
Jak3 Build Environment

jabermony:
This sets out the bones of a Jak 3 build, many things are stubbed out, guessed, or copied from Jak 2 but it should at least be good enough to:
run `task set-game-jak3`
launch the repl
run builds from the repl

build outputs themselves are untested but the build itself runs without errors


jabermony:
I've pushed a change for most version == GameVersion::Jak2  checks to make them  version >= GameVersion::Jak2 in decompiler\
I've left any switch statements with a default case that asserts an error untouched as those are probably better done as/when they come up as they can be properly checked and handled when they come up.

iliasdevpas:
with this i can install jak 3 on open goal?

decompiler/IR2/AtomicOpTypeAnalysis.cpp
```diff
@@ -662,7 +662,7 @@ TP_Type SimpleExpression::get_type_int2(const TypeState& input,
     }
   }
 
-  if (env.version == GameVersion::Jak2 && tc(dts, TypeSpec("symbol"), arg1_type) &&
+  if (env.version >= GameVersion::Jak2 && tc(dts, TypeSpec("symbol"), arg1_type) &&
```
water111, line 665: I think this one should be reverted. String->symbol is different and shouldn't use the jak2's constant.

xTVaser:
add to whitelist

PR #3128, opened 2023-11-02
Add fallback text functionality to Jak 2

### LINKED ISSUE #3118: fallback to english if a text-id is not translated

Zedb0T:
It works now so I figured Id open it for input on the final remaining things.

Things Im not satisfied with

*fallback-text-lookup?* being defined in pc-kernel-impl.gc, Isnt there a way to just add this to pc-kernel-common??

I would like a close double check on the logic on lines 198-207 I'm pretty sure all the bases are still covered and it will never not find anything but I just want a double check

Fixes #3118

### DIFF
--- a/goal_src/jak1/engine/ui/text.gc
+++ b/goal_src/jak1/engine/ui/text.gc
@@ -242,7 +242,7 @@
     (load-game-text-info "common" '*common-text* *common-text-heap*)
     (#when PC_PORT
       (protect ((-> *pc-settings* text-language))
-        (set! (-> *pc-settings* text-language) (the language-enum (pc-language english)))
+        (set! (-> *pc-settings* text-language)  (pc-language english))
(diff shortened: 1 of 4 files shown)


Zedb0T:
#3118

Zedb0T:
Would it be worth changing `(defun load-game-text-info ((arg0 string) (arg1 (pointer object)) (arg2 kheap))` to just take in a symbol instead of duplicating a entire hardcoded copy that does English??

ManDude:
You don't actually need to define a separate function at all. Since the language picked is based on some variable in the settings, you can `protect` around a function call where you change the language right before calling.

Zedb0T:
Wow `Protect` is so cool wtf I've been backing up all of my settings for my mod then doing mod stuff then restoring them. 
Thanks!

I used Protect instead and removed the duplication from both jak1 and jak2

xTVaser (approved):

PR #3130, opened 2023-11-02
[jak2] Hopefully improve sky performance

water111:
Switches the slime look up table to be a texture, since I guess intel drivers are terrible and putting the array in the shader makes it extremely slow.

Also, a few minor changes:
- removed art-groups from the test-zone levels since this causes the compiler to re-decompile the game, and makes the launcher slower. (left it in commented out)
- Switched `decompile_code` to false by default in jak 2, in case people run the decompiler and don't want to wait forever
- Fixed build warnings


dallmeyer:
Tested on my Intel UHD 770 @ 1080p and it's a clear improvement, made things playable again for me.

Without the fix I'd get 60fps in prison, but terrible fps anywhere outdoors with sky. With the fix I get 60fps everywhere.

PR #3143, opened 2023-11-04
[jak2] Pass 0x01 through format

water111:
fixes the crash when this message is displayed:

image

as far as I can tell, it looks right.

### DIFF
--- a/game/kernel/jak2/kprint.cpp
+++ b/game/kernel/jak2/kprint.cpp
@@ -229,6 +229,7 @@ s32 format_impl_jak2(uint64_t* args) {
         case 'w':
         case 'y':
         case 'z':
+        case 1: // jak 2 japanese encoding
           while (arg_start < format_ptr + 1) {
             *output_ptr = *arg_start;
             arg_start++;
--- a/game/kernel/jak2/kprint.cpp
+++ b/game/kernel/jak2/kprint.cpp
@@ -229,7 +229,7 @@ s32 format_impl_jak2(uint64_t* args) {
         case 'w':
         case 'y':
         case 'z':
-        case 1: // jak 2 japanese encoding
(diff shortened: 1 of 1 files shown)


ManDude:
I think our default case in `format` is wrong. Our version:

```cpp
        default:
          MsgErr("format: unknown code 0x%02x\n", format_ptr[1]);
          MsgErr("input was %s\n", format_cstring);
          ASSERT(false);
          break;
```
the ghidra output is:
```cpp
switchD_00101b6c_caseD_26:
      MsgErr__3ndiPCce(0x12ce48,bVar4);
      goto LAB_0010230c;
```
```cpp
LAB_0010230c:
    pbVar15 = pbVar15 + 1;
LAB_00102310:
    bVar4 = *pbVar15;
    bVar11 = *pbVar15;
    pbVar10 = pbVar15;
    puVar3 = local_1b0;
```
which is closer to:
```cpp
default:
  MsgErr("format: unknown code 0x%02x\n", format_ptr[1]);
  *output_ptr = *format_ptr;
  output_ptr++;
  format_ptr++;
  break;
```

We have this branch:
```cpp
    } else {
      // got normal char, just copy it
      *output_ptr = *format_ptr;
      output_ptr++;
    }
```
that we should probably just `goto` to in this case.

ManDude:
in other words, I think something like this is more accurate to the original code (they either had a `goto` or a completely different control flow in this function), less annoying, and also functional:
```cpp
        default:
          MsgErr("format: unknown code 0x%02x\n", format_ptr[1]);
          // MsgErr("input was %s\n", format_cstring);
          goto copy_char_case;
      }
      format_ptr++;
    } else {
copy_char_case:
      // got normal char, just copy it
      *output_ptr = *format_ptr;
      output_ptr++;
    }
    format_ptr++;
``

PR #3161, opened 2023-11-05
[jak2] Localize board trick names

ManDude:
It also looks like this now which I think is a bit cleaner:

image


goal_src/jak2/engine/target/board/target-board.gc
```diff
@@ -53,112 +53,66 @@
...
+          (let ((times (-> freq trick))
+                (color (case trick
+                          (((board-tricks board-boost) (board-tricks board-jump) (board-tricks board-duck-jump) (board-tricks board-quick-jump))
+                          "~9L")
+                          (((board-tricks board-rail) (board-tricks board-rail-jump))
+                          "~3L")
+                          (((board-tricks board-spin) (board-tricks board-flip) (board-tricks board-board-flip) (board-tricks board-nosegrab))
+                          "~26L")
+                          (((board-tricks board-board-spin) (board-tricks board-method) (board-tricks board-noseflip) (board-tricks board-kickflip))
+                          "~15L")))
```
xTVaser, line 93: can you revert back to using the color codes instead as i did in my most recent PR? It makes the code much more readable

PR #3178, opened 2023-11-07
[high fps] Increase input buffer size and fix cloud scroll speed

Brent-Hickey:
The game stores the last 3 frames of input (50ms of time at 60fps) and often checks if a button was pressed within those 3 saved frames as a condition. 

When transitioning states, it often checks if some input was received during the previous state (within those 3 frames) in order to quickly transition out. A good example of this is when transitioning to standing, it checks if you can jump this frame and if you had pressed X recently, and if so, transition immediately to jump. This allows transitions between states to feel more smooth/forgiving by letting you jump at a later time when you are transitioning from falling->standing than if you were only falling.

At 165fps the last 3 frames is only 18ms of time so the input windows for these smooth transitions are almost 3x shorter. 

This PR saves 15 input frames (enough to cover 50ms of time at 300fps) for each controller and adds a (recently-pressed?) macro that checks all 15 frames. However, it only updates the necessary frames in history based on the current frame rate. This way, 60fps continues to only check against 3 input frames, 165fps checks against 9, 240fps checks against 12, and 300fps checks all 15.


NineKain:
Is this still alive?

water111 (commented):
Sorry this took me so long to get to.

This looks good to me and I like this approach. Do you mind adding `;; og:preserve-this` comments to the changes in the `target-` files too?

PR #3185, opened 2023-11-11
989snd: sound bank loading overhaul

### LINKED ISSUE #1417: [music] unique soundbanks get messed up if they have non-unique names

Ziemas:
I made the mistake of trying to rewrite all of the 989snd code in one go and quickly lost motivation. Starting to bring over the stuff that was actually done.

This is very likely still buggy, so needs testing and work.

Should probably fix #1417


ManDude:
Isn't `std::span` C++20? I think we use C++17 for this project.

xTVaser (approved):
Pulled it down and noticed no noticeable issues, formatting can be fixed on whatever PR comes in next.

PR #3191, opened 2023-11-11
[jak2] Fix "Select controller"

### LINKED ISSUE #3129: `Select controller` is not working
> If I have more than one controller connected, in the input options, the "Select controller" option does nothing. When I select another controller, the setting returns to what it was before.

Trying to change controller the "dirty way" (that is, unplugging the controller I don't want to use, so that the one I want to use is the only one that appears) makes it work.

SuperSamus:
Fixes #3129.

I can't test the last commit, and because this is my first contribution here, I may have made a mistake.

### DIFF
--- a/game/kernel/common/kmachine.cpp
+++ b/game/kernel/common/kmachine.cpp
@@ -664,6 +664,13 @@ u64 pc_get_controller_count() {
   return 0;
 }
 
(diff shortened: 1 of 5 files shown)


xTVaser:
ok to test

xTVaser (approved):

PR #3193, opened 2023-11-11
Update to C++20

Ziemas:
Just putting this here for consideration, I'm personally not in a big rush to get it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,7 +219,6 @@ build_third_party_lib(sqlite3 sqlite3)
 
 # build tree-sitter parser
 include_directories(third-party/tree-sitter/tree-sitter/lib/include)
-include_directories(third-party/tree-sitter/tree-sitter-opengoal/include)
 build_third_party_lib(tree-sitter tree-sitter)
 
 # native OS dialogs for error messages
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -189,7 +188,6 @@ include_directories(SYSTEM third-party/inja)
 # Enable SSL Support, most URLs now-a-days use SSL!
 # TODO - probably integrate with ZSTD since we have it already
 set(CURL_USE_LIBSSH2 OFF)
-set(CURL_ZLIB OFF)
```
xTVaser, line 192: curious if you found a reason this wasn't needed, I unfortunately can't remember why i added it -- but i know it was when i was troubleshooting why the binaries were expecting some libraries to be dynamically linked

CMakeLists.txt
```diff
@@ -219,7 +217,6 @@ build_third_party_lib(sqlite3 sqlite3)
 
 # build tree-sitter parser
 include_directories(third-party/tree-sitter/tree-sitter/lib/include)
-include_directories(third-party/tree-sitter/tree-sitter-opengoal/include)
```
xTVaser, line 222: similarly curious, not actually needed?

Ziemas: That path doesn't exits, so I was getting spammed with warnings for it.

CMakeLists.txt
Ziemas: I was getting linker errors with this option set. But I can't reproduce them now :thinking:

PR #3195, opened 2023-11-12
Add jak 2 all open orbs speedrun category

Zedb0T:


goal_src/jak2/pc/features/speedruns.gc
```diff
@@ -20,6 +19,15 @@
       (initialize! *game-info* 'game (the-as game-save #f) "game-start"))
     (((speedrun-category newgame-heromode))
       (initialize! *game-info* 'game (the-as game-save #f) "game-start-hero"))
+    (((speedrun-category newgame-allopenorbs))
+      (set! (-> *speedrun-menu* 0 draw-menu?) #f)
+      (set-master-mode 'game)
```
xTVaser, line 24: hiding the menu here shouldn't be required, none of the other categories require that?  The menu automatically hides when hitting "reset".  I believe that should also change the master-mode as well.

xTVaser: Ah I understand, kk well as long as it works then maybe it can be cleaned up later.

xTVaser (changes requested):
This has a problem, the run info isn't displayed because right now it is controlled by fortress escape being completed (since you start in the nest with tasks unlocked, its immediately hidden).  We should break out that criteria into a separate function on a per-category basis.

Zedb0T:
> This has a problem, the run/verification info isn't displayed because right now it is controlled by fortress escape being completed (since you start in the nest with tasks unlocked, its immediately hidden). We should break out that criteria into a separate function on a per-category basis.
You fine with me birthing a process that displays the info for the first 15 seconds of the run instead?

xTVaser:

PR #3199, opened 2023-11-12
i18n: transcribe missing english subtitles

xTVaser:
Working through transcribing the remaining subtitle lines into English, I will then add them to Crowdin.

Issues noticed with the subtitle editor as I do this:
- deleting a non-cutscene after making it is a crash
- saving a new cutscene doesnt play properly (or atleast consistently), function that runs it returns #f
- making a new scene and selecting it doesnt populate the "currently selected scene" label properly
- circle does not reset a voice line like the UI implies
- it'd be nice when clicking "add text entry" if it updated your old starting frame to your old ending frame (for the next line) similarly, when changing scenes if the fields reset (leaving speaker alone is good though).  Separate buttons for these are likely required.
- selecting Samos for some reason doesn't want to switch speakers


game/assets/jak2/subtitle/subtitle_lines_en-US.json
```diff
@@ -93,6 +93,369 @@
...
+      "Stay sharp, we got a tough part of our town coming up."
+    ],
+    "agnt065": [
+      "Now we're in for it."
+    ],
+    "agnt066": [
+      "You're wasting citizens!"
+    ],
+    "agnt067": [
+      "Don't hit the civvies man!"
```
Hat-Kid, line 253: "Don't hit the civvies, man!"

ManDude (approved):

PR #3230, opened 2023-11-28
Toggle Cpp openGL with F12

Zedb0T:


Zedb0T:
Current on hold, as I dont see where the `Roaming\OpenGOAL\jak1\misc\debug-settings.json` gets initialized

Zedb0T:
Yeah neverminded, good for review let me know if any adjustments need to be made.

Zedb0T:
Is there other industry standard safety precautions I should do here?

Thoughts to consider
- windowed resolution on toggle
- windowed location on toggle

ManDude:
F12 tends to mess with debuggers. Any reason you didn't go for Alt+Enter?

Zedb0T:
None outside of personal preference I'll adjust to alt enter at end of day.

Zedb0T:
It was surprisingly more difficult/annoying to change to alt+enter than I thought it would be, but I have it working locally. I think this is worse because holding alt causes the imgui bar to flash and end in a unpredictable state, and enter pauses the game so it feels kind of clunky to me. 

I still have quite a bit of cleanup to do on it, but just wanted to see if you still think alt+enter is a better combination than f12. I am fine moving forward with whatever you think/prefer. @ManDude.

xTVaser:

Zedb0T:
This PR can be considered dead,continued over here -> #3901

PR #3263, opened 2023-12-19
g/j2: save window-size to pc-settings when it's changed

### LINKED ISSUE #3262: Window size setting is not saved to pc-settings

xTVaser:
Fixes #3262


ManDude:
Oops.

PR #3266, opened 2023-12-20
[wip] build actor tool

Hat-Kid:
This does a couple of things:

- The `custom_levels` folder was renamed to `custom_assets` and contains `levels`, `models` and `texture_replacements` folders for Jak 1, 2 and 3 in order to keep everything regarding custom stuff in one place.
	- With this, texture replacements now use separate folders for all games
- A build actor tool was added that generates art groups for custom actors
- Custom levels can now specify what custom models from the `models` folder they want to import, this will add them to the level's FR3.
- A `test-zone-obs.gc` file was added, containing a `test-actor` process that uses a custom model as an example.

The build actor tool is still very WIP, the joints and the default animation are hardcoded, but it allows for importing any GLB file as a merc model.


dallmeyer:
we should update docs to go with this update, started a draft: https://github.com/open-goal/open-goal.github.io/pull/210

PR #3287, opened 2024-01-01
j2-overlord: Clean up some VAG functions

Ziemas:
Some cleanup of vag related functions in overlord.

### DIFF
--- a/game/overlord/jak1/ssound.cpp
+++ b/game/overlord/jak1/ssound.cpp
@@ -54,13 +54,13 @@ void InitSound_Overlord() {
   snd_StartSoundSystem();
   snd_RegisterIOPMemAllocator(SndMemAlloc, SndMemFree);
   snd_LockVoiceAllocator(1);
-  u32 voice = snd_ExternVoiceAlloc(2, 0x7f);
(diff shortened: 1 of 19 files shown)


Ziemas:
I was about to do more here, but it's probably for the best if split stuff up in case there's regressions.

Help with testing would be appreciated.

ManDude:
Can you also rename `CalculateFallofVolume` to `CalculateFalloffVolume` and `gCurve` to `gCurves` as per https://github.com/open-goal/jak-project/pull/3288/commits/354f3a2f81b493364f282c60aa05b888de64b689 ?

game/overlord/jak1/ssound.cpp
```diff
@@ -115,6 +115,32 @@ void InitSound_Overlord() {
   }
 }
 
+void UpdateLocation(Sound* sound) {
+  if (sound->id == 0) {
+    return;
+  }
+
+  if (g_game_version == GameVersion::Jak1) {
```
ManDude, line 123: This can be removed now I think.

ManDude (approved):
Didn't notice any regressions.

PR #3290, opened 2024-01-03
goalc: Some more MacOS ARM64 work

xTVaser:
Not a whole lot in this PR, but I want to get this off my laptop where it's been sitting for a few months.

Started to split apart the implementation for different architectures and get CodeTester tests passing.   At one point I proved to myself that I got it working but then things fell apart when i began the grueling refactor.


water111 (commented):
Nice, I think this is a good start.

My main concern at this point is that ARM is actually quite a bit different from x86, so it will be hard to fit the ARM instructions into the existing pattern.  I also think that using `ifdef`s to change between arm/x86 code generation will make it harder to maintain.

I'm looking through the compiler code now, and I'm wondering if we should just do the division at the `codegen_object_file` level - just totally new code for x86 and ARM at that point. We'll need a new linking format, different logic for picking instructions, different function prologues/epilogues, etc. With this approach, we'd probably want each IR to support an `do_codegen_x86` and `do_codegen_arm` version, which would give a `x86Instruction` on x86 or `ArmInstruction` on ARM. I haven't thought this through fully yet.

PR #3294, opened 2024-01-07
game: Fix issues related to remapping with SDL axii's (ie. triggers)

xTVaser:
This was just not implemented end to end.  There are still two notable issues, one that I can live with, one I need to narrow down eventually:
1. Rebinding confirm buttons with trigger (ie. X) behaviour is not 100% as it should be.  I fixed it enough that I can live with it but it's still not proper.  The difficulty is that unlike a button it will re-trigger the pressed state on the journey back to neutral (aka unpressed).
2. If you change the bind for the confirm button, then reset your bindings, the next confirm input is eaten.  This is unrelated to these changes but I briefly looked into it and was unable to find the root cause.

### DIFF
--- a/game/system/hid/devices/game_controller.cpp
+++ b/game/system/hid/devices/game_controller.cpp
@@ -91,6 +91,23 @@ void GameController::process_event(const SDL_Event& event,
                binds.button_axii.find(event.caxis.axis) != binds.button_axii.end()) {
       // Binding re-assignment
       if (bind_assignment) {
+        // In the event that the user binds an analog input to the confirm binds
+        // (ie. a trigger on X) we wait until it hits a neutral position (0)
+        // before proceeding to rebind.
+        // 
+        // TODO - this still isn't perfect as the data mutation below will trigger it after the bind
+        // assignment has been set but atleast it's in a mostly working state right now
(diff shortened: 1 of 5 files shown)


xTVaser:
1. That would not fix the problem completely anyway, we use SDL events for the controller and these are not guaranteed to come in before the keyboard events, so that idea where steam simultaneously triggers both events to happen presents a race condition.

Also I don't agree in general:
- The keyboard controls are already so non-obvious that the majority of people are likely already consulting the documentation to know what the bindings are, that documentation can be updated to allow people to enable it.
- I prefer explicit options instead of a bunch of automatic magic that might hamper some use-cases, or cause even more bugs that we need to fix.  

Atleast if someone does ask "how do i use the keyboard" it's an actual simple, consistent answer, instead of the current answer to "my controller's binds are wrong" is "i dunno, close steam, maybe it will work".  It's a huge improvement and we should probably wait to see if we get an influx of people asking about enabling the keyboard before trying to over engineer something.

xTVaser:
Also this isn't even the right PR to comment on, I believe you meant to comment on https://github.com/open-goal/jak-project/pull/3295 ?

ManDude:
wtf? Yeah I meant to comment on that one lol.

PR #3297, opened 2024-01-08
Fix `(-> *autosplit-info-jak1* training-num-orbs)` when blacked out

Zedb0T:
When loading a save, these values are temporarily 0, and it messes with the autosplitter so we want to ignore them then.


Zedb0T:
`(when (not (> (-> *game-info* blackout-time) (current-time)))`

Should I change this to ` (< (-> *game-info* blackout-time) (current-time))`?

In my eyes ` (> (-> *game-info* blackout-time) (current-time))` Is just "in blackout" so I wrote "not in blackout" but idk if it compiles down differently and it would be better to just not have the `(not` call

ManDude:
It doesn't matter, although the actual inverse of `>` is `<=`

xTVaser:
Readability wise, if you are already confused it's a good sign that it would benefit from a properly named intermediary variable
```clj
(let ((in-blackout? (<= (-> *game-info* blackout-time) (current-time))
  (when (not in-blackout?))
    ....
```

Zedb0T:
Oh I didnt see that you requested changes for Jak 2

xTVaser:
Cleaning up stale PRs - this PR has sat for a while with feedback and no movement, re-open if work resumes.

Zedb0T:
Added the fix to Jak 2 and added the local variable per your request as far as I know this is tested and in a good state to merge, If any other changes need to be made please feel free to let me know.

PR #3300, opened 2024-01-09
d/j3: all-types: guess at associated process for non-virtual states and fix some issues along the way

xTVaser:
This will make a best effort attempt at guessing which process a non-virtual state belongs to:
Screenshot 2024-01-08 195309

I also noticed some issues while working on this, mainly around how virtual states were being output.  They were being duplicated, for example:
Screenshot 2024-01-08 184733
or
Screenshot 2024-01-08 193730

I think I've fixed that, but @Hat-Kid I defer to you to see if i've done something terrible.
Screenshot 2024-01-08 194513

### DIFF
--- a/.vs/launch.vs.json
+++ b/.vs/launch.vs.json
@@ -221,6 +221,19 @@
     {
       "type": "default",
       "project": "CMakeLists.txt",
+      "projectTarget": "decompiler.exe (bin\\decompiler.exe)",
+      "name": "Decompiler - Jak 3",
+      "args": [
+        "${workspaceRoot}/decompiler/config/jak3/jak3_config.jsonc",
+        "${workspaceRoot}/iso_data",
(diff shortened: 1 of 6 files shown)


Hat-Kid (commented):
Overall, this looks good to me.

I think the reason for the duplication was that since the addition of the `:state-methods` key, the old code for generating the `:states` list was not updated to accommodate that change and it wasn't really correct before anyway, as that was only grabbing virtual states which were previously supposed to go in the `:methods` list.

One tiny nitpick I have is to maybe omit the parens when the state has no arguments. I think it will still compile with them, but all of our code did it that way before.

xTVaser:
Yeah you're probably right, i was thinking it would make it easier if you had to add arguments, but I would hazard a guess that the majority of states take no arguments.

Hat-Kid (approved):

PR #3312, opened 2024-01-16
fix zydis alignment issue during build for intel macs

JamesAldridge:
Fix zydis alignment issues when building on Intel macOS builds was discussed with zydis devs https://github.com/zyantific/zydis/issues/474

Verified locally that I can run open goal with the change.


water111 (approved):
Thanks for fixing this!

PR #3321, opened 2024-01-20
[jak3] Decompile `gcommon`

water111:
Decompile `gcommon`. I adjusted the spacing of docstring comments, and removed some spammy decompiler warning prints.

I also added some random notes I had on VU programs from jak1/jak2. They are not polished, but I think it's still worth including since we'll have to go through them again for jak 3.


decompiler/config/jak3/all-types.gc
```diff
@@ -5386,7 +5571,7 @@
   :method-count-assert 11
   :size-assert         #xa0
   :flag-assert         #xb000000a0
-  ;; field extra-sound-bank uses ~A with a signed load. 
+  ;; field extra-sound-bank uses ~A with a signed load.
```
water111, line 5574: I stripped a bunch of trailing whitepsace, which is why the diff looks huge. Nothing past here should actually be changed.

PR #3326, opened 2024-01-21
[jak3] Decomp for gkernel, gkernel-h, gstate, gstring

water111:
I ported the kernel test from jak1/jak2 to jak 3, and it's passing!


water111:
This is failing on CI only with a crash:
- it's only the `state-test` one
- The `xmm-check`/`xmm-wreck` isn't required to trigger the crash.
- It seems to die before dispatching the kernel after `state-test` runs.

water111:
dropping the `go` inside the lambda makes it not crash... which makes me think that some prints are being dropped.
```
  (let ((proc (get-process *nk-dead-pool* process 1024 #f)))
    (activate proc *active-pool* "wreck-proc" *kernel-dram-stack*)
    (run-next-time-in-process proc (lambda (x y z w)
                                     (go xmm-wreck-state) ;; <- remove this, and it doesn't crash.
```

PR #3343, opened 2024-01-27
cmake: Remove -Wmissing-include-dir

Ziemas:
I'm tired of it spamming and I couldn't figure out what adds to non-existent fmt/include dir.


water111 (approved):
this is a good idea, thanks.

PR #3348, opened 2024-01-29
Rip collision based on config flag

dallmeyer:
Added a new `rip_collision` flag to the decompiler config and wired it up. Cleaned up the comments around these fields in the JSON too for clarity

Also refactored a bunch of the related extract functions to read from `config` object rather than pass down a bunch of booleans

### DIFF
--- a/decompiler/config/jak3/jak3_config.jsonc
+++ b/decompiler/config/jak3/jak3_config.jsonc
@@ -102,14 +102,16 @@
   // LEVEL EXTRACTION
   ////////////////////////////
 
-  // turn this on to extract level background graphics data
-  "levels_extract": false,
-  // turn this on if you want extracted levels to be saved out as .obj files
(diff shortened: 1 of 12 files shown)


decompiler/config/jak3/jak3_config.jsonc
```diff
@@ -102,14 +102,16 @@
   // LEVEL EXTRACTION
   ////////////////////////////
 
-  // turn this on to extract level background graphics data
-  "levels_extract": false,
-  // turn this on if you want extracted levels to be saved out as .obj files
+  // turn this on to extract level background graphics data as .fr3 files in out/<game>/fr3
+  "levels_extract": true,
```
Hat-Kid, line 106: We might still want to keep this off by default in Jak 3 for now until we have everything needed for proper level extraction. Right now, the decompiler is probably going to fail when run with this flag.

Hat-Kid (approved):

PR #3357, opened 2024-02-02
game: fix intermittent sprite_distort related crash

xTVaser:
Apparently a long standing bug, with https://github.com/open-goal/jak-project/pull/3194 making it more likely to occur.


water111 (approved):
:partying_face:

PR #3366, opened 2024-02-04
decomp: Copy over new docstrings back to jak 2's common code, write some scripts to help automate this

xTVaser:


water111 (approved):
thanks!

PR #3369, opened 2024-02-10
Decompile `joint`, `collide-func`, clean up joint decompression code for all games

water111:
I finally read through all the joint code and wrote up some documentation. I think this will be really helpful when we try to understand all the functions in `process-drawable`, or if somebody ever wants to import/export animations.

This switches all three games to using a new faster GOAL joint decompressor. It is on by default, but you can go back to the old version by setting `*use-new-decompressor*` to false.

Also fix the log-related crash, fix the clock speed used in timer math.


decompiler/config/jak3/jak3_config.jsonc
```diff
@@ -7,7 +7,7 @@
 
   // if you want to filter to only some object names.
   // it will make the decompiler much faster.
-  "allowed_objects": [],
+  "allowed_objects": ["joint"],
```
ManDude, line 10: accidental change

PR #3377, opened 2024-02-17
decomp3: more engine stuff, detect non-virtual state inheritance

Hat-Kid:
- `speech`
- `ambient`
- `water-h`
- `vol-h`
- `generic-obs`
- `carry-h`
- `pilot-h`
- `board-h`
- `gun-h`
- `flut-h`
- `indax-h`
- `lightjak-h`
- `darkjak-h`
- `target-util`
- `history`
- `collide-reaction-target`
- `logic-target`
- `sidekick`
- `projectile`
- `voicebox`
- `ragdoll-edit`
- most of `ragdoll` (not added to gsrc yet)
- `curves`
- `find-nearest`
- `lightjak-wings`
- `target-handler`
- `target-anim`
- `target`
- `target2`
- `target-swim`
- `target-lightjak`
- `target-invisible`
- `target-death`
- `target-gun`
- `gun-util`
- `board-util`
- `target-board`
- `board-states`
- `mech-h`
- `vol`
- `vent`
- `viewer`
- `gem-pool`
- `collectables`
- `crates`
- `secrets-menu`

Additionally:

- Detection of non-virtual state inheritance
- Added a config file that allows overriding the process stack size set by `stack-size-set!` calls
- Fix for integer multiplication with `r0`
- Fixed detection for the following macros:
	- `static-attack-info`
	- `defpart` and `defpartgroup` (probably still needs adjustments, uses Jak 2 implementation at the moment)
	- `sound-play` (Jak 3 seems to always call `sound-play-by-name` with a `sound-group` of 0, so the macro has been temporarily defaulted to use that)

One somewhat significant change made here that should be noted is that the return type of `process::init-from-entity!` was changed to `object`. I've been thinking about this for a while, since it looks a bit nicer without the `(none)` at the end and I have recently encountered init methods that early return `0`.


decompiler/config/jak3/all-types.gc
```diff
@@ -16835,63 +16836,67 @@
...
+  (race-daxter-win 51)
+  (race-errol-start 52)
+  (race-errol-last-lap 53)
+  (race-errrol-hit 54)
+  (race-errol-got-hit 55)
+  (race-errol-ambient 56)
+  (race-errol-pass 57)
+  (race-errol-got-passed 58)
+  (race-errol-win 59)
+  (race-errol-lose 60)
```
ManDude, line 16899: these could be added to Jak 2 later!

goal_src/jak3/engine/common-obs/gem-pool.gc
```diff
@@ -7,3 +7,89 @@
...
+      0.0
+      )
+    )
+  )
+
+;; WARN: Return type mismatch symbol vs none.
+(defun compare-gems ((arg0 object) (arg1 (pointer float)))
+  (< (-> arg1 0) (get-gem-pool-size (the-as int arg0)))
+  (none)
+  )
```
ManDude, line 41: careful with returning `none`

goal_src/jak3/engine/common-obs/generic-obs.gc
```diff
@@ -5,5 +5,4061 @@
...
+  (set! (-> self options) arg4)
+  (when (-> self draw shadow)
+    (set! (-> self draw shadow-ctrl) (new
+                                       'process
+                                       'shadow-control
+                                       0.0
+                                       0.0
+                                       614400.0
+                                       (the-as vector #f)
+                                       0.000000000000000000000000000000000000000000084
```
water111, line 1035: I think this isn't a float?

water111 (approved):
:+1:

PR #3378, opened 2024-02-19
g/j2: Dynamic speedrun mode categories and implement a significant amount of a practice mode

xTVaser:
For example, `AppData/OpenGOAL/jak2/features/speedrun-categories.json` is defined as such:
```json
[
  {
    "cheats": 0,
    "completed_task": 0,
    "continue_point_name": "",
    "features": 0,
    "forbidden_features": 992,
    "name": "Gunless",
    "secrets": 0
  },
  {
    "cheats": 1,
    "completed_task": 29,
    "continue_point_name": "ctypal-shaft",
    "features": 1024,
    "forbidden_features": 0,
    "name": "Turbo Jetboard - After Praxis 1",
    "secrets": 0
  }
]
```
> These entries can be created using the in-game menu as well.

https://github.com/open-goal/jak-project/assets/13153231/9b17a116-4aa9-40ad-b9f5-02b04e0ad4f3


ManDude:
idea: cheat-mask that determines which cheats are allowed to be toggled? e.g. so you can have jet-board trick display or no-texture mode on or some other visual cheat if you want (this would also let us get rid of the hacks to set/unset the cheats when speedrun mode is toggled)

xTVaser:
>  (this would also let us get rid of the hacks to set/unset the cheats when speedrun mode is toggled)
It would not, there is value in autosetting the cheats as explained here https://github.com/xTVaser/jak-project/blob/555d136395cf252bce5fbb4a62dafb035e955209/goal_src/jak2/pc/features/speedruns.gc#L104-L110

I don't think there is a lot of value in your idea, in what speedrun do you randomly toggle on a cheat like jetboard trick display, or no-texture mode.  It sounds like unnecessary bloat, the current features are aimed to provide what is needed to do speedrunning.

ManDude:
> in what speedrun do you randomly toggle on a cheat like jetboard trick display, or no-texture mode

You wouldn't toggle it during the speedrun (though that would be allowed), you would do it before. That isn't very different in practice to someone running the game with texture mods if they so desired.

> It would not

Yes, it would, but also I don't think I made very clear. You can use the `pc-cheats?` macro (which is what's already used in the cheats code and how cheat checking is done in general) to automatically filter out cheats not allowed during the speedrun, instead of manually setting a value and a backup which has a possibility of not working for some reason or running into some stale state, e.g. after changing some menu code or the user (or developer) changes settings in some manner unaccounted for/something we forgot/didn't bother with. A mask that makes sure the filter out/in specific cheats seems a lot more robust and is also less code.

In most practical scenarios it wouldn't be different to the current behavior, it simply allows a bit more freedom in letting people run with whatever settings they want that don't affect the run. This PR is already doing it with the `forbidden-features` field.

xTVaser:
> edit: also the current method of "set the cheats to 0 every time" has the potential issue of the cheat still working for 1 frame unless you also add guards to every spot where you can set cheats. That seems like a lot more bloat than a simply mask tucked away in a macro.

In practice this 1 frame difference is irrelevant...it takes multiple frames to exit the menu.  Also this is irrelvant because the tools are not trying to 100% perfectly enforce the cheats not being set.  For the last time -- it is to be convenient for the majority of situations where people just want the right settings set for them instead of having to do it manually.

> You wouldn't toggle it during the speedrun (though that would be allowed), you would do it before

This is literally already accomplished with this PR.  If you want to turn on a cheat -- make a category with the cheat on.  I don't understand what your point is.

Issue #2988, opened 2023-09-14
[jak2] overlord crash in `ruins-mech` (speedrun route)


Hat-Kid:
<https://youtu.be/x5UP--2qCbI?t=7147>

![ruins-mech](https://github.com/open-goal/jak-project/assets/6624576/c06ddd5d-0652-4375-adad-25ef63aac4c8)

xTVaser:
This happens when doing it the speedrunner way (the hut is not loaded in fully).  I've narrowed down a few things, but still not the true root-cause for why things get into this state.

So first off, the crash is here in the `memcpy` in `sceSdVoiceTrans`:
![Screenshot 2024-01-11 012933](https://github.com/open-goal/jak-project/assets/13153231/361554c1-7468-4ae7-a090-186810fd026a)
> the channel and the iop_addr seem all wrong here

Stepping back up the chain, this only happens when the dma code receives a `VagCmd` with the name `ruins-pillar-collapse-1`.  This command is inconsistently received (it doesn't happen every time) -- and the game _only_ crashes when it gets this command.
![image](https://github.com/open-goal/jak-project/assets/13153231/6c5f6442-326c-4780-ac55-9135cbfc2e25)

My naive assumption is that in the original game this exact same thing happens, but the game timeouts gracefully.  Notice how in a normal speedrun, loading this cutscene takes a considerable amount of time: https://youtu.be/HYTg_dNDdv8?t=516

I was able to prevent the crash (but i wouldn't be surprised if this causes other issues) by returning early from the DMA code if a `channel` below `0` is received.  This kinda breaks the cutscene, but atleast it does not crash:

https://github.com/open-goal/jak-project/assets/13153231/65ff0982-9e0b-4f2d-a74a-1aa9f703cd1b
> the cutscene starts paused and skips on it's own

Ziemas:
Doesn't seem like the real implementation of sceSdVoiceTrans does any sanity checking on arguments other than the size. Given these parameters I'm pretty sure it would actually perform the DMA from 0x0 using SPU core 1 (channel: -1 & 1).

Idk if that's actually what happens or real hw though.

ManDude:
I always assumed the slow loading times was due to `sagehut` needing to be loaded, as you skip all its loading triggers. This could perhaps be related, but would be weird because the same situation happens... whenever you select anything from the scene player.

ManDude:
[...]

PR #3418, opened 2024-03-09
g/j2: workaround intermittent lifeseed DMA crash

xTVaser:
This is not a fix, but it temporarily works around the potential crash discussed here https://github.com/open-goal/jak-project/issues/2988

As long as this doesn't impact other things, this feels like a good interim fix, as having a broken cutscene is better than a fatal crash and it might be some time before this is properly identified and fixed.

### DIFF
--- a/game/overlord/jak2/dma.cpp
+++ b/game/overlord/jak2/dma.cpp
@@ -1,5 +1,8 @@
 #include "dma.h"
 
+#include "common/log/log.h"
+#include "common/util/string_util.h"
+
 #include "game/overlord/jak2/ssound.h"
 #include "game/overlord/jak2/vag.h"
 #include "game/sound/sdshim.h"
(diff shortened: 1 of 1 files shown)


ManDude (approved):

xTVaser:
macOS tests seems to often intermittently crash over exit code `132` now after updating from 11 to 12?

Seemingly no logs to shed light on the problem, but i did see:
> Warning:  /Users/runner/work/jak-project/jak-project/third-party/googletest/googletest/src/gtest-death-test.cc:1124:: Death tests use fork(), which is unsafe particularly in a threaded context. For this test, Google Test detected 5 threads. See https://github.com/google/googletest/blob/master/docs/advanced.md#death-tests-and-threads for more explanation and suggested solutions, especially if this is the last message you see before your test times out.

PR #3432, opened 2024-03-21
decomp3: more engine files, `get-texture` macro, use `print` method in autogenerated inspect, fix bitfield float print

Hat-Kid:
- `fma-sphere`
- `prim-beam-h`
- `cam-start`
- `ragdoll`
- `light-trails-h`
- `light-trails`
- `menu`
- `water`
- `water-flow`
- `hud`
- `hud-classes`
- `progress`
- `progress-draw`

---

The `get-texture` macro replaces calls to `lookup-texture-by-id` and `lookup-texture-by-id-fast`. The `defpart` macro detection was modified to print a pair like `(texture-name tpage-name)` for the texture field that gets turned into a `texture-id` constant. Only used in Jak 3 at the moment, I'll probably go through the other games at a later point.


goalc/compiler/compilation/Type.cpp
```diff
@@ -170,7 +170,24 @@ void Compiler::generate_field_description(const goos::Object& form,
     format_args.push_back(get_field_of_structure(type, reg, f.name(), env)->to_gpr(form, env));
   } else if (m_ts.tc(m_ts.make_typespec("structure"), f.type())) {
     // Structure
-    str_template += fmt::format("{}{}: #<{} @ #x~X>~%", tabs, f.name(), f.type().print());
+    auto ts = m_ts.lookup_type_no_throw(f.type());
+    if (ts) {
+      // try to use print method if the structure implements it
```
water111, line 175: This probably needs more testing, but I think you could always use the print method case (`{}{}: ~`{}`P~%`). That should work on everything. If you don't define a custom print method, I think it will fall back on a default one.

water111 (approved):

goalc/compiler/compilation/Type.cpp
Hat-Kid: This is what I tried at first, but this makes `structure`s without a `print` method default to `structure`'s `print`, which would always print `#<structure @ #x~X>` instead of `#<actual-type-name @ #x~X>`.

water111: Oh, good point.

I guess you'd need to check to see if the `print` method of the type is different from the `print` method of `structure`, and only use the actual print method if it is... maybe not worth the effort. Just seeing the `vector`s in inspect will be a huge improvement.

PR #3434, opened 2024-03-24
decomp drawable, main

water111:


decompiler/config/jak3/all-types.gc
```diff
@@ -100,7 +100,7 @@
 (define-extern file-stream-write (function file-stream pointer uint uint))
 (define-extern file-stream-close (function file-stream file-stream))
 (define-extern new-dynamic-structure (function symbol type int structure))
-(define-extern kernel-shutdown (function none))
+(define-extern kernel-shutdown (function int none))
```
Hat-Kid, line 103: I added a `runtime-exit-status` enum in Jak 2's `kernel-defs.gc` (matches the one in the kernel) that could be used here.

water111: unfortunately, ND has added an `int` here and made up their own meaning... but it's for the movie stuff so we can just ignore it. I'll add the enum.

PR #3449, opened 2024-04-05
[goal] Support loop unrolling with :unroll factor

Brent-Hickey:
Not sure if this will actually be useful but thought it'd be cool to support

Supports optional [loop unrolling ](https://en.wikipedia.org/wiki/Loop_unrolling#:~:text=Loop%20unrolling%2C%20also%20known%20as,known%20as%20space%E2%80%93time%20tradeoff.) with :unroll factor
Supports (_ x) syntax to unroll without binding and incrementing a variable when fully unrolled. Otherwise uses gensym var to iterate.

Could add to countdown as well

Examples:
``` lisp
 (dotimes (_ 4) :unroll 4
    (print "hi")
  )

(begin (print "hi") (print "hi") (print "hi") (print "hi"))
```
(unchanged)
``` lisp
(dotimes (i 4)
    (format #t "i: ~D~%" i)
  )

((lambda :immediate #t (i) (while (< i 4)
                             :label #f
                             (block continue (format #t "i: ~D~%" i) (1+! i))
                             )
         )
 0
 )
```

``` lisp
(dotimes (_ 4) :unroll 2
    (print "hi")
  )

((lambda :immediate
   #t
   (gensym20077)
   (while (< gensym20077 4)
     :label #f
     (block continue (print "hi") (1+! gensym20077) (print "hi") (1+! gensym20077))
     )
   )
 0
 )
```

``` lisp
(dotimes (i 4) :unroll 4
    (format #t "i: ~D~%" i)
  )

((lambda :immediate
   #t
   (i)
   (format #t "i: ~D~%" i)
   (1+! i)
   (format #t "i: ~D~%" i)
   (1+! i)
   (format #t "i: ~D~%" i)
   (1+! i)
   (format #t "i: ~D~%" i)
   (1+! i)
   )
 0
 )
```

### DIFF
--- a/goal_src/goal-lib.gc
+++ b/goal_src/goal-lib.gc
@@ -372,19 +372,89 @@
   )
 )
 
-(defmacro dotimes (var &key (label #f) &rest body)
-  "Loop like for (int i = 0; i < end; i++)
-   var is a list made up of a variable to bind the amount to (second item), and the remaining forms are evaluated after the loop is finished."
-  (let ((continue-label (string->symbol (if label (string-append (symbol->string label) "-continue") "continue"))))
-    `(let (( ,(first var) 0))
-      (while (< ,(first var) ,(second var)) :label ,label
-              (block ,continue-label
-                ,@(if (null? body) (list `(return-from ,continue-label #f)) body)
-              )
-              (1+! ,(first var))
-              )
-      ,@(cddr var)
+(defmacro dotimes (var &key (label #f) &key (unroll 1) &rest body)
+  "Loop like for (int i = 0; i < end, i++)
+   var is a list made up of a variable to bind the amount to (second item), and the remaining forms are evaluated after the loop is finished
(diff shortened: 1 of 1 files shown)


goal_src/goal-lib.gc
```diff
@@ -376,19 +376,89 @@
...
+       )
+
+    (when (and (integer? limit)
+               (> unroll limit)
+           )
+      (error ":unroll factor must be <= the loop limit.")
+    )
+    (when (< unroll 1)
+      (error ":unroll factor must be >= 1.")
+    )
```
xTVaser, line 398: For whatever reason, in the original code they left `dotimes` in that iterate `0` times.  This highlights an edge-case with the assumption that every loop will execute atleast once.

See `cloth.gc` in jak3

xTVaser (changes requested):

PR #3452, opened 2024-04-06
Add fall back condition

Zedb0T:


Zedb0T:
https://github.com/open-goal/jak-project/blob/69c628dd74ca96fa10cd8e3dce5c7d08021ee5dc/goal_src/jak1/pc/progress-pc.gc#L867-L880

Its probably better to just move `(set! (-> *options-remap* (progress-screen graphic-settings)) *graphic-options-pc*)` to line 867 before the let right? Since we don't need to display different menus conditionally anymore?

xTVaser (approved):

PR #3484, opened 2024-04-25
Fix "logical not is only applied to the left hand side of comparison" warnings

dallmeyer:


ManDude:
...wow that's a dumb warning

PR #3489, opened 2024-04-28
decomp3: some textures files

Nick0778:
I wanted to contribute to the decompilation process, so I decompiled some texture files present in some levels. I'm not sure if this will be helpful at the moment, but I just wanted to help. It is worth mentioning that, for some strange reason, the game crashes when trying to access these levels. I believe that the problem is not the decompiled files, considering that there are no errors in them. Maybe animated textures still don't work correctly at the moment?
Anyway, here are the files I decompiled:

- `templea-texture`
- `templex-texture`
- `volcanox-texture`
- `volcano-texture`
- `sewer-texture`


Hat-Kid:
We haven't decompiled these files yet because texture animations aren't implemented yet in Jak 3. The `texture-anim` code also hasn't been decompiled yet, which is why the game is crashing.

xTVaser:
ok to test

PR #3494, opened 2024-04-30
jak3 - fill out some `text-id` entries

dallmeyer:
Slam Dozer > Ram Rod


xTVaser:
If conflicts were fixed this could probably be merged.

Issue #1353, opened 2022-05-03
[goalc] New system for tracking GOAL dependencies


water111:
### Current issues that need solving:
- the dependencies for GOAL files in `game.gp` are "correct" in that they result in correct builds, but they add a lot of false dependencies that make incremental builds slow
- there is no easy way for a `.gc` file to explicitly state its dependencies or include another file.
- There is confusion between the build order and the order of files in the DGOs
- Incremental builds generate tons of object files they don't need
- the compiler memory usage increases with each build

### New idea

Input: the user will provide a list of files that should be compiled. This can be one file, or a list of all the files in the DGO.  The compiler will generate up-to-date object files for all the specified files. 
New build procedure:

1. Parse `import` statements
2. Build dependency graph
3. Compile

### Step 1: parsing `import` statements
For each file in the input, the compiler will read and parse the file. It will store the `goos::Object` for this file and will look for `(import <filename>)` statements in the top level.  For example, a file might have:
`(import "kernel/gcommon.gc")` if it relies on macros in `gcommon`.  This process will follow `import` statements and build up a list of edges in a dependency graph.

I think we'd quickly end up with massive lists of dependencies, so it will be possible to create "libraries". For example, a library like `math` can be created by writing a file like:
```
;; libmath.gc
(import "game/math/vector.gc")
(import "game/math/trigonometry.gc")
...
```
and then any file can simply do `(import "game/math/libmath.gc")` to depend on all the math files.

A file that requires an art-group constants file might do: `(import "game_data/art-groups/hopper-ag-constants.gc")` or something like that.

Note that it will not possible to use `import` within a macro.

Parsing files into a `goos::Object` and list of imports will be cached, so if a file is unchanged, the compiler can use the result from the last compilation.

### Step 2: build dependency graph
This step will come up with a build order. It will also make sure there are no circular dependencies.

Files that are explicitly requested in the input will always be compiled to `.o` object files.  Other files will not be fully compiled - just the front end will run on them.  This means that library files or art-group constant files won't have associated `.o` files.

### Step 3: build!

The compiler will build each file in order. It will also perform dependency checking.
If you are missing an `import` statement, it's possible to get lucky and the compiler happened to compile the file before. This case should still be an error.  The error should look like this:
```
(vector+! a b c)
 ^
Error: using definition of function vector+!, defined in math/vector.gc, but this file was not imported
```
[...]

PR #3505, opened 2024-05-06
Dependency graph work - Part 1 - Preliminary work

xTVaser:
Relates to #1353 

This adds no new functionality or overhead to the compiler, yet.  This is the preliminary work that has:
- added code to the compiler in several spots to flag when something is used without being properly required/imported/whatever (disabled by default)
- that was used to generate project wide file dependencies (some circulars were manually fixed)
- then that graph underwent a transitive reduction and the result was written to all `jak1` source files.

The next step will be making this actually produce and use a dependency graph.  Some of the reasons why I'm working on this:
- eliminates more `game.gp` boilerplate.  This includes the `.gd` files to some extent (`*-ag` files and `tpage` files will still need to be handled) this is the point of the new `bundles` form. This should make it even easier to add a new file into the source tree.
- a build order that is actually informed from something real and compiler warnings that tell you when you are using something that won't be available at build time.
- narrows the search space for doing LSP actions -- like searching for references.  Since it would be way too much work to store in the compiler every location where every symbol/function/etc is used, I have to do ad-hoc searches.  By having a dependency graph i can significantly reduce that search space.
- opens the doors for common shared code with a legitimate pattern.  Right now jak 2 shares code from the jak 1 folder.  This is basically a hack -- but by having an explicit require syntax, it would be possible to reference arbitrary file paths, such as a `common` folder.

Some stats:
- Jak 1 has about 2500 edges between files, including transitives
- With transitives reduced at the source code level, each file seems to have a modest amount of explicit requirements.

Known issues:
- Tracking the location for where `defmacro`s and virtual state definitions were defined (and therefore the file) is still problematic.  Because those forms are in a macro environment, the reader does not track them.  I'm wondering if a workaround could be to search the reader's text_db by not just the `goos::Object` but by the text position.  But for the purposes of finishing this work, I just statically analyzed and searched the code with throwaway python code.


goalc/compiler/compilation/ConstantPropagation.cpp
```diff
@@ -245,19 +247,43 @@ Compiler::ConstPropResult Compiler::constant_propagation_dispatch(const goos::Ob
...
+            env->file_env()->m_missing_required_files.insert(
+                enum_type->m_metadata.definition_info->filename);
+          }
+        }
         bool success;
-        u64 as_enum =
-            enum_lookup(prop.value, enum_type, prop.value.as_pair()->cdr, false, &success);
+        // TODO - changed false to true so you'd actually once again get a useful error (something
+        // wasn't found in the enum instead of just "cant be used as a constant") Ramifications of
+        // this?
```
water111, line 276: I think you're right, and this was probably just a copy-paste mistake. If this enum lookup fails, you are doomed.

goalc/compiler/compilation/Function.cpp
```diff
@@ -315,10 +317,26 @@ Val* Compiler::compile_lambda(const goos::Object& form, const goos::Object& rest
...
+      const auto& result = symbol_info.at(0);
+      if (result->m_def_location.has_value() &&
+          !env->file_env()->m_missing_required_files.contains(result->m_def_location->file_path) &&
+          env->file_env()->m_required_files.find(result->m_def_location->file_path) ==
+              env->file_env()->m_required_files.end() &&
+          !str_util::ends_with(result->m_def_location->file_path,
+                               env->file_env()->name() + ".gc")) {
+        lg::warn("Missing require in {} for {} over {}", env->file_env()->name(),
+                 result->m_def_location->file_path, uneval_head.print());
+        env->file_env()->m_missing_required_files.insert(result->m_def_location->file_path);
```
water111, line 336: yeah I think this would be nicer if it was in a method in `FileEnv` instead.

goalc/compiler/compilation/Type.cpp
```diff
@@ -89,6 +90,33 @@ RegVal* Compiler::compile_get_method_of_object(const goos::Object& form,
     }
   }
 
+  if (m_settings.check_for_requires) {
+    std::vector<symbol_info::SymbolInfo*> symbol_info =
+        m_symbol_info.lookup_exact_method_name(method_info.name, compile_time_type.base_type());
+    // TODO - check for virtual state requirement, if i could track the defining form!
+    // if (symbol_info.empty()) {
+    //  // maybe it's a virtual state
```
water111, line 98: yeah - I think this would be easier if we tracked this in the method info's. But don't worry about it for now

goalc/compiler/compilation/Type.cpp
```diff
@@ -1464,7 +1492,12 @@ Val* Compiler::compile_defenum(const goos::Object& form, const goos::Object& res
   (void)form;
   (void)env;
 
-  parse_defenum(rest, &m_ts, {});
+  DefinitionMetadata def_metadata;
+  const auto new_enum = parse_defenum(rest, &m_ts, &def_metadata);
+  new_enum->m_metadata.definition_info = m_goos.reader.db.get_short_info_for(form);
+  m_symbol_info.add_enum(new_enum->get_name(), new_enum->get_parent(), new_enum->entries(),
+                         new_enum->is_bitfield(), form,
+                         def_metadata.docstring.has_value() ? def_metadata.docstring.value() : "");
```
water111, line 1500: The data in here seems somewhat redundant with the existing `EnumType`. Could we just store a `EnumType*` in the symbol info?

water111 (commented):
Thanks, this looks like a great start toward solving the current disaster of dependency tracking.

PR #3535, opened 2024-05-25
g/j1: Cleanup all main issues in the formatter and format all of `goal_src/jak1`

xTVaser:
This PR does two main things:
1. Work through the main low-hanging fruit issues in the formatter keeping it from feeling mature and usable
2. Iterate and prove that point by formatting all of the Jak 1 code base.  **This has removed around 100K lines in total.**
    - The decompiler will now format it's results for jak 1 to keep things from drifting back to where they were.  This is controlled by a new config flag `format_code`.

How am I confident this hasn't broken anything?:
- I compiled the entire project and stored it's `out/jak1/obj` files separately
- I then recompiled the project after formatting and wrote a script that md5's each file and compares it (`compare-compilation-outputs.py`
- The results (eventually) were the same:

Screenshot 2024-05-25 132900
> This proves that the only difference before and after is non-critical whitespace for all code/macros that is actually in use.

I'm still aware of improvements that could be made to the formatter, as well as general optimization of it's performance.  But in general these are for rare or non-critical situations in my opinion and I'll work through them before doing Jak 2.  The vast majority looks great and is working properly at this point.  Those known issues are the following if you are curious:
image


xTVaser:
Merged master and rechecked the compilation outputs against vanilla master.  They still match (i purposed changed the gitignore file as a litmus test, it was the only file that was found to diff.)
image This also means that any changes to `preserve-this` comments is superficial -- the compiled output is the same!

Was waiting a few days after last month's release to make sure it was stable, appears to be.  Will merge once CI passes

PR #3553, opened 2024-06-07
[jak1] Enforce certain pc-settings in speedrunner mode

dallmeyer:
Currently PS2 Actor Vis and FPS are only enforced when starting a run - this enforces them on every frame similar to cheats.
In the progress menu, FPS is already disabled in speedrunner mode - this adds the same restriction for PS2 Actor Vis.

Jak 2 already does this properly, no change needed there

### DIFF
--- a/goal_src/jak1/pc/features/speedruns.gc
+++ b/goal_src/jak1/pc/features/speedruns.gc
@@ -228,7 +228,6 @@
      (initialize! *game-info* 'game (the-as game-save #f) "intro-start"))
     (else
      (format 0 "start-speedrun: unrecognized category ~S~%" (enum->string speedrun-category (-> *speedrun-info* category)))))
-  
   (enforce-speedrun-pc-settings)
   ;; enable auto saving by default
   (set! (-> *setting-control* default auto-save) #t)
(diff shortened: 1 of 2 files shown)


goal_src/jak1/pc/features/speedruns.gc
```diff
@@ -133,6 +133,18 @@
...
 
+(defun enforce-speedrun-pc-settings ()
+  ;; Disable any active cheats
+  (set! (-> *pc-settings* cheats) (the-as pc-cheats #x0))
+  ;; ensure PS2 actor vis is enabled
+  (set! (-> *pc-settings* ps2-actor-vis?) #t)
+  ;; ensure FPS set to `60`
+  (when (!= (-> *pc-settings* target-fps) 60)
+    (set-frame-rate! *pc-settings* 60 #t)
+    )
```
dallmeyer, line 144: I got a bunch of console spam about "Tried to set unsupported video-mode" if I actually called this every frame, since `set-game-setting` resets `current` and assigns the specified value to `default`, but `(get-video-mode)` reads from `current`:
```lisp
(defmethod set-game-setting! ((obj pc-settings-jak1) (setting symbol) (value symbol))
  (case setting
    (('video-mode)
     (set! (-> *setting-control* current video-mode) #f)
     (set! (-> *setting-control* default video-mode) value))
    (('aspect-ratio) (set! (-> *setting-control* default aspect-ratio) value))
    (else (format #t "unknown setting  ~A (~A) to set-game-setting!" setting value))))

PR #3554, opened 2024-06-07
ci: fix broken windows builds

xTVaser:
A recent windows image rolled out which has multiple problems:
- https://github.com/actions/runner-images/issues/10001
  - VS was updated, this new toolchain requires atleast Clang17
  - This is fine for us, but by default the image chooses to use the system installed clang (which is still clang16)
- https://github.com/actions/runner-images/issues/10004
  - If you work around that, your binaries will crash upon launching.  This is allegedly because for whatever reason, an old `msvcp140.dll` is being used.  There is a potential workaround for this but hopefully we can just wait for a fixed image.


xTVaser:
Allegedly resolved (re-running other builds)

PR #3560, opened 2024-06-20
decompiler: Cleanup duplication in extractor/decompiler and make it easier to enable streamed audio ripping from CLI

xTVaser:
This centralizes the code that both `extractor` and the decompiler executes.  In the past this code was partially-duplicated, meaning that the `extractor` could only do _some_ operations and not others (ie. could not extract the audio files).

I also simplified the process to enable audio streaming in the configuration.  This is to support a new feature in the launcher that allows you to enable these options for the decompiler:

image

### DIFF
--- a/common/util/os.cpp
+++ b/common/util/os.cpp
@@ -3,17 +3,29 @@
 #include "common/common_types.h"
 #include "common/log/log.h"
 
-#ifdef __linux__
-
+#ifdef _WIN32
+// clang-format off
+#define NOMINMAX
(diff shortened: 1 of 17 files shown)


water111 (approved):
Thanks. I didn't look too closely, but this seems like a good idea to clean up.

xTVaser:
Tested via the launcher, can still install both games normally and now can also decompile the various assets using the launcher and passing in config overrides.

PR #3578, opened 2024-07-14
[high fps] Increase input buffer for jak1 and jak3

Brent-Hickey:
Applying https://github.com/open-goal/jak-project/pull/3178 to jak1 and jak3

This also fixes cloud speed in jak3


Hat-Kid (approved):
I fixed the whitespaces and added a couple more `preserve-this` comments, looks good otherwise.

PR #3595, opened 2024-07-20
Jak1: Add Autosplitting options for tasks completed in one level and turned in in another

zakfaulk:
For the FJ Mirrors, Muse, Lightning Moles, and Gambler's race the tasks are completed in one level, then turned in at the hub. This presents a spot of confusion for new players with the autosplitter and requires hacky workarounds for those who want to split on these conditions. This pull request adds in the necessary code to the autosplit-h.gc and autosplit.gc files so that the autosplitter can identify these events, similar to the talk to fisherman and catch fish options in the current autosplitter.


xTVaser:
Remove the trailing whitespace in `autosplit.gc` and this can merge:
image

xTVaser:
ok to test

Issue #3563, opened 2024-06-23
Black screen right on launch 


ItsModern:
### Acknowledgements

- [X] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak2+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [X] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

Just as the title described, I have the game, I click launch and its just a black screen. I can hear the game and can hear im moving the menu but thats it. I tried debug mode, also doesnt work. Tried going in window mode, doesnt work. It does work on PCSX2 but not open goal. 

### How To Reproduce

Launch the game, boom black screen. 

### Does this problem occur on original hardware or PCSX2?

Yes, it's unique to OpenGOAL

### Expected Behavior

When I load the game... I should see the game. 

### Environment Information

[opengoal-support-package.zip](https://github.com/user-attachments/files/15942550/opengoal-support-package.zip)
Windows 10, Nvidia 3070, Core i7 processor 10700

### Game Version

NTSC 1.0 (black label)

### Have you set the game to something other than `60fps`?

No

xTVaser:
Is your problem related to this https://github.com/open-goal/jak-project/issues/3279#issuecomment-1975011557

This has not been able to be traced back to anything in our code as of yet, and we havn't had reports of it in months.

ItsModern:
> Is your problem related to this [#3279 (comment)](https://github.com/open-goal/jak-project/issues/3279#issuecomment-1975011557)
> 
> This has not been able to be traced back to anything in our code as of yet, and we havn't had reports of it in months.
> 
> > [31:29] [error] [end] OpenGL error 0x502 S8246 T824C: GL_INVALID_OPERATION error generated. Source and destination dimensions must be identical with the current filtering modes.
> 
> from logs appears relevant.

Just checked, neither of those are in my services sadly. So i dont know what else can be causing it, Jak 1 runs perfectly fine

xTVaser:
Odd, as that error is both in your jak1 and jak2 logs.

I'd try the following, open the `pc-settings.gc` file in notepad.  You can find it by clicking on the following in the jak 2 menu:
![image](https://github.com/open-goal/jak-project/assets/13153231/a713f313-cac6-49b4-af01-9d9e8de4f953)

- Try replacing `(display-mode borderless)` with `(display-mode fullscreen)`
[...]

PR #3601, opened 2024-07-24
game: cleanup some display settings related code, forbid invalid `game-size` resolutions

xTVaser:
Fixes https://github.com/open-goal/jak-project/issues/3563

These users have the following spamming in logs:
> OpenGL error 0x502 S8246 T824C: GL_INVALID_OPERATION error generated. Source and destination dimensions must be identical with the current filtering modes.

And the solution is to correctly set their game-size.  The way this change accomplishes that is by confirming whether or not the set `game-size` is a valid resolution informed by SDL, if not, it defaults to the monitor's currently set display mode's resolution.

This also moves the selected display id, and the display mode into the C++ settings -- closer to where it's actually managed and used.  I'm tempted to do this eventually for the resolutions as well but that stuff is much more burdensome.  This hopefully simplifies debugging, reduces startup flickering, and removes back-and-forth complexity.  Hopefully this makes debugging display related problems easier.  It also adds a bunch more logging to the related code.


xTVaser:
Tested this in Jak1 and Jak 2:
- [x] Starting from settings from master (ensure backwards compatibility -- the game still launches)
- [x] Set and restore all display settings
- [x] Set monitors, restore properly for all display settings
- [x] Have an invalid game-size for all display settings

PR #3619, opened 2024-07-31
[jak1] update finnish translations

Aloqas:
A long overdue Finnish translation update with better subtitle timings and wording.


xTVaser:
ok to test

PR #3625, opened 2024-08-02
decompiler: `call-parent-state-handler` and `suspend-for` macros

Hat-Kid:
Also fix `hud-draw-pris2` bucket in Jak 3 to make subtitles work and foreground HUD envmap.


ManDude:
Should probably add `suspend-for` to the list of forms that force a line break for the body portion in the PrettyPrinter files.

PR #3632, opened 2024-08-10
Extract merc models into individual glb files

dallmeyer:
e.g. instead of just `beach-foreground.glb` and `beach-background.glb`, you'd now get:
- `beach/babak-lod0.glb`
- `beach/babak-lod1.glb`
- `beach/babak-lod2.glb`
- `beach/barrel-lod0.glb`
- `beach/beach-background.glb`
- `beach/beachcam-lod0.glb`
- ... (42 other files)
- `beach/windmill-one-lod2.glb`

`common` models are also grouped into their own folder

### DIFF
--- a/decompiler/level_extractor/extract_level.cpp
+++ b/decompiler/level_extractor/extract_level.cpp
@@ -323,8 +323,7 @@ void extract_common(const ObjectFileDB& db,
 
   if (config.rip_levels) {
     auto file_path = file_util::get_jak_project_dir() / "glb_out" /
-                     game_version_names[config.game_version] / "common.glb";
-    file_util::create_dir_if_needed_for_file(file_path);
+                     game_version_names[config.game_version] / "common";
     save_level_foreground_as_gltf(tfrag_level, art_group_data, file_path);
   }
 }
@@ -361,14 +360,12 @@ void extract_from_level(const ObjectFileDB& db,
 
   if (config.rip_levels) {
     auto back_file_path = file_util::get_jak_project_dir() / "glb_out" /
-                          game_version_names[config.game_version] /
+                          game_version_names[config.game_version] / level_data.level_name /
                           fmt::format("{}-background.glb", level_data.level_name);
     file_util::create_dir_if_needed_for_file(back_file_path);
     save_level_background_as_gltf(level_data, back_file_path);
     auto fore_file_path = file_util::get_jak_project_dir() / "glb_out" /
-                          game_version_names[config.game_version] /
-                          fmt::format("{}-foreground.glb", level_data.level_name);
-    file_util::create_dir_if_needed_for_file(fore_file_path);
+                          game_version_names[config.game_version] / level_data.level_name;
     save_level_foreground_as_gltf(level_data, art_group_data, fore_file_path);
   }
   file_util::write_text_file(entities_folder / fmt::format("{}-actors.json", level_data.level_name),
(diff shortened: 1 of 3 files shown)


decompiler/level_extractor/extract_level.cpp
```diff
@@ -323,8 +323,7 @@ void extract_common(const ObjectFileDB& db,
 
   if (config.rip_levels) {
     auto file_path = file_util::get_jak_project_dir() / "glb_out" /
-                     game_version_names[config.game_version] / "common.glb";
-    file_util::create_dir_if_needed_for_file(file_path);
+                     game_version_names[config.game_version] / "common";
     save_level_foreground_as_gltf(tfrag_level, art_group_data, file_path);
```
xTVaser, line 327: dont get rid of the `create_dir_if_needed...` call here.  That is what prevents the issues of things not working if the directory doesn't already exist -- unless this was moved elsewhere (i dont see it)

decompiler/level_extractor/extract_level.cpp
```diff
@@ -361,14 +360,12 @@ void extract_from_level(const ObjectFileDB& db,
...
-                          game_version_names[config.game_version] /
+                          game_version_names[config.game_version] / level_data.level_name /
                           fmt::format("{}-background.glb", level_data.level_name);
     file_util::create_dir_if_needed_for_file(back_file_path);
     save_level_background_as_gltf(level_data, back_file_path);
     auto fore_file_path = file_util::get_jak_project_dir() / "glb_out" /
-                          game_version_names[config.game_version] /
-                          fmt::format("{}-foreground.glb", level_data.level_name);
-    file_util::create_dir_if_needed_for_file(fore_file_path);
+                          game_version_names[config.game_version] / level_data.level_name;
```
xTVaser, line 368: same here, explicitly ensure the directory you are about to save a file to, exists.

xTVaser (changes requested):

decompiler/level_extractor/extract_level.cpp
dallmeyer: I just moved the create_dir_if_needed call into `save_level_foreground_as_gltf` right before we actually write the file. Since it now writes to multiple different filenames, I only have the prefix at this point - and it didn't seem to work properly if there wasn't a filename after the deepest folder in the path. For example, using ".../jak2/common" as the path would only create up to the jak2 folder and not the common folder, and then it would fail to write the common glbs because the folder wouldn't exist.

decompiler/level_extractor/fr3_to_gltf.cpp
```diff
@@ -1026,30 +1021,38 @@ void save_level_background_as_gltf(const tfrag3::Level& level, const fs::path& g
...
-                            true,   // embedImages
-                            true,   // embedBuffers
-                            true,   // pretty print
-                            true);  // write binary
+    add_merc(level, art_data, mmodel, model, tex_image_map);
+
+    model.asset.generator = "opengoal";
+
+    auto glb_file = glb_path / fmt::format("{}.glb", mmodel.name);
+    file_util::create_dir_if_needed_for_file(glb_file);
```
dallmeyer, line 1049: create_dir_if_needed here now

xTVaser: ah my bad, didn't see it in the diff

xTVaser:
ok to test

xTVaser:
add to whitelist

xTVaser (approved):

PR #3633, opened 2024-08-11
jak x: preliminary type definition decomp work

VodBox:
Still very early, has plenty of definitions or hacks copied over that haven't been fully checked or produce new errors/warnings, but slowly working my way through types and checking them against relevant files as I go.

Not intended to commit anyone to Jak X with this, but figured doesn't hurt to have something started.


xTVaser:
work merged into master via https://github.com/open-goal/jak-project/pull/4112

PR #3646, opened 2024-08-25
Create `abutton-idx` enum, replace magic numbers across codebase

dallmeyer:
Found out the hard way that the `abutton` array for pressure sensitivity uses a different order than the `pad-buttons` enum, so I created a new enum for abutton indexing.

I replaced any magic numbers across the codebase where it made sense, e.g.
`(-> *cpad-list* cpads (-> self control unknown-cpad-info00 number) abutton 6)`
becomes
`(-> *cpad-list* cpads (-> self control unknown-cpad-info00 number) abutton (abutton-idx x))`


goal_src/jak1/engine/ps2/pad.gc
```diff
@@ -156,22 +171,19 @@
...
     gp-0))
 
- ;; Allocates enough input frames to support counting 50ms of time at 300fps.
+;; Allocates enough input frames to support counting 50ms of time at 300fps.
 (deftype cpad-history (basic)
   ((button0-abs pad-buttons 15) ;; bitmask of buttons, pressed or not, with history
    (button0-rel pad-buttons 15) ;; bitmask of if button going down.
-  )
-)
+   ))
```
dallmeyer, line 178: All the whitespace changes should just be from running the formatter on `goal_src/jak1`

dallmeyer: Looks like lint is complaining that this changed line numbers for a handful of `;; og:preserve-this` comments

xTVaser (approved):

PR #3649, opened 2024-08-29
Use custom normals in model replacement!

The-Stellar:
A small change that makes the models look a lot better. 
Will likely cause a crash if imported model has no normal data, but that's on the user to fix on their model.


xTVaser:
ok to test

xTVaser:
ok to test

xTVaser:
ok to test

PR #3651, opened 2024-09-03
Fix github release-pipeline so version tag is available for cmake

dallmeyer:
Followup to https://discord.com/channels/756287461377703987/1280601431975661599 where the v0.2.16 release is showing git SHA in the speedrun verification text, instead of the version tag

### DIFF
--- a/.github/workflows/linux-build-clang.yaml
+++ b/.github/workflows/linux-build-clang.yaml
@@ -23,6 +23,9 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
 
       - name: Install Package Dependencies
         run: |
--- a/.github/workflows/linux-build-clang.yaml
+++ b/.github/workflows/linux-build-clang.yaml
@@ -21,7 +21,14 @@ jobs:
     timeout-minutes: 60
 
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/linux-build-clang.yaml
```diff
@@ -23,6 +23,9 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
```
xTVaser, line 28: doing a full checkout adds ~10s to this step comparing to older runs.  

I think we should add it to it's own step, ie. `Checkout Repository with Tags` that has an `if: ${{ inputs.uploadArtifacts }}` and then this step can have an `if: !${{ inputs.uploadArtifacts }}`.  This is so it only runs when it is required on a release workflow.

xTVaser (changes requested):

.github/workflows/release-pipeline.yaml
```diff
@@ -45,6 +45,8 @@ jobs:
 
   # Windows
   build_windows_clang:
+    needs:
+      - cut-release
```
xTVaser, line 49: s/cut-release/cut_release

dallmeyer: ah good catch

xTVaser (approved):

PR #3652, opened 2024-09-05
[jak2] Custom speedrun category continue select

dallmeyer:
WIP - need to consolidate the existing `continue_point_name` with the `continue_point` field I added. 

Right now its completely functional, but `continue_point` is persisted as a u32 in the json - would be nice if it could be the string literal instead like `continue_point_name`.

### DIFF
--- a/goal_src/jak2/pc/features/speedruns.gc
+++ b/goal_src/jak2/pc/features/speedruns.gc
@@ -16,6 +16,58 @@
 (set! (-> *speedrun-info* active-practice-objective start-zone-init-params) (new 'static 'objective-zone-init-params))
 (set! (-> *speedrun-info* active-practice-objective end-zone-init-params) (new 'static 'objective-zone-init-params))
 
+;; dynamically populate continue-point list, for custom category menu
+(define *main-continue-point-list* '())
+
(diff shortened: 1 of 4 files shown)


goal_src/jak2/pc/features/speedruns.gc
```diff
@@ -16,6 +16,58 @@
...
+             (continue-ptr (-> lvl continues)))
+        (when (is-main-level? (-> lvl name))
+          ;; loop thru level's continues
+          (while (not (null? continue-ptr))
+            (let ((cont (the-as continue-point (car continue-ptr))))
+              (when (is-main-continue? cont)
+                ;; this continue is valid, push to list
+                (cons! *main-continue-point-list* (-> cont name))))
+            (set! continue-ptr (cdr continue-ptr)))))
+      (set! lvl-ptr (cdr lvl-ptr)))))
```
dallmeyer, line 53: Since the `popup-menu-dynamic-submenu` uses iterates their lists by index, I figured it was better to flatten these once, rather than traverse the entire `*level-load-list*` every time we want to lookup continue points by index

dallmeyer:
closing old pr, not really needed

PR #3690, opened 2024-09-30
[jak1] Patch "NG+ glitch"

### LINKED ISSUE #3644: Hub 2 100% warp gate cutscene bug
### Acknowledgements

- [X] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak1+sort%3Aupdated-desc+) for duplicate or closed bug reports

### Describe the Bug

When i reset the hub 2 100% category extension using speedrun options menu from precursor basin, the cutscene trigger around the blue sage hut stays active after the reset as if the button is pressed and the cutscene hasn't played yet. 

### How To Reproduce

Start a hub 2 100% category extension run from the speedrun options menu
Press the warp gate button
Go to Precursor Basin 
Reset current speedrun
Approach the blue sage hut

This bug does not occur from anywhere else in hub 2 as far as i know. I do have a video of this but the file size is too large to attach here. I can be contacted on discord @sasqunjico for the video. 

Yes, it's unique to OpenGOAL

### Expected Behavior

When i reset, the warp gate should also reset.

### Environment Information

[opengoal-support-package.zip](https://github.com/user-attachments/files/16736638/opengoal-support-package.zip)

### Game Version

NTSC 1.0 (black label)

No

dallmeyer:
Fixes #3644 which I believe is the same underlying issue as "NG+ glitch"

To reproduce the issue in #3644 you can:
- choose Hub 2 100% in the speedrun fast reset menu
- hit the blue sage warp gate switch
- go deep enough into any adjacent level (e.g. basin) where `village2` display is turned off
- reset speedrun in the fast reset menu
- tasks are reset but the switch will be pressed, giving the cutscene early

You can also grab orbs/scout flies in `village2`, and they won't be reset properly because of this same bug.

It happens because of the way entity perm status is managed across both `level` vs `game-info` objects.
- when `village2` is deactivated (still loaded but display hidden), its entity perms are copied to `game-info`'s `perm-list` 
  - this is how we persist `warp-gate-switch-7` being pressed if village2 is ever unloaded
- during the speedrun reset `reset-actors` is called:
  - any active levels (loaded+displayed) have their entity perm statuses reset
    - because `village2` is not displayed yet, its entity perm statuses are not touched
  - `game-info` is re-initialized, clearing out its `perm-list`
  - continue is set to `firecanyon-end`
    - this unloads `rolling` or whatever to make room for `firecanyon`
    - `village2` is already loaded, and just gets displayed
      - at this point the game does copy entity perm status from `game-info` back to the `village2` level
      - but we reset the game, so it has no data about the warp-gate-switch, leaving it pressed!

### DIFF
--- a/goal_src/jak1/engine/level/level.gc
+++ b/goal_src/jak1/engine/level/level.gc
@@ -623,6 +623,12 @@
      ;; copy data from the level to the game-info storage.  This will remember permanent level stuff, like
      ;; what you collected/completed.
      (copy-perms-from-level! *game-info* this)
+     ;; og:preserve-this fully clear entity perm status in the level itself (based on reset-actors)
+     ;;  it should be copied back out of game-info on birth to prevent "NG+ glitch"
+     (let ((lev-ents (-> this entity)))
+       (dotimes (idx (-> lev-ents length))
+         (let ((ent (-> lev-ents data idx entity)))
+           (update-perm! (-> ent extra perm) 'game (the-as entity-perm-status 1919)))))
      (send-event *camera* 'level-deactivate (-> this name))
      (send-event *target* 'level-deactivate (-> this name))
      ;; remove this BSP from the engine. This will stop us from being drawn.


dallmeyer:
I also tried an alternative of letting `reset-actors` run on `loaded` levels in addition to `active` ones, but it caused crashes that I couldn't quite make sense of.

ManDude:
What's the meaning of the number? Is there no way to get the value that's used to initially set the perm status?

dallmeyer:
Not sure offhand, I just borrowed this from `reset-actors`
https://github.com/open-goal/jak-project/blob/fedfb6fd094403cd3bdda78cce3dd81a39c7b4eb/goal_src/jak1/engine/entity/entity.gc#L985-L992

dallmeyer:
Looks like its every bit in `entity-perm-status` besides `bit-7`?

ManDude:
Strange. I think it would be better if you called `update-perm!` like that function does, if that's where you're copying it from.

PR #3692, opened 2024-09-30
Parameterize the iso_data folder for goalc

massimilianodelliubaldini:
I hope this is everything I needed, and nothing I didn't.

## What's Changed

This update adds a command-line parameter to goalc, `--iso-path`. Providing a path to a directory like `D:\Files\Repositories\ArchipelaGOAL\iso_data\jak1` will inform the compiler to use that directory instead. 

## Why is this useful?

When combined with `--proj-path`, the compiler can be pointed to a completely different project folder, given the `(mi)` command, and immediately begin building from that directory, with everything it needs. This eliminates the need to copy `iso_data` to multiple `data` directories.

If a subsequent change to the Launcher is made, each mod could be passed an --iso-path pointing to a single shared folder, allowing mods to each run their own REPL _without_ requiring a copy of `iso_data` in a subfolder.

## Independent testing required!

My local repositories are a little suspect, with a mod, a fork of mod-base, and a fork of jak-project, all on the same drive. My decompiler_out and iso_data folders are in the mod repo, not mod-base nor jak-project. So what I did was make the change in the mod-base fork, point `--proj-path and --iso-path` to the mod folders, and then ran `(mi)`. The output showed a build starting with no errors. 

Then I had to create this PR, which my fork of mod-base is unable to do, so I created a patch file, forked jak-project, then applied the patch there. 

All this is to say that it would be preferable if someone could apply this code to their own installation and see if it works. Even I wouldn't take my own word for this.

### DIFF
--- a/common/repl/repl_wrapper.cpp
+++ b/common/repl/repl_wrapper.cpp
@@ -43,8 +43,14 @@ void Wrapper::print_welcome_message(const std::vector<std::string>& loaded_proje
       fmt::format("  Project Path: {}\n",
                   fmt::format(fg(fmt::color::gray), file_util::get_jak_project_dir().string()));
   message += fmt::format(fmt::emphasis::bold | fg(fmt::color::orange), "  -       :===:       -");
+  std::string effective_iso_path;
+  if (repl_config.iso_path.empty()) {
+    effective_iso_path = file_util::get_file_path({"iso_data"});
+  } else {
+    effective_iso_path = repl_config.iso_path;
+  }
   message +=
-      fmt::format("  ISO Data Path: {}\n", fmt::format(fg(fmt::color::gray), repl_config.iso_path));
+      fmt::format("  ISO Data Path: {}\n", fmt::format(fg(fmt::color::gray), effective_iso_path));
   message += fmt::format(fmt::emphasis::bold | fg(fmt::color::orange), "  --.   .--: :--.   .--");
   message += "  nREPL:";
   if (!nrepl_alive) {
@@ -291,7 +297,6 @@ REPL::Config load_repl_config(const std::string& username,
       // do nothing
     }
   }
-  loaded_config.iso_path = file_util::get_iso_dir_for_game(game_version).string();
   loaded_config.temp_nrepl_port = nrepl_port;
   return loaded_config;
 }
--- a/common/repl/repl_wrapper.cpp
+++ b/common/repl/repl_wrapper.cpp
@@ -45,8 +45,8 @@ void Wrapper::print_welcome_message(const std::vector<std::string>& loaded_proje
   message += fmt::format(fmt::emphasis::bold | fg(fmt::color::orange), "  -       :===:       -");
   message +=
       fmt::format("  ISO Data Path: {}\n",
-        fmt::format(fg(fmt::color::gray),
-        file_util::get_iso_dir_for_game(repl_config.game_version).string()));
+                  fmt::format(fg(fmt::color::gray),
+                              file_util::get_iso_dir_for_game(repl_config.game_version).string()));
   message += fmt::format(fmt::emphasis::bold | fg(fmt::color::orange), "  --.   .--: :--.   .--");
   message += "  nREPL:";
   if (!nrepl_alive) {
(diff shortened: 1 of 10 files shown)


xTVaser:
ok to test

goalc/compiler/Compiler.cpp
```diff
@@ -21,11 +21,12 @@ using namespace goos;
...
                    const std::optional<REPL::Config> repl_config,
                    const std::string& user_profile,
-                   std::unique_ptr<REPL::Wrapper> repl)
+                   std::unique_ptr<REPL::Wrapper> repl,
+                   const std::string& iso_path)
     : m_version(version),
       m_goos(user_profile),
       m_debugger(&m_listener, &m_goos.reader, version),
-      m_make(repl_config, user_profile),
+      m_make(repl_config, user_profile, iso_path),
```
xTVaser, line 29: rather than adding a new argument, this looks like something that could be set on the `repl_config`, it's something that is already passed along to the `m_make` instantiation as well.

xTVaser (commented):
Tested the main use-case on the launcher, it seems to work https://github.com/open-goal/launcher/pull/582

Just one suggestion around cleaning up some of the parameter passing changes.

~~The compilation check failure is a weird one....i cant really explain what's going on there.  I would be surprised if it was related to the PR coming in from a fork but that would be my first guess.~~
- Nevermind i think it makes sense, I think you uncovered a flaw in that test since the binaries are dynamically linked, and this PR changed a function signature.  ~~I'll have to fix that.~~ Cherry pick this commit https://github.com/open-goal/jak-project/pull/3697/commits/dc957f60fadcadc417676c94a6e90fd24651f955

goalc/compiler/Compiler.cpp
massimilianodelliubaldini: I wrote the latest commit to handle this.

massimilianodelliubaldini:
With No Paths:
image

With Proj Path only (note: I have my own copy of iso_data inside this mod directory):
image

With Proj Path and ISO Path:
image

Compiling with the last:
image

common/repl/repl_wrapper.cpp
```diff
@@ -289,6 +291,7 @@ REPL::Config load_repl_config(const std::string& username,
       // do nothing
     }
   }
+  loaded_config.iso_path = file_util::get_iso_dir_for_game(game_version).string();
   loaded_config.temp_nrepl_port = nrepl_port;
   return loaded_config;
```
xTVaser, line 296: this is odd behaviour -- you support loading the `isoPath` from the json file, but then set it here.  This means that if someone wanted to set it in the file, their value gets immediately clobbered.  This also has the problem of this line of code only executing if the user _doesnt_ have a `repl-config.json` (see the return above).

I think this line should be removed.  in `main.cpp` you should set this field on the struct there if `iso_path_override` is provided.

This also means that this value would only be set when it's actually set via the CLI flag (right now `iso_path` is always non-empty if you don't have a `repl-config.json`, so it always executes your new code path in `MakeSystem.cpp`)

xTVaser (changes requested):

common/repl/repl_wrapper.cpp
xTVaser: Also reminder to copy in this fix to resolve the CI issue https://github.com/open-goal/jak-project/commit/dc957f60fadcadc417676c94a6e90fd24651f955

massimilianodelliubaldini: Wanna make sure I have this right before committing.

Here I removed the line you mentioned.
image

Then here, I have to move our `iso_path_override` check after the instantiation of `repl_config`, and we set the `repl_config.iso_path`. 
image

So if a user has a repl_config _and_ a CLI override, the override takes priority. If they have a repl_config with no override, the value in the config applies.

Is there anything I need to do to handle a situation where the config has no value and we don't have an override?

xTVaser: > Is there anything I need to do to handle a situation where the config has no value and we don't have an override?
There shouldn't be, you're later code checks if `iso_path` is empty, and if so you continue to do what `MakeSystem.cpp` was already doing prior to this PR.

massimilianodelliubaldini: Ok, I'll test later tonight/tomorrow.

xTVaser (approved):
Seems to work properly for the vanilla installation, and for mods end-to-end.  Thanks!

PR #3704, opened 2024-10-05
[jak3] fix envmap texture and strength

water111:
Fix the envmap texture using the eye texture by default (will require a re-extract to fix), fix envmap strength in all three games being off by 2x. (oops)

With envmap scale fix:
image

without envmap scale fix (but with texture fix):
image


water111:
https://github.com/user-attachments/assets/a682054d-27d5-474f-90b4-8ba51055ecec

water111:
we decided to not change the envmap scaling since it looks closer to the ps2 version. (but not the ps3 version?).

PR #3740, opened 2024-11-02
[jak2] Fix bug where spinning into grind doesnt break clasps in dig1

dallmeyer:
if you sideflip into grind in the digsite mission, the clasps wont break because it is considered a `board-spin` attack instead of `board`. This extends the condition to allow for `board-spin` as well.

~~Should I wrap this in an `#if PC_PORT`?~~ done

### DIFF
--- a/goal_src/jak2/levels/dig/dig-digger.gc
+++ b/goal_src/jak2/levels/dig/dig-digger.gc
@@ -657,7 +657,10 @@
       (('attack)
        (let ((v1-2 (the-as attack-info (-> block param 1))))
          (when (and (logtest? (-> v1-2 mask) (attack-mask mode))
-                    (or (= (-> v1-2 mode) 'board) (= (-> v1-2 mode) 'board-spin))) ;; og:preserve-this fix bug where spinning into grind doesnt break
+                    ;; og:preserve-this fix bug where spinning into grind doesnt break
+                    (#if PC_PORT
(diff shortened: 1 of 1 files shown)


ManDude:
Yes

ManDude:
However, I'm not sure we should be pushing this. I don't know if this is unintended behavior or not and it's not really the cause of the clasps not breaking even when you grind the rail next to them.

dallmeyer:
I hear the point about not changing intended vanilla behavior of course, though I'm not sure what you mean about this not being the cause, maybe we are talking about different things?

If you grind across the rails the clasps break as intended, but if you spin straight into the same exact grind, they don't break. I can't imagine why that would be intentional

ManDude:
"I can't imagine why that would be intentional"

The game specifically asks you to grind on them:
image

"Daxter: Grind on the support bases to break the cables."

Zedb0T:
Yeah, but you do grind into them, the game just incorrectly counts it as a board-spin if you spin into the grind no?

Hat-Kid:
When you do a trick right before starting to grind on the rail, the cables don't break is what Barg was getting at.

xTVaser:
Yeah to clarify it a different way, when you do a 360 spin your boost gets "stored" and if you are in this state, the cables don't break.

dallmeyer:
@ManDude I added a `(focus-test? *target* rail)` condition for board-spin so it should only change the behavior if you're actually grinding the rail 👍

ManDude (approved):

PR #3751, opened 2024-11-06
docs: Remove instructions to add go-task tap to Homebrew.

mrpippy:
Homebrew has its own formula for go-task which is up-to-date and working.


xTVaser (approved):
Thanks!

PR #3752, opened 2024-11-07
Taskfile: Call 'python3' on non-Windows platforms

mrpippy:


xTVaser (changes requested):
Your changes will break other environments in order for it to work properly on the default homebrew setup, you'd need to extract this into the `Taskfile_darwin.yml` file.

I'm not sure if this is definitely the right thing to do either, as even on macOS you can setup your python environment in a myriad of ways so that `python` is python3 (ie. `pyenv` and since monterary python2 is finally been removed).  Considering it's the default for homebrew I'd merge it but the change has to be isolated from windows and linux.

mrpippy:
Is there any environment where `python3` doesn't launch Python 3? Out of all the environments I have handy (Gentoo Linux, FreeBSD, OpenBSD, macOS), `python3` works on all of them.

Plain `python` only works on Gentoo, and I don't think that's universal for Linux distros either (doesn't Ubuntu need `python-is-python3` installed?)

VodBox:
The default installation for Python on Windows does not add a ``python3`` to the PATH. It does have the ``py`` launcher, where you can reliably call on Python 3.X with ``py -3``, but I believe that is not a default on the Linux packages.

xTVaser:
`python` is also what `pyenv` will use out of the box.

xTVaser:
Do you plan to make the change or should this be closed?

mrpippy:
I don't think using a `Taskfile_darwin.yml` is the right solution. This is an issue on many Linux/Unix OSes.

[PEP 394](https://peps.python.org/pep-0394/) makes clear that distributors are free to not include a `python`, but `python3` will always be present (excepting Windows). In addition, `python` may be Python 2--do any of the scripts work with that? `python3` is obviously guaranteed to be Python 3.
If there's really a desire to use `python`, maybe a [dynamic variable](https://taskfile.dev/usage/#dynamic-variables) could be used which is set to `python` if it's present, and `python3` if not? I haven't tested this yet to see if it's possible.

As for Windows, maybe that dynamic variable could be special cased to `python.exe` or `py.exe`?

xTVaser:
using the `Taskfile_{OS}.yml` files is already how this exact problem is being handled successfully. For example calling our built binaries with a `.exe` extension or not.  You just have to make it `python3` for linux and macos, and `python` for windows.

mrpippy:
Ah I hadn't noticed the scripts/tasks/Taskfile_ files. I moved those vars into the main file (to remove the Mac/Linux duplication), and the Windows file overrides them.

xTVaser:
Can you revert the removal?  I'd rather there be a bit of duplication so that they exist when things need to be different between the two environments.

xTVaser (approved):

PR #3761, opened 2024-11-15
jak3: speedrunner mode

Hat-Kid:
Base implementation of the popup menu and speedrunner mode in Jak 3. Autosplitter is untested because I'm on Linux.

Also a couple of other misc changes:

- Model replacements can now have custom bone weights. Needs the "Use Custom Bone Weights" property (provided by the OpenGOAL Blender plugin) enabled in Blender.
- Better error message for lump syntax errors in custom level JSON files.


dallmeyer:
works on Windows 👍 

few minor things:
- hero mode runs usually have a bunch of secrets unlocked from the start, we should probably enable those? infinite ammo, infinite ecos(?), turbo jetboard, etc
- jak1/2 auto skip the intro cutscene when resetting, so probably we should do the same here?
- blaster mod splits after you switch weapons a few times, when `arena-fight-1-resolution` closes. should maybe be on `arena-fight-1-throne` instead, which is after the cutscene? (there are probably a few cases like this throughout the run)

goal_src/jak3/pc/features/speedruns.gc
```diff
@@ -72,6 +72,76 @@
...
   (set! (-> this category) category))
 
+(defconstant HERO_MODE_SECRETS
+  (game-secrets hero-mode
+                endless-ammo
+                invulnerable
+                endless-dark
+                endless-light
+                unlimited-turbos
+                vehicle-hit-points
```
Zedb0T, line 82: This is not taken in hero mode runs because you want the car to die fast.
https://youtu.be/AnrHvAayHmE?t=227

PR #3771, opened 2024-11-19
Remove 128 character buffer causing long filenames to crash gk

massimilianodelliubaldini:
Several users have reported that ArchipelaGOAL is not launching properly, even when using the OpenGOAL Launcher. The window pops up with a black screen, and then quits. The only way they can run it is if they double click gk.exe. 

This comes down to the Launcher providing gk.exe with the `config_path` parameter, which leads the program to find `archipelagoal-settings.gc`. Here is an example from me:

```
D:\Applications\Games\OpenGOAL\features\jak1\mods\JakMods\_settings\archipelagoal\OpenGOAL\jak1\settings/Mods/archipelagoal-settings.gc
```

If a user's base OpenGOAL install directory is long enough, this path becomes longer than 128 characters. This overflows the character buffer in `kopen` which is used to open file streams. If you're only slightly over the limit like myself, at 135 characters, you may not have noticed a problem. But some users have paths a little longer, like 168 characters, and they report the issue is consistent.

Water111 suggested we remove the 128 character buffer and use the filename data directly. This fix requires no changes to the Launcher, just to the kernel, and every mod could stand to benefit from this fix.

### DIFF
--- a/game/kernel/jak1/kmachine.cpp
+++ b/game/kernel/jak1/kmachine.cpp
@@ -391,14 +391,12 @@ u64 kopen(u64 fs, u64 name, u64 mode) {
   file_stream->name = name;
   file_stream->flags = 0;
   lg::print("****** CALL TO kopen() ******\n");
-  char buffer[128];
   // sprintf(buffer, "host:%s", Ptr<String>(name)->data());
-  sprintf(buffer, "%s", Ptr<String>(name)->data());
   if (!strcmp(info(Ptr<Symbol>(mode))->str->data(), "read")) {
-    file_stream->file = sceOpen(buffer, SCE_RDONLY);
+    file_stream->file = sceOpen(Ptr<String>(name)->data(), SCE_RDONLY);
   } else {
     // 0x602
-    file_stream->file = sceOpen(buffer, SCE_TRUNC | SCE_CREAT | SCE_WRONLY);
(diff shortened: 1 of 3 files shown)


massimilianodelliubaldini:
## Before Test

I verified the issue is reproducible with the following command:

```
D:\Files\Repositories\ArchipelaGOAL-But-Its-Way-Too-Long-For-Windows\out\build\Release\bin\gk.exe -v --game jak1 --config-path D:\Files\Repositories\ArchipelaGOAL-But-Its-Way-Too-Long-For-Windows\out\build\Release\_settings\archipelagoal -- -boot -fakeiso
```

Here are the last few console lines, showing the app halting right as it tries to run `kopen`.

```
Resetting archipelago settings                                                           x1cd5a31
[Archipelago] Settings found at 'D:\Files\Repositories\ArchipelaGOAL-But-Its-Way-Too-Long-For-Windows\out\build\Release\_settings\archipelagoal\OpenGOAL\jak1\settings/Mods/archipelagoal-settings.gc'... loading!                                                         -For-Windows\out\build\Release\_settings\archipelagoal\OpenGOAL\jak1\settings/Mods/archipelagoal-settings.gc'... loading!  
****** CALL TO kopen() ******
D:\Files\Repositories\ArchipelaGOAL-But-Its-Way-Too-Long-For-Windows [main ↑78]>
```

xTVaser:
add to whitelist

water111 (approved):
Thanks!

PR #3791, opened 2024-12-03
Add option to hide the imgui bar with a click event

Zedb0T:
Should cover edge cases where the bar is displayed and a keyboard type device is not connected to the system (Steam deck/touch input based device instead).

While it should probably be impossible to display the bar in the first place on such devices it apparently isnt -> https://www.reddit.com/r/jakanddaxter/comments/1h4mht1/how_to_remove_grey_bar_at_top_on_steamdeck/

And even ignoring the whole steamdeck/no keyboard thing adding another way to hide it that is visible on the bar should help reduce people asking about how to hide it. (And hopefully help people who dont ask/dont know where to ask!)

https://github.com/user-attachments/assets/7f132781-675c-451e-8b39-2a78fa098b9e


Zedb0T:
Just to be clear, I did make it match the rest of the bar after recording the example video, but maybe it was a good thing that it had extra emphasis on it?
image

PR #3801, opened 2024-12-10
Start resolution menu at current resolution 

### LINKED ISSUE #1623: jak1: progress menu lists dont open with the current selected option focused
For example, if i currently have `1280x720` selected as my resolution, the list still opens with `2560x1440` focused / highlighted as it is the first in the list.

This is a little confusing from a UX perspective as theres no way to know your currently selected option by consulting the menu.

image

Zedb0T:
Also run cpp and goal formatter

Fixes #1623


xTVaser:
This doesn't really fix #1632 because it only applies to the resolution menu.  Unless I am misremembering and this is the _only_ menu like this in Jak 1?

I do have changes in progress that are getting close to fully refactoring Jak 1's menu code (and fixing that along with 4+ other issues)

In the future, you might want to put full-project formatting changes in a separate PR as well, makes something simple much harder to review.

Zedb0T:
Ah alright, I tired to split it out into two commits my apologies.

Zedb0T:
Also yeah I guess I got baited by the discussion in the issue. I see now that its more of a global scope and its probably better to make all menus behave in this manner somehow.

Hat-Kid:
Also don't run the GOAL formatter on `textures.gc` or `tpages.gc`. The formatter currently inserts newlines between every top-level statement so it just inflates the line count for those files.

xTVaser:
You didn't have to close it, if it works and is an improvement.  But up to you.

Zedb0T:
Nah, I'm only doing it to save you guys time and to learn a bit. If you have something in progress that fixes this and more, then I am not saving any time. And I've already succeeded on the learning part 👍

PR #3813, opened 2024-12-27
Add collision viewer for Jak 1 Custom levels

Zedb0T:


game/graphics/opengl_renderer/OpenGLRenderer.cpp
```diff
@@ -1304,7 +1304,8 @@ void OpenGLRenderer::dispatch_buckets_jak1(DmaFollower dma,
     m_category_times[(int)m_bucket_categories[bucket_id]] += bucket_prof.get_elapsed_time();
 
     // hack to draw the collision mesh in the middle the drawing
-    if (bucket_id == 31 - 1 && Gfx::g_global_settings.collision_enable) {
+    if (bucket_id == (int)jak1::BucketId::MERC_PRIS_LEVEL0 &&
+        Gfx::g_global_settings.collision_enable) {
```
water111, line 1308: I think moving this here will break shadows on the collision geometry but I'm not sure. What if you put `GENERIC_ALPHA` as the bucket instead?

goal_src/jak1/pc/debug/default-menu-pc.gc
```diff
@@ -227,11 +227,11 @@
     (cond
       ((-> arg value)
        (false! (-> arg value))
-       (logior! *vu1-enable-user-menu* (vu1-renderer-mask tfrag trans-tfrag tie tie-near))
+       (logior! *vu1-enable-user-menu* (vu1-renderer-mask tfrag tie tie-near))
```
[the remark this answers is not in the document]
Zedb0T: When I did my binary search to find out where this worked, the bucket that it draws during is in trans-tfrag, and removing it from this call did not change anything else that I noticed during my testing.
image
image

game/graphics/opengl_renderer/OpenGLRenderer.cpp
Zedb0T: image
I binary searched the enum and found an entry that "worked" - I did a test through hub 1 and didn't notice any issues after my change but I can do another test and pay attention to shadows specifically.

PR #3822, opened 2025-01-04
[jak2, jak3] Unstretch moon and sun

water111:
Fix the issue where the moon gets wide when using PC aspect ratios

image


ManDude:
Thank you!

Issue #233, opened 2021-02-05
[Compiler/Low-Priority] Incorporate VU operation `dest` mask into the op-name itself


xTVaser:
It would be nice if rather than:
```lisp
(.add.vf vf1 vf2 vf3 :mask #b1110)
```
We could do
```lisp
(.add.yzw.vf vf1 vf2 vf3
```

However, in order to avoid code complexity, this needs to be supported using some kind of macro.

PR #3824, opened 2025-01-04
gsrc: Add convenience macros for masking VF operations

xTVaser:
Migrates all code, there should be no change in the compilation output (linter should check this)

At first i was considering making these builtins, which short of a bunch of code generation, would require some sort of dynamic definition in `Atoms.cpp`.  This isn't hard but, i figured it would be better to keep it simple and just generate the OG macros.

Fixes https://github.com/open-goal/jak-project/issues/233


water111 (approved):
awesome, this is much nicer to read!

The `vf-macros.gc` file approach seems fine to me

Issue #3833, opened 2025-01-08
[jak1] Palm trees and other objects are invisible


massimilianodelliubaldini:
### Acknowledgements

- [x] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak1+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [x] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

In v0.2.21, a bug was introduced where certain objects are no longer visible. This includes palm trees in Sentinel Beach, fish on racks near the Fisherman's hut, and the molten rocks of Fire Canyon. They still have collision however.

![Image](https://github.com/user-attachments/assets/80ca6c1e-d9de-4bd7-ae0b-faec7d8d9960)

Barg did the research and pointed to #3825 as introducing the bug.

### How To Reproduce

Walk to Sentinel Beach. There should be a copse of trees right after the first scout fly.

### Does this problem occur on original hardware or PCSX2?

Yes, it's unique to OpenGOAL

### Expected Behavior

Should see palm trees when walking to Sentinel Beach, or molten rocks when walking to Fire Canyon, etc.

### Environment Information

Processor	AMD Ryzen 7 3700X 8-Core Processor                3.59 GHz
Installed RAM	32.0 GB
System type	64-bit operating system, x64-based processor
Edition	Windows 10 Education
Version	22H2
OS build	19045.5247

OpenGOAL Version 0.2.21

### Game Version

NTSC 1.0 (black label)

### Have you set the game to something other than `60fps`?

No

PR #3837, opened 2025-01-12
[jak1] Fix wind

water111:
I accidentally broke this with the camera matrix improvements, but it's fixed now.

image


massimilianodelliubaldini:
Fixes #3833.

PR #3852, opened 2025-01-29
deps: update to SDL3

### LINKED ISSUE #2773: Pressure Sensitive button support

My idea is related to Pressure Sensitive buttons used in the DualShock 2, DualShock 3, and Original Xbox controllers.

Pressing down on the accelerate/brake buttons lightly make the vehicle go/brake slowly, and pressing the accelerate/brake buttons down firmly make the vehicle go/brake faster.

### Additional context

This could be useful for people who have a DualShock 3.

xTVaser:
This PR updates to SDL3, and with it, adds a handful of new features.  Everything seems to work but I'm going to look over the code once last time before merging, some of the API changes are hard to spot.

Fixes #2773

### Pressure sensitivity support for DS3 Controllers

SDL3 adds pressure sensitivity support for DS3 controllers on windows.  I have not tested on linux.  The option is disabled by default.

On windows you will need https://docs.nefarius.at/projects/DsHidMini/ and to be using SXS mode.

### DualSense and Xbox One Trigger Effects

If enabled, Jak 2 will have certain trigger effects.  They are:
   - xbox1:
     - small vibrate when collecting dark eco
     - big vibrate when changing to dark jak
     - vibrate when shooting gun, proportional to gun type
   - ps5:
     - resistance when changing to dark jak
     - different gun shooting effects
       - red (resistance)
       - yellow (weapon trigger)
       - blue (vibrates)
       - purple (less resistance)
> **Gun Shooting effects are only enabled if the new "Swap R1 and R2" option is enabled**

There are more effects that could be used in `dualsense_effects.cpp`, but I only exposed the ones I needed to OpenGOAL.  If a modder wants to use some of the others and wires them up end-to-end, please consider contributing that upstream.

### New ImGUI Menu

Added new imgui options for selecting the active controller, for those people that struggle to select the initial controller.

image

### Testing

The highlights of what I tested successfully:
   - display
     - [x] all mode switch permutations
     - [x] launch with all modes saved
     - [x] switch monitors / unplug monitor that was active, how does it handle it
     - [x] load with alternate monitor saved and all modes
     - [x] allowing hidpi doesnt break macos
   - controls
     - [x] keyboard and mouse still work
     - [x] pressure sensitivity on linux


BreakPoints:
I've done some quick tests in Jak2+3 and haven't found any issues so far. I'll test on some other devices/DPI scalings when I have some time.

The trigger effects are a nice touch! Feel great on an Xbox controller and the Blaster is actually easier to fast fire on a DualSense with the resistance just after the trigger point. Only thing I noticed was the Vulcan Fury vibrates for every shot on Xbox but the DualSense is more of a constant vibration that gets faster.

PR #3855, opened 2025-02-01
[decompiler] `as-type` and font method support

water111:
Add support for `as-type` macro, and detecting inline font methods. This works in all three games but I've only updated jak 3's goal_src for now.  Eventually I will go back and work through the others, but I want to get more decompiler features in first.

image


goal_src/jak1/kernel/gcommon.gc
```diff
@@ -294,9 +294,16 @@
   `((the (current-method-function-type) (find-parent-method (current-method-type) (current-method-id))) ,@args))
 
 (defmacro as-type (this type)
-  "Macro to _safely_ convert to a different type, returning #f if the type doesn't match.
-   Does a runtime type check so it's expensive."
-  `(if (and (nonzero? ,this) (type-type? (-> ,this type) ,type)) (the-as ,type ,this)))
+  "If this is the specified type, return this cast to that type. Otherwise, return #f."
+  (with-gensyms (obj)
+    `(the ,type (let ((,obj ,this))
+                  (if (type? ,obj ,type)
```
Hat-Kid, line 300: Jak 1 doesn't have `type?`, only `type-type?`.

xTVaser:
Probably closes https://github.com/open-goal/jak-project/issues/1280

goal_src/jak1/kernel/gcommon.gc
water111: oops - I should have left this. The decompiler doens't recognize the jak 1 version yet.

PR #3292, opened 2024-01-05
Support extracting collision to obj for jak2/3 formats

dallmeyer:
image
image

PR #3882, opened 2025-03-12
Extract levels+collision to `decompiler_out` (adjacent to textures+audio)

dallmeyer:
should make it nice and easy to jump to the decompiler_out/<game> folder in the launcher and show the user all their exported things.

also adds a taskfile command for ripping audio.

and fixes jak1 collision extract scaling. I [previously fixed this for jak2/3](https://github.com/open-goal/jak-project/pull/3292#discussion_r1442595308) but never fixed for jak1 apparently

before:
image

after:
image


.github/workflows/linux-build-gcc.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
             -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
 
       - name: Build Project
-        run: cmake --build build --parallel $((`nproc`)) -- -w dupbuild=warn
```
dallmeyer, line 55: https://github.com/ninja-build/ninja/pull/2356

this flag is no longer supported as of 1.12.0, looks like we're using 1.12.1 now

Hat-Kid (approved):

Issue #3799, opened 2024-11-25
Installation error - `WATER-AN.CGO` not found, despite ISO being validated.


karumeyaki:
### Acknowledgements

- [x] I have [searched](https://github.com/open-goal/launcher/issues?q=is%3Aissue+is%3Aopen+label%3Abug+sort%3Aupdated-desc) for duplicate or closed bug reports
- [x] I understand that I am supposed to provide my own legitimately obtained copy of the game
- [x] If possible, I've attached the support package zip file to the issue [see how here](https://github.com/open-goal/launcher#asking-for-help)

### Attach Support Package / Log Files (drag and drop)

[opengoal-support-package.zip](https://github.com/user-attachments/files/17895619/opengoal-support-package.zip)

### Describe the Bug

![Image](https://github.com/user-attachments/assets/c57744c0-0ba8-4362-b82f-c9d360d9b2bb)

### How To Reproduce

The ISO file is called JADPL because I don't actually have a CD/DVD driver so I went on a different PC which had one, copied the DVD files into a folder, sent that folder to my PC and from there I used ImgBurn to convert these files into the ISO file... If that is the reason for the whole problem I'm very sorry, I'm new to this whole thing.

I couldn't upload the video where I recreated the issue here for some reason so here's an Imgur link (I sped it up because the limit there is 60 seconds):
https://imgur.com/a/FMpta2e

### Operating System

Windows 11

### Launcher Version

v.2.5.4

### Tooling Version

v.0.2.18

Matty2024:
Hey OP,

Thanks very much for posting your issue. I actually had this installation error previously and there is a very easy fix. 
As @xTVaser mentioned, this is indeed a error during the ISO creation while using ImgBurn and has nothing to to with the Launcher or Tooling version used.

The Solution
- Do NOT transfer files off the CD and then create and ISO from those files. Something unknown goes wrong here with either Windows 11 or ImgBurn and the file "WATER-AN.CGO" gets changed to include an underscore instead of a hyphen. This is most likely a read error when creating a copy from the disk (even though the files still shows a hyphen after the copy.
- Create an ISO **_directly_** from the CD and then run the "Validate Disk" function to compare the CD and the newly created ISO file.

A bubble gum & sticky tape temporary solution is to downgrade your Tooling Kit to v0.2.16 and then the game will install normally.
This however is not the best idea as you always want to be running the latest version for new bug fixes and feature additions.

PR #3890, opened 2025-03-31
Rename `WATER_AN.CGO` to `WATER-AN.CGO` if found during extraction

trippjoe:
During the iso creation process it's possible that this file in will be internally renamed to `WATER_AN.CGO`, despite the external file name displaying `WATER-AN.CGO`. We expect the file to be named `WATER-AN.CGO`.

We can't control how our users create their iso files. We can alleviate future friction regarding this matter by adopting this simple change.


trippjoe:
closes #3799

xTVaser (approved):
I've thought about this for a while / did some further research:
- While it is true that ISO-9660 only allows upper case and underscores, basically no ISO, and especially not DVDs use this standard.  They all use an ISO-9660 extension like `joliet`, or `rock ridge`
- However, ISO-9660 was basically supplanted by UDF, and this is what a proper dump of the games should use (atleast for the jak games).

With that said, I looked through the ISOs of Jak 1 through X, and this is the ONLY instance where they violate the restrictions of 9660.  So even though this is kinda a hack, it isn't really something we should have to solve for the later games anyway.  

It's still unfortunately a hack because of course, the ISO creation programs these people are using can choose to replace that `-` with whatever character it wants.  So checking for `WATER_AN.CGO` is not a sure-fire thing.  The better solution might be to detect if the ISO is not UDF and reject it -- but this is a rabbithole that we don't want to go down.

I'm probably going to adjust our documentation to stop recommending imgburn -- MPF dumps ISOs properly on windows and has a fool-proof interface that shouldn't allow people to use these bad deprecated 40year old formats lol.

PR #3896, opened 2025-04-12
Fix for running extractor without optional game flag

trippjoe:


decompiler/extractor/main.cpp
```diff
@@ -337,6 +339,8 @@ int main(int argc, char** argv) {
           (flag_fail_on_validation && validate_code != ExtractorErrorCode::SUCCESS)) {
         return static_cast<int>(validate_code);
       }
+      data_subfolder = data_subfolders.at(version_info->game_name);
+      game_name = version_info->game_name;
```
xTVaser, line 343: there needs to be a check to see if `version_info` is an empty optional or not

PR #3901, opened 2025-04-21
Toggle Cpp openGL with F12 / Alt + Enter / Anything else

Zedb0T:
This is a successor to #3230


game/settings/settings.h
```diff
@@ -24,11 +36,14 @@ struct DebugSettings {
   std::vector<DebugTextFilter> text_filters = {};
   bool text_check_range = false;
   float text_max_range = 0;
+
   u32 hide_imgui_key = SDLK_LALT;
 
-  void load_settings();
+  std::vector<KeyWithModifiers> toggle_fullscreen_keys = {
+      KeyWithModifiers(SDLK_RETURN, InputModifiers(SDL_KMOD_ALT)) void load_settings();
```
Hat-Kid, line 43: I'm not sure if having multiple binds for this one functionality is necessary, if it's already configurable, just having one is fine.

Also, looks like you are not properly closing the bracket here.

Hat-Kid (changes requested):

iso_data/jak2/.gitignore
```diff
@@ -1,2 +0,0 @@
-*
-!.gitignore
```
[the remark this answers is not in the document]
Zedb0T: Nice catch! This was due to a way that I symlink iso directories locally

game/settings/settings.h
Zedb0T: Yeah I made a mistake when I removed the other default of F12, Ive since resolved this error. Is there a benefit to moving this to a singular key as opposed to *potentially* multiple binds if the user wants?

For me it seems to make sense to support potential multiple binds, but default to 1. This way if the user wants more for whatever reason we can always dynamically support however many. Im not aware of a big tradeoff pro/con with switching to singular bind support. (I would want to add both locally myself)

Hat-Kid: In general, I'd say having more options is not a bad thing, but in this case, I think it might cause more confusion/complications.

With your current changes, this is the only bind that works this way, which is inconsistent with the rest of the options and would just make it more confusing for anyone who wants to edit their binds and wonders why only the fullscreen bind supports multiple possible bindings. While we could probably make it so all binds support multiple variants, that would not only be extra work, it might also potentially break users' already existing custom binds.

game/system/hid/display_manager.cpp
```diff
@@ -300,6 +300,36 @@ void DisplayManager::set_display_mode(game_settings::DisplaySettings::DisplayMod
...
+    case game_settings::DisplaySettings::DisplayMode::Fullscreen:
+      lg::info("Fullscreen\n");
+      lg::info("Switching to Windowed mode...\n");
+      enqueue_set_window_display_mode(game_settings::DisplaySettings::DisplayMode::Windowed);
+      break;
+
+    case game_settings::DisplaySettings::DisplayMode::Windowed:
+      lg::info("Windowed\n");
+      lg::info("Switching to Fullscreen mode...\n");
+      // maybe there is a arguement for bordless here instead?
```
Zedb0T, line 317: @ManDude  Yeah, I noticed that weakness as well and pointed it out in a comment here. That's probably my bad and I should have pointed that out and asked for advice/a proper approach in the PR wasn't really sure the proper etiquette there.

Basically I would just add a flag that defaults to fullscreen, and when we switch to borderless update that flag, then read that flag here and go borderless/fullscreen conditionally depending on its value?

Zedb0T: Oh, I think I figured out a cleaner option without a hacky global flag.

Zedb0T: Whats better/preferred?

Currently doing this and its tested/working well:

```void DisplayManager::toggle_display_mode() {
  const auto current_mode = m_display_settings.display_mode;
  // Store the current prefered fullscreen mode
  if (current_mode != game_settings::DisplaySettings::DisplayMode::Windowed) {
    m_previous_fullscreen_display_mode = current_mode;
  }
  lg::info("Current display mode: ");
  switch (current_mode) {
    case game_settings::DisplaySettings::DisplayMode::Fullscreen:
      lg::info("Fullscreen\n");
      lg::info("Switching to Windowed mode...\n");
      enqueue_set_window_display_mode(game_settings::DisplaySettings::DisplayMode::Windowed);
      break;

    case game_settings::DisplaySettings::DisplayMode::Windowed:
      lg::info("Windowed\n");
      if (m_previous_fullscreen_display_mode ==
          game_settings::DisplaySettings::DisplayMode::Fullscreen) {
        lg::info("Switching to Fullscreen mode...\n");
      } else if (m_previous_fullscreen_display_mode ==
                 game_settings::DisplaySettings::DisplayMode::Borderless) {
        lg::info("Switching to Borderless mode...\n");
      } else {
        lg::info("Switching to unknown preferred mode...\n");
      }
      enqueue_set_window_display_mode(m_previous_fullscreen_display_mode);
      break;

    case game_settings::DisplaySettings::DisplayMode::Borderless:
      lg::info("Borderless\n");
      lg::info("Switching to Windowed mode...\n");
      enqueue_set_window_display_mode(game_settings::DisplaySettings::DisplayMode::Windowed);
      break;

    default:
      lg::info("Unknown display mode!\n");
      break;
  }
}```

However Im not sure if I should change it to a `if` statement, so I can combine both the `DisplayMode::Borderless/Fullscreen cases by checking if its not == windowed, because they both just switch it back to windowed.

File: game/system/hid/display_manager.cpp [hunk 4]
```diff
@@ -300,6 +300,46 @@ void DisplayManager::set_display_mode(game_settings::DisplaySettings::DisplayMod
...
+  if (current_mode != game_settings::DisplaySettings::DisplayMode::Windowed) {
+    m_previous_fullscreen_display_mode = current_mode;
+  }
+  lg::info("Current display mode: ");
+  switch (current_mode) {
+    case game_settings::DisplaySettings::DisplayMode::Fullscreen:
+      lg::info("Fullscreen\n");
+      lg::info("Switching to Windowed mode...\n");
+      enqueue_set_window_display_mode(game_settings::DisplaySettings::DisplayMode::Windowed);
+      break;
```
Hat-Kid, line 315: You can combine the fullscreen and borderless cases as they do the same thing:
```suggestion
    case game_settings::DisplaySettings::DisplayMode::Fullscreen:
    case game_settings::DisplaySettings::DisplayMode::Borderless:
      lg::info("Fullscreen/Borderless\n");
      lg::info("Switching to Windowed mode...\n");
      enqueue_set_window_display_mode(game_settings::DisplaySettings::DisplayMode::Windowed);
      break;
```

Zedb0T: Oh thats actually really cool, I was thinking of converting it to an IF statement instead, this is much better thanks! Noted.

Hat-Kid (approved):

PR #3911, opened 2025-05-12
Hide imgui only on key press, instead of while pressed.

Zedb0T:
The imgui bar is currently toggled repeatedly if the key is held down, this can lead to annoyances where the key is held for a short time and cycles back to "show" or "hide" when that's not what the user wants.
Current behavior: 

https://github.com/user-attachments/assets/5d719364-9949-4ba2-bf72-7d432650a614

Fixed behavior

https://github.com/user-attachments/assets/b816f502-c155-4d86-adb7-41ca1a7a40cf


game/graphics/pipelines/opengl.cpp
```diff
@@ -340,12 +340,16 @@ GLDisplay::GLDisplay(SDL_Window* window, SDL_GLContext gl_context, bool is_main)
...
-  m_input_manager->register_command(CommandBinding::Source::KEYBOARD,
-                                    CommandBinding(Gfx::g_debug_settings.hide_imgui_key, [&]() {
-                                      if (!Gfx::g_debug_settings.ignore_hide_imgui) {
-                                        set_imgui_visible(!is_imgui_visible());
-                                      }
-                                    }));
+  m_input_manager->register_command(
+      CommandBinding::Source::KEYBOARD,
+      CommandBinding(Gfx::g_debug_settings.hide_imgui_key, [&](const SDL_Event& event) {
+        if (event.type == SDL_EVENT_KEY_DOWN && event.key.repeat == 0) {
```
trippjoe, line 346: `if (event.type == SDL_EVENT_KEY_UP) { ...` would toggle on release.

xTVaser:
Any reason the PR is still in draft, or is it good to go?

xTVaser (approved):

PR #3919, opened 2025-05-19
Implement analog stick sensitivity setting, default it to 133%

ManDude:
This is a simple multiplier to the gamepad axis input value received from SDL events. Normally the values it provides cannot satisfy the square range of the stick input. This is usually fine but it might play differently with different controllers and compared to consoles, especially considering the DualShock 1/2 have automatic calibration which works in mysterious ways. The setting is there so any user can adjust it for their controllers.

Saved to and loaded from the input-settings.json file.

133% matches PCSX2's default setting and is generally a good value to map the square stick range within most modern(ish) controllers' circular stick motion.

Progress menu option added to Jak 1 and 2. Setting can be changed from 50% all the way to 200%.

~~Renamed the analog deadzone options to stick deadzone since they don't apply to the other analog buttons and only the (analog, yes) sticks.~~


game/assets/jak1/text/game_custom_text_en-US.json
```diff
@@ -205,7 +205,7 @@
   "1524": "ALL CUTSCENES",
   "1600": "INPUT OPTIONS",
   "1601": "SELECT CONTROLLER",
-  "1602": "ANALOG DEADZONE",
+  "1602": "STICK DEADZONE",
```
xTVaser, line 208: thoughts on walking back this change considering these strings have likely already been translated?  I don't really think this was a major source of confusion.  I understand that technically the pressure sensitive buttons are analog buttons, but conventionally, most people don't refer to them as that.

ManDude: Do they not show up as 'outdated' in crowdin if we change them here?

xTVaser: They would but im just saying that might be causing unnecessary work for several languages

ManDude: I can revert it, but keep in mind this adds a new string as well, so I'm not sure how much work it really adds. 2 strings vs 1.

xTVaser: 1 string is already translated by people that may never come back, that's the difference.

Issue #3009, opened 2023-09-21
Shadow related crash around pegasus enemies


xTVaser:
This isn't a very easy one to repro, I was just moving around the area for a few minutes and eventually got this crash in the shadow2 renderer
```
----------- kill #<level active forestb @ #x825e44> (status active)
Displaying level forestb [display]
Done birth in -8543812
ERROR: SCRIPT: could not spawn task-manager for node #<game-task-node-info fortress-escape-introduction open @ #x1acf6e4> level forexita
WARNING: #<pegasus pegasus-10 running :state idle :flags r   :stack 0/512 :heap 9616/9616 @ #x1e7b54> could not do (ja-channel-push! 1) because it has 5/6 channels.
[22:59] [die] Assertion failed: 'offset + 16 == transfer.size_bytes'
        Source: C:/Users/xtvas/Repos/opengoal/jak-project/game/graphics/opengl_renderer/foreground/Shadow2.cpp:158
        Function: virtual void __cdecl Shadow2::render(DmaFollower &, SharedRenderState *, ScopedProfilerNode &)
```

water111:
The assertion is here: https://github.com/open-goal/jak-project/blob/master/game/graphics/opengl_renderer/foreground/Shadow2.cpp#L158

My guess is that it's not a shadow issue, but instead some memory corruption that happened to write into the DMA buffer. This assertion is basically checking that the pointers/offsets in the DMA data are valid, and if they aren't it probably means that the GOAL side shadow rendering went really wrong, or the DMA buffer data got modified. I'm going to guess it's the second option because the shadow code has been pretty reliable.

PR #3934, opened 2025-06-01
[jak3] Fix crash with shadow in precursor ship

water111:
When the `dp-bipedal`s get blown up by the nuke gun, their bones go to NaNs on the last frame. The shadow renderer doesn't handle this well and draws all possible triangles as single tris. This overflows the vif `unpack` field and triggered an assert when sizes inside the shadow renderer weren't consistent.

My guess is that this works on the real game either because:
- their shadow renderer draws garbage data, but you can't tell because the screen is white from the nuke
- no NaN on PS2 means that the shadow renderer behaved differently, not using all single tris.

As a workaround, if the bones are NaN, the shadow renderer treats them as 0, meaning there is no shadow drawn.


xTVaser:
Hopefully also fixes https://github.com/open-goal/jak-project/issues/3009 but that was always difficult to reproduce

PR #3943, opened 2025-06-03
game: default MSAA to off

xTVaser:
There's a suspicion that MSAA might be what causes the black screen problem for some users, additionally this can cause perf issues for people with really bad PCs so this is probably a better default.

Needs testing before merging, i think this is all inclusive though.

### DIFF
--- a/game/graphics/gfx.h
+++ b/game/graphics/gfx.h
@@ -65,7 +65,7 @@ struct GfxGlobalSettings {
   int game_res_h = 480;
 
   // multi-sampled anti-aliasing sample count. 1 = disabled.
-  int msaa_samples = 2;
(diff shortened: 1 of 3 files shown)


Calinou:
I saw this linked in the Q2 2025 progress report, so I thought I'd mention some extra information just in case.

> There's a suspicion that MSAA might be what causes the black screen problem for some users

This might due to the renderer not resolving MSAA correctly: https://github.com/godotengine/godot/issues/18320

It's especially common when forcing MSAA through NVIDIA Control Panel or other means that are external to the game.

Doing something like this fixes it: https://github.com/godotengine/godot/pull/106267

naturally-intelligent:
Will this be in 4.5?

PR #3965, opened 2025-06-22
Setup for new CKernel

yodaxtah:
Here is the code I'm working from. See this Discord discussion for more details: https://discord.com/channels/756287461377703987/1386044162326921346

The currently stubbed jak3 code builds when I hit "Build all" in Visual Studio, but that's about all I can say about it. Still searching for a way to test the new code without too much overhead.

I've built further upon the code from PR #3633 and updated the branch as it was 10 months behind.


decompiler/config/jakx/pal/type_casts.jsonc
xTVaser, line 1: are these legitimate type casts as part of decompilation or has this (and potentially a lot of this PR) just been copy and pasted from jak 3's

yodaxtah: Not sure what type casts you mean, but yes, about all code up until 246fb8f is copy-pasted from jak3. I wanted to start off with something that builds. So this is intended as a draft branch -- did I miss an option to mark it as such?

I just pushed some commits to move over updated code from Jak X's kernel: https://github.com/yodaxtah/jakx-c-kernel-decompiled/tree/main/elf/kernel
They appear above your comment.

As I don't see any other way to check for bugs, I'm currently just adding these updates one by one and rebuilding. I guess I'll just deal with the run-time problems afterward then. Fortunately, the code is very similar ^^

PR #3968, opened 2025-06-29
Assert  - added missing macros

clagrisenti:
Compilation with the NO_ASSERT flag set results in errors.
This happens because some the macros are not defined in the else branch of #ifndef NO_ASSERT.


xTVaser:
Thanks but im curious why you are compiling with `NO_ASSERT` set?  Assertions are in places where them passing is usually critical.  What problem are you trying to bypass?

ManDude:
Shouldn't these macros be defined as `((void)0)` as well? Avoids syntax changes.

clagrisenti:
> Thanks but im curious why you are compiling with `NO_ASSERT` set? Assertions are in places where them passing is usually critical. What problem are you trying to bypass?

I haven't encountered any problem that made me disable asserts.
While learning the codebase I tried compiling the project with the flag set to try I it out.

clagrisenti:
> Shouldn't these macros be defined as `((void)0)` as well? Avoids syntax changes.

You are right.
My mistake. I updated the definitions to match the already present ones.

xTVaser:
kk was just wondering if there was something we should be fixing.

PR #3976, opened 2025-07-05
[build actor] Fix displaying actor materials

gratefulforest:
Known issue confirmed by Hat Kid shortly after the release of the build actor tool as something that came up during testing.

There is a problem which is you can't go over 255 vertices or faces in a single collision primitive. The solution then is to split a single object into two different primitives.

To do this, you need to separate an object in Blender into two separate objects. However, this causes issues because you can't share the same material across two different Blender objects. If you do, any subsequent uses of the same material show up invisible because they get skipped during the build process as they were presumed to already have been built.

This simple fix adds an additional check to check materials AND objects when building materials.

This is a pretty important fix, as there's no way to get around the 255 hex limit, so at that point you have no choice but to separate objects into two separate Blender objects. At this point, it doesn't make sense to not be able to share the same material across two different objects.

Before this PR:

<video controls src="https://github.com/user-attachments/assets/fabcd951-d855-4d52-8ed0-0aeea0b75414"></video>

After this PR:

<video controls src="https://github.com/user-attachments/assets/3e3b7d53-e4ed-43bb-aef1-08cc147d44c0"></video>

### DIFF
--- a/goalc/build_actor/common/MercExtract.cpp
+++ b/goalc/build_actor/common/MercExtract.cpp
@@ -14,7 +14,7 @@ void extract(const std::string& name,
              u32 tex_offset) {
   ASSERT(out.new_vertices.empty());
 
-  std::map<int, tfrag3::MercDraw> draw_by_material;
+  std::map<std::pair<int, int>, tfrag3::MercDraw> draw_by_material;
   int mesh_count = 0;
   int prim_count = 0;
   int joints = 3;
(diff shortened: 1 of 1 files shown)


gratefulforest:
Tested, works, ready. Thank you

gratefulforest:
The reason this failed is because between the time this PR was created and now, a PR was added that included a ;; og:preserve-this comment, so it thought I removed one. Even though the file it failed on is not present anywhere in this PR's diff comparison.

PR #3980, opened 2025-07-13
Replace PCRTC framebuffer blit with screen quad draw

ManDude:
Taking the suggestion from @Calinou (https://github.com/open-goal/jak-project/pull/3943#issuecomment-3017359144), this replaces the resolve/render framebuffer -> window framebuffer blit with an actual drawn tri-strip which covers the entire viewport, which the PCRTC blackout already does.

It appears we have no guarantee what state the internal window framebuffer will be in, so drawing an actual primitive and letting the fragment shader do all the work seems to be the more compatible/functional solution here.

Thanks for the suggestion!


water111 (approved):
Tested on Linux, no issues! This seems better than the blit.

xTVaser:
Looks fine on windows on Jak 1/2 on my end as well.

PR #3986, opened 2025-07-26
New shadow renderer

water111:
This PR ports the shadow renderer from MIPS2C to a new C++ implementation.

This solves the issue where the shadows get weird when the camera is inside the camera volume. 

At this point, I'm not sure if the performance is better or not. The C++ is faster than the GOAL implementation by a lot, but there are more separate index buffer uploads to the GPU, giving more work to the graphics driver.

I'm not really sure that it matters since the meshes for shadow are so simple.


BryceBarbara:
@water111 Can you fix the failing CI check?

xTVaser:
the failing CI check doesn't need to be fixed, it's just a linter.  But there are conflicts / likely a reason (lack of testing) for why this isn't merged yet.

PR #3988, opened 2025-07-27
game: Support korean in Jak 2 and Jak 3

### LINKED ISSUE #3075: korean is not supported
Known since forever, but making an issue now. Not supported in the compiler or decompiler, Korean text cannot be modified or touched in any way right now.

xTVaser:
Resolves #3075 

TODO before merge:
- [x] Properly draw non-korean strings while in korean mode (language selection)
- [x] Check jak 3
- [x] Translation scaffolding (allow korean characters, add to Crowdin, fix japanese locale, etc)
- [x] Check translation of text lines
- [x] Check translation of subtitle lines
- [x] Cleanup PR / some performance optimization (it's take a bit too long to build the text and it shouldn't since the information is in a giant lookup table)
- [x] Wait until release is cut

I confirmed the font textures are identical between Jak 2 and Jak 3, so thank god for that.

Some examples of converting the korean encoding to utf-8.  These show off all scenarios, pure korean / korean with ascii and japanese / korean with replacements (flags):
Screenshot 2025-07-26 191511
Screenshot 2025-07-26 191529

And it working in game.  `Input Options` is a custom not-yet-translated string.  It now shows up properly instead of a disgusting block of glyphs, and all the original strings are hopefully the same semantically!:
Screenshot 2025-07-26 202838

Quite the challenge.  The crux of the problem is -- Naughty Dog came up with their own encoding for representing korean syllable blocks, and that source information is lost so it has to be reverse engineered.  Instead of trying to figure out their encoding from the text -- I went at it from the angle of just "how do i draw every single korean character using their glyph set".

One might think this is way too time consuming but it's important to remember:
- Korean letters are designed to be composable from a relatively small number of glyphs (more on this later)
- Someone at naughty dog did basically this exact process
- There is no other way!  While there are loose patterns, there isn't an overarching rhyme or reason, they just picked the right glyph for the writing context (more on this later).  And there are even situations where there IS NO good looking glyph, or the one ND chose looks awful and unreadable (we could technically fix this by adjusting the positioning of the glyphs but....no more)!

Information on their encoding that gets passed to `convert-korean-text`:
- It's a raw stream of bytes
- It can contain normal font letters
- Every syllable block begins with: `0x04 <num_glyphs> <...the glyph bytes...>`
  - DO NOT confuse `num_glyphs` with num jamo, because some glyphs can have multiple jamo!
- Every section of normal text starts with `0x03`.  For example a space would be `0x03 0x20`
- There are a very select few number of jamo glyphs on a secondary texture page, these glyph bytes are preceeded with a `0x05`.  These jamo are a variant of some of the final vowels, moving them as low down as possible.

Crash course on korean writing:
- Nice resource as this is basically what we are doing - https://glyphsapp.com/learn/creating-a-hangeul-font
- Korean syllable blocks have either 2 or 3 jamo.  Jamo are basically letters and are the individual pieces that make up the syllable blocks.
- The jamo are split up into "initial", "medial" and "final" categories.  Within the "medial" category there are obvious visual variants:
  - Horizontal
  - Vertical
  - Combination (horizontal + a vertical)
- These jamo are laid out in 6 main pre-defined "orientations":
  - initial + vertical medial
  - initial + horizontal medial
  - initial + combination
  - initial + vertical medial + final
  - initial + horizontal medial + final
  - initial + combination + final
- Sometimes, for stylistic reasons, jamo will be written in different ways (ie. if there is nothing below a vertical vowel will be extended).
  - Annoying, and ND's glyph set supports this stylistic choice!
- There are some combination of jamo that are never used, and some that are only used for a single word in the entire language!

With all that in mind, my basic process was:
- Scan the game's entire corpus of korean text, that includes subtitles.  It's very easy to look at the font texture's glyphs and assign them to their respective jamo
- This let me construct a mapping and see which glyphs were used under which context
- I then shoved this information into a 2-D matrix in excel, and created an in-game tool to check every single jamo permutation to fill in the gaps / change them if naughty dogs was bad.  Most of the time, ND's encoding was fine.
  - https://docs.google.com/spreadsheets/d/e/2PACX-1vTtyMeb5-mL5rXseS9YllVj32BGCISOGZFic6nkRV5Er5aLZ9CLq1Hj_rTY7pRCn-wrQDH1rvTqUHwB/pubhtml?gid=886895534&single=true anything in red is an addition / modification on my part.
  - This was the most lengthy part but not as long as you may think, you can do a lot of pruning.  For example if you are checking a 3-jamo variant (the ones with the most permutations) and you've verified that the medial jamo is as far up vertically as it can be, and you are using the lowest final jamo that are available -- there is nothing to check or improve -- for better or worse!  So those end up being the permutations between the initial and medial instead of a three-way permutation nightmare.
  - Also, while it is a 2d matrix, there's a lot of pruning even within that.  For example, for the first 3 orientations, you dont have to care about final vowels at all.
- At the end, I'm left with a lookup table that I can use the encode the best looking korean syllable blocks possible given the context of the jamo combination.


common/goos/Reader.cpp
```diff
@@ -158,37 +158,50 @@ Reader::Reader() {
...
+  // Removed all the logic behind checking the source text, here's my rationale for why:
+  // - we were just assuming everything was a ascii char, 0-255
+  // - then we started allowing multi-byte utf-8 (ie. japanese).
+  //   - the way this was done was by just allowing their individual char bytes, but this is
+  //   meaningless -- it could be allowing ASCII characters like 0x0 or 0x1 because the individual
+  //   bytes are meaningless on their own
+  //
+  // So at this point it's questionable if we should even have this kind of validation if we are
+  // going to break it ourselves to make it still work.
+  return true;
```
xTVaser, line 203: potentially controversial change

water111: I'm also okay with this, but I'd prefer just deleting instead of commenting out and leaving notes. We'll always have it in the git history if we ever want it again :slightly_smiling_face: 

I think this made sense when (at some point) we imagined all files read by a `Reader` were ASCII.

PR #3996, opened 2025-08-13
[jak1] Don't update cell/orb/fly autosplit counters during blackout

dallmeyer:
Similar issue to the orbs-per-level counts, while loading a save (in blackout) these values are reset to 0 briefly. This can cause extra autosplits (see https://github.com/open-goal/speedrunning/pull/25)

### DIFF
--- a/goal_src/jak1/pc/features/autosplit.gc
+++ b/goal_src/jak1/pc/features/autosplit.gc
@@ -17,14 +17,14 @@
   `(set! (-> *autosplit-info-jak1* ,field-name) (if (task-complete? *game-info* (game-task ,task-name)) 1 0)))
 
 (defun update-autosplit-info-jak1 ()
-  ;; general statistics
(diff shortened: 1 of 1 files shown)


Zedb0T:
Is there anytime we would want to update the autosplit struct during a blackout? If it was a legitimate reset/whatever, then surely the 0's would persist after the blackout.

`(>= (-> *game-info* blackout-time) (current-time)`
I wonder if just removing the blackout check here, and adding it where we call `update-autosplit-info-jak1` would just prevent this ever happening again and being more future proof.

Either way I like the change, just trying to think if we can future proof it even more. I guess at this point the current implementation has been tested pretty thoroughly out in the field and if its not broken don't go out of your way to fix/improve it bc you could introduce new bugs.

dallmeyer:
> Is there anytime we would want to update the autosplit struct during a blackout?

I think the task ones we should keep as-is. They might get changed during blackout before/after a cutscene

Zedb0T:
👍

Issue #4005, opened 2025-08-19
[jak1] LPC `qbert-plat-master` crash


dallmeyer:
In LPC, players will sometimes crash in the central whirlpool room. It's somewhat common in noLTS speedruns after doing puzzle skip.

From my investigation, this happens only after you've completed the puzzle room, due to one of the `send-event`s in this bit of code:
https://github.com/open-goal/jak-project/blob/98c1aa4f8f66d6f3dc03c43d6b74c59c54bf7446/goal_src/jak1/levels/sunken/qbert-plat.gc#L425-L440

In some scenarios, the door process `gp-0` and/or the platform process `s5-0` will not be valid processes (maybe because of the `(suspend)` and some process relocation?), and so we crash when trying to read their `event-hook` here:
https://github.com/open-goal/jak-project/blob/98c1aa4f8f66d6f3dc03c43d6b74c59c54bf7446/goal_src/jak1/kernel/gstate.gc#L399-L402

PR #4006, opened 2025-08-19
Check that we have valid process before invoking event-hook (fix jak1 qbert crash)

dallmeyer:
More info in #4005, which this fixes

### DIFF
--- a/goal_src/jak1/kernel/gstate.gc
+++ b/goal_src/jak1/kernel/gstate.gc
@@ -401,7 +401,7 @@ It type checks the arguments for the entry function.
   (with-pp
     ;; og:preserve-this check that we have a _valid_ process. originally this just checked
     ;; that type != process-tree, but sometimes we might get stale process ref (qbert-plat-master)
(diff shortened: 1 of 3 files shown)


goal_src/jak1/kernel/gstate.gc
```diff
@@ -399,7 +399,11 @@ It type checks the arguments for the entry function.
 (defun send-event-function ((proc process-tree) (msg event-message-block))
   "Function to send an event to a process. Please use the send-event macros when possible"
   (with-pp
-    (when (and proc (!= (-> proc type) process-tree) (-> (the process proc) event-hook))
+    ;; og:preserve-this check that we have a _valid_ process. originally this just checked
+    ;; that type != process-tree, but sometimes we might get stale process ref (qbert-plat-master)
+    (when (and proc (valid? proc process #f #f 0)
```
ManDude, line 404: This is fine, but you can remove the lone `proc` here if you want, as the `valid?` call with these args will return false if the input is `#f`

PR #4020, opened 2025-09-10
Bot/update controller db

xexarlc:


xTVaser:
This file is automatically updated from the upstream repo here https://github.com/mdqinc/SDL_GameControllerDB

You should open your PR there not here, otherwise next week your changes will just be removed/replaced.

PR #4041, opened 2025-10-07
Add support for brightness and contrast modifiers

ManDude:


ManDude (commented):
lint*

wojtulace:
Thanks for these but sadly, they do not save the values in my game - always reset to 50.

ManDude:
They are saved to your savefile like on PS2. Are you certain it's not saving?

PR #4063, opened 2025-10-31
jak1: subtitle fixes

gratefulforest:
Listened with RIG 500 headphones:

The farmer:

I really could not hear that "spitting", percussive sound of the "p" for "pack". I think it's unlikely the line is "pack" or "packed into the pen".
The strong, healthy bass of "b" is more likely. Furthermore, he points visually over his shoulder in the same moment, physically referring to his own "back" as if referencing the past or a previous state.

The geologist:

"Lighting" to "lightning" would match consistency to the other references to lightning moles. Listening to the sound, you can hear the distinctly "nasally" sound of the "n" connecting the "light" and "ing".

### DIFF
--- a/goal_src/jak1/engine/game/game-info.gc
+++ b/goal_src/jak1/engine/game/game-info.gc
@@ -418,7 +418,7 @@
         ;; subtract lives.
         (if (= (-> this health) 0.0)
           (adjust (-> (the-as target (-> this process)) game) 'life (- (-> *GAME-bank* life-single-inc)) source-handle))))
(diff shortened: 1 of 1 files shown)


MuchWhittering:
Yeah it's definitely "back". "Pack" doesn't even really work grammatically.

xTVaser:
ok to test

ManDude:
some of these are definitely typos. It is indeed meant to be "back" and "lightning" (they are called lightning moles).

PR #4067, opened 2025-11-01
jak1: fix high fps ripples + frozen ripples

gratefulforest:
Have done some custom ripples so thought I could fix this. At first I scaled ripple speeds before realizing that wouldn't work if you change FPS mid-game.

Instead I found the delta increment is there as a variable, and just scaled that increment directly. I followed the same convention used everywhere; just multiply by `DISPLAY_FPS_RATIO`.

That was the easy bit, the hard part was a second, unrelated issue, whether you scale the delta or not, which is on level entry at high fps, it shows no water movement at all.

I saw the delta was always being created at 0.0 on high FPS, and saw the delta was computed nowhere in GOAL, only mips2c. I found inspiration from [this patch](https://github.com/open-goal/jak-project/blob/5ac5b6aae8f8058e76a27cd14d37db6c0b4d5821/goal_src/jak1/engine/common-obs/ropebridge.gc#L605-L611), where the time ratio is changed to 60 FPS temporarily. However, fixing the time ratio still created 0.0 deltas, and no matter what I did, it always created 0.0 deltas.

I was about to give up, thinking there's no way to scale the mips2c deltas. However, I found out why it was giving 0.0 deltas, which is on high FPS, the video-mode is neither `pal` or `ntsc`, but `custom`, and the misps2c code uses settings for `pal` or `ntsc`.

To include support for high FPS, I temporarily change it to `ntsc`, and only apply this at the wave table creation when `custom`.


Zedb0T:
Seems like this would be a usecase where `protect` could be utilized.

https://github.com/open-goal/jak-project/blob/5ac5b6aae8f8058e76a27cd14d37db6c0b4d5821/goal_src/goal-lib.gc#L439-L458

https://github.com/open-goal/jak-project/blob/5ac5b6aae8f8058e76a27cd14d37db6c0b4d5821/goal_src/jak1/engine/ui/text.gc#L187-L192

gratefulforest:
Oh, I like that. Less lines too, thanks Zed!

water111:
ok to test

PR #4068, opened 2025-11-03
jak3: enable rigid-body during light jak freeze

gratefulforest:
It was not just gun-red-2-shockwaves, or even all guns, but all impacts being blocked during light jak freeze.
It was also not just all hvehicles, but all rigid bodies.
I noticed whilst impacts did nothing, they were definitely accumulating, and went flying at freeze end.
Since it accumulated and applied it afterwards, this made me suspect `rigid-body-queue` as the point of failure.
After some brute-force hacks such as scaling everything by `(-> *setting-control* user-current slow-time)`, I found the exact point of failure, which was this [threshold](https://github.com/open-goal/jak-project/blob/5ac5b6aae8f8058e76a27cd14d37db6c0b4d5821/goal_src/jak3/engine/physics/rigid-body-queue.gc#L76), and making it 0.0001 re-enabled rigid body impacts.
This threshold is whether there's enough time-remaining for one step of rigid body movement.
The problem is the threshold isn't scaled for slow-time, whilst the time-remaining is scaled for slow-time, that is where the mismatch occurs. By printing it you can watch the time-remaining ramp from 0.0166 down to 0.0008333 as light jak freeze gets activated, even though the threshold stays at 0.001. This means the check (enough time-remaining for another rigid body step) is always false when light jak freeze is active.
This mismatch is ultimately caused by the fact (seconds-per-frame) itself is scaled for light jak freeze, used [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/engine/physics/rigid-body-queue.gc#L49).
This minimalist fix simply scales the threshold by slow time, so that it moves with it, this same scalar, multiply by `(- 1.0 (-> *setting-control* user-current slow-time)` is used elsewhere throughout the codebase.


ManDude:
If this is the original, unmodified code, then it should be working without any adjustments. The cause likely lies elsewhere.

gratefulforest:
Here's the full flow:

When a process is created, its clock is set to its active tree's clock [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/kernel/gkernel.gc#L2062).

An entity's active tree is `*entity pool*`, entity-pool's clock is set to `*display* entity-clock` [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/engine/gfx/hw/display.gc#L82).

`*display* entity-clock` is scaled by light jak freeze [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/engine/target/target-lightjak.gc#L2445).

`(seconds-per-frame)` isn't a global, but a macro pointing to each entity's clock [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/kernel/gkernel-h.gc#L674).

Rigid body uses `(seconds-per-frame)` (scaled by light jak freeze) to check rigid body step [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/engine/physics/rigid-body-queue.gc#L49).

Rigid body compares against an unscaled threshold (expecting real time) [here](https://github.com/open-goal/jak-project/blob/0fa93ce7b83d1621f3fd040625bfa865eadaeef0/goal_src/jak3/engine/physics/rigid-body-queue.gc#L76).

You can fix it by making the entity clock not scale by light jak freeze, but of course, things run at full speed. There is nowhere where it diverges into two things. As an alternative fix, you could ask rigid body to use an unscaled timer, but I thought scaling the threshold was more explicit.

```lisp
(let ((f0-0 (-> *display* real-clock seconds-per-frame)))

water111:
thanks for finding this! I ended up going with a slightly different fix in https://github.com/open-goal/jak-project/pull/4095 that should also take into account changes in `time-remaining` due to different frame rates.

gratefulforest:
Thank you! Happy to see it solved!

PR #4073, opened 2025-11-08
jak1: fix opening logo on non-PS2 aspects

gratefulforest:
There are two issues here:

1 is the "punch in" where the logo punches inwards towards the camera on first appearance, this effect reveals the edges of village1 significantly and happens for all users even on regular 16:9.

2 is on aspect ratios wider than 16:9, you see even larger areas of village1 on either side.

The blackout is an artgroup that breaks apart and is set to the shape of the Jak 1 logo. It has a baked in animation and is a set size, so it cannot simply be scaled or adjusted to be wider.

Seeing the sides of village1 before seeing the logo is not the immersive experience this opening was going for, where the game "bursts" into the screen.

My first thought to fix was to simply call `set-letterbox-frames` until the right moment. This worked, however, it didn't produce an ideal result simply because the letterbox also went on top of the light blue "logo volumes" starburst.

Having experimented with the range of edits possible here, I decided the best approach is actually to simply delay displaying village1 ever so slightly. This avoids editing the very particular blackout art group, with set sizes of chunks that break apart. This also avoids covering the blue volumes with a side letterbox/covering.

This scales with any aspect, supports both different language logo blackout shapes, and fixes both wide aspect issues as well as the regular 16:9 punch in issue. This is an important edit as it's the first thing people see.

<video controls src="https://github.com/user-attachments/assets/9e5bfd39-4189-4c7c-bf76-e7ea0df0464f"></video>


ManDude:
this is probably fine. I think you should explicitly set the display to `#f` after starting the load.

gratefulforest:
Thanks for checking it out, it's always interesting to learn what the "right" way to do things is, I've added this in.

goal_src/jak1/levels/title/title-obs.gc
```diff
@@ -435,33 +435,62 @@
...
-                 (115 kill "fishermans-boat-2")
-                 (115 kill "evilplant-2")
-                 (115 kill "revcycle-8")
-                 (115 kill "revcycleprop-2")
-                 (115 kill "water-vol-8")
-                 (115 kill "water-vol-9")
-                 (115 kill "water-vol-10")
-                 (115 kill "reflector-middle-2"))))
+           (if (and PC_PORT (not (-> *pc-settings* use-vis?)))
+             ;; added for pc port, custom spool with vil1 display delayed for high aspect
```
Hat-Kid:
```suggestion
             ;; og:preserve-this added for pc port, custom spool with vil1 display delayed for high aspect
```

gratefulforest:
Thanks for the review HatKid, appreciate it! I was also curious if I need the PC PORT check or if `(-> *pc-settings* use-vis?)` is sufficient on it's own, or should I leave it as is?

Hat-Kid:
> Thanks for the review HatKid, appreciate it! I was also curious if I need the PC PORT check or if `(-> *pc-settings* use-vis?)` is sufficient on it's own, or should I leave it as is?

You can keep it as is.

water111:
ok to test

PR #4075, opened 2025-11-08
jak2: fix hud alignment

gratefulforest:
There were a few things going on here.

Jak 2's menu wheel is zoomed in at 16:9 but zoomed out at 4:3. However, the custom aspects use the scaling from the 4:3 mode, this makes sense as you'd want the zoomed out wheel.

However I noticed HUD elements scaled correctly on PS2 16:9 but not PC 16:9, which was a hint.

The hardcoded values for every individual HUD element for 16:9 and 4:3 are in pixels, normalized across 480 being the stretch of the screen.

To fix the HUD alignment I scaled those offsets, scaling them inside `set-hud-piece-position` meant both textures and merc models got scaled at the same time.

Note that you can get (very tiny) snapping into different integer values as you scale to different aspects. This is simply the result of scaling integer offsets which is what the game uses.


ManDude:
make sure you test this on all the missions that have custom HUD elements

gratefulforest:
Thanks for checking it out, good call!

Through this I found one HUD element appearing incorrectly, which was the floating jetboard trick points. This isn't something introduced by this PR, but something that got left out.

This is because it isn't a texture or merc model, but live text drawn [here](https://github.com/open-goal/jak-project/blob/1f8362d991f5c39057d077c13489df1af6d887e1/goal_src/jak2/engine/game/game-info.gc#L893).

Through this I found an original game bug, which is they scale it by `(-> *video-params* relative-x-scale)` in an incorrect way, which produces mismatched results. You can see this on a real PS2, where setting it to 4:3 shows the points slightly to the left of Jak, and setting it to 16:9 shows the points quite a bit to the right of Jak.

At PC 16:9, the original mistake gets compounded, appearing all the way on the right of screen. In order to preserve the original game bug, I guard on `(-> *pc-settings* use-vis?)`, so 16:9 and 4:3 act the same, but custom aspect ratios don't scale by this factor. This makes custom aspects act like the 4:3 mode (slightly left of Jak).

water111:
ok to test

PR #4079, opened 2025-11-12
[jak3] Implement `city-peace` and `desert-peace` cheats

dallmeyer:
https://www.youtube.com/watch?v=VrEfiMFvUKU

### DIFF
--- a/goal_src/jak3/levels/city/common/kg-squad-control.gc
+++ b/goal_src/jak3/levels/city/common/kg-squad-control.gc
@@ -180,18 +180,6 @@
   (dotimes (v1-3 3)
     (dotimes (a1-3 4)
       (let ((a2-12 (the-as kg-squad-settings (+ (+ (* a1-3 16) 860 (* v1-3 64)) (the-as int this)))))
-        (+! (-> this
-                traffic
-                object-type-info-array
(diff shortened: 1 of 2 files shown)


goal_src/jak3/levels/city/common/ff-squad-control.gc
```diff
@@ -672,6 +672,23 @@
...
                   )
+            ;; og:preserve-this peaceful city cheat
+            (#when PC_PORT
+              (when (and (pc-cheats? (-> *pc-settings* cheats) city-peace) 
+                         (not (task-node-open? (game-task-node city-port-fight-resolution)))
+                         (not (task-node-open? (game-task-node city-port-assault-resolution)))
+                         (not (task-node-open? (game-task-node city-protect-hq-resolution)))
+                         )
+                (dotimes (guard-info-type 3) ;; normal/war/mission
+                  (when (= (-> a1-13 data guard-info-type object-type) (traffic-type guard-a))
```
dallmeyer: `wlander-male` can be part of ff-squad-control supposedly, so I'm explicitly checking for `guard-a` here. There is a `guard-b` too but as far as I could tell it's not used?

PR #4083, opened 2025-11-23
[jak3] Support dpad pacman controls

dallmeyer:
Not sure if we really want this change in vanilla openGOAL, but I had it working on a branch so figured I'd PR it.


xTVaser:
Could be a setting, i dont see the harm

ManDude:
straight up don't remember how this minigame even plays but if it's what I'm thinking then yea sure

PR #4085, opened 2025-11-27
[jak3] Some autosplit changes

dallmeyer:
Adds some additional autosplit points that I think will be useful. 

Also tweaks the `errol-dead?` condition to update immediately on the final hit (there's a sec or so of animation before it returns to `initial-state`, where it was being updated previously)

Corresponding autosplitter updates: https://github.com/open-goal/speedrunning/pull/27


dallmeyer:
> Flagged goal_src code has been deleted, either you made a mistake or you better know what you're doing!
>  - goal_src/jak3/levels/desert/boss/terraformer-head.gc:2608 - ;; og:preserve-this

(yeah, this is the `errol-dead?` condition I moved)

PR #4093, opened 2025-12-08
[jak3] Fix audio language setting

water111:
Rename `Language` to `Audio Language` in the menu, and connect it to `audio-language` instead of `language`.

After this change, we have:
image

where 

- Subtitle Language controls `subtitle-language` on both PC settings and the normal setting control (unchanged)
- Audio Language controls `audio-language` on setting-control
- Text Language controls `text-language` on PC settings and `language` on setting control. The game looks at `(-> *setting-control* user-default language)` to determine text scaling sometimes, so I think it's good for this to set both.

### DIFF
--- a/goal_src/jak3/pc/progress/progress-static-pc.gc
+++ b/goal_src/jak3/pc/progress/progress-static-pc.gc
@@ -484,7 +484,7 @@ This gives us more freedom to write code how we want.
 
 (defmacro game-options-pc-sound-language ()
   `(new 'static 'menu-generic-carousel-option
-      :name (text-id progress-language)
+      :name (text-id progress-audio-language)
       :items (new 'static 'boxed-array :type text-id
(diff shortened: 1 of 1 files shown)


cmalvi:
Hi, i've tested it now pulling the branch and redo all from scratch, now the game crash after selecting a language different from english (italian in my case) in the first instance of an audio voice, plus the game text still reset to english at every boot

NOTE: the crash does not occur with cutscenes only with in game audio

`[54:06:548] [error] Invalid VAG data is 0x643e0627`
```
[54:06:549] [die] Assertion failed: 'not reached' Source: /home/deck/jak-project/game/overlord/jak3/spustreams.cpp:210
        Function: jak3::EIsoStatus jak3::ProcessVAGData(ISO_Hdr*)
```

Exit with error 134

PR #4101, opened 2025-12-14
jak1: fix aspect settings selection

gratefulforest:
I was changing aspect ratios when I noticed something was off, getting different results for the same setting.
On screen positions are determined by both the aspect ratio you select as well as the last time you selected a PS2 aspect ratio.
You get different result by choosing 4x3 (PS2), then 16:9, compared to 16:9 (PS2), then 16:9. The same is true for all aspects and Fit to Screen.

By contrast, Jak 2 always uses the 'aspect4x3 scale factor for all custom aspects, making the menu option you select consistent.

There is an intention to use 4x3 for custom aspects [here](https://github.com/open-goal/jak-project/blob/abf27960f0033436b066c712c787160d2c9a5346/goal_src/jak1/pc/progress-pc.gc#L2618); when `fit-to-screen` then `(set-aspect-ratio 'aspect4x3)`. However it needs the extra lines to take effect.

This is helpful as the only cause of vertical misalignment on memory cards is using PS2 16x9 on custom aspects:


xTVaser:
ok to test

PR #4105, opened 2025-12-20
jak1: fix high fps menu sliders

gratefulforest:
Menu sliders are very slippery at high FPS and make it touchy and difficult to select precisely:

https://github.com/user-attachments/assets/3c9b9ae5-0c1f-4a3a-bb18-1bfae5396b6d


xTVaser:
ok to test

PR #4106, opened 2025-12-28
Prevent array OOB access for traffic-height-map

dallmeyer:
If the `arg0` vector has extreme values, `a2-0` and `a1-1` can end up being negative, which leads to out-of-bounds access when using these values to read from the `xz-height-map`'s `data` array, and crashes the game (you can pretty easily reproduce this in jak 3 by overcharging the "superfasthoverbikeglitch")

The `min` here already makes sure these values don't go too high, but doesn't cover the case when they go negative.

As I understand it, this `data` in the `*traffic-height-map*` functions as a 2d array, bucketing x/z coordinates into heights for the vehicle.

### DIFF
--- a/goal_src/jak2/levels/city/common/height-map.gc
+++ b/goal_src/jak2/levels/city/common/height-map.gc
@@ -103,7 +103,8 @@
          (a1-7
            (the-as
              (pointer int8)
-             (+ (+ (min a2-0 (+ v1-0 -2)) (* (min a1-1 (+ a3-0 -2)) v1-0) 0) (the-as int (the-as pointer (-> this data))))
+             ;; og:preserve-this prevent array out-of-bounds access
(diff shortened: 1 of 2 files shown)


dallmeyer:
> if `v1-0` is less than 2

I don't think that is possible, these are effectively static values since this is only ever used for the `*traffic-height-map*`
```lisp
         (v1-0 (-> this x-dim))
         (a3-0 (-> this z-dim))
```

```lisp
(deftype xz-height-map (structure)
  (...
   (dim            int16  2)
   (x-dim          int16  :overlay-at (-> dim 0))
   (z-dim          int16  :overlay-at (-> dim 1))
```

```lisp
(define *traffic-height-map* (new 'static 'xz-height-map
                               ...
                               :dim (new 'static 'array int16 2 80 #x70)
```

PR #4116, opened 2026-01-04
jak1: fix memcard screen on custom aspects

gratefulforest:
The particles for the grey memcard slots already had functions associated with them, so scaling the grey boxes was simply a matter of making use of those functions.

The memcard boxes expand and shrink with the aspect to resemble the PS2 behavior, where at 4:3 you get a short grey box and at 16:9 you get a long grey box.

Although I added scaling to the grey boxes, I wasn't satisfied with the way elements would "slip" as they would scale differently to each other. Mainly, the collectable counts and their totals slide more and more off from each other.

This is due to the game's use of integer offsets to decided UI placements; they were the right numbers for 4:3, but once you start to scale, they can slide more and more off. You can see this on a real PS2, where at 4:3, the collectibles are centered above the totals, and at 16:9, the collectibles are more right-aligned above the totals.

To fix this, I used the same text flags on the totals as the counts, so everything is snapped to each other. 
Guards against `(-> *pc-settings* use-vis?)` ensure PS2 aspects remain untouched, and at PS2 16:9, you still get the original game bug of more right aligned counts.

https://github.com/user-attachments/assets/0fa00a5d-0090-4cd0-b68b-8b8f28816185

### DIFF
--- a/goal_src/jak1/engine/ui/progress/progress-draw.gc
+++ b/goal_src/jak1/engine/ui/progress/progress-draw.gc
@@ -474,11 +474,11 @@
                   (s0-5 *temp-string* arg0 #f s4-1 22))
                 (let ((v1-116 arg0)) (set! (-> v1-116 scale) 0.5))
                 (+! (-> arg0 origin y) 9.0)
-                ;; og:preserve-this added 'middle' flag when custom aspect to snap totals with counts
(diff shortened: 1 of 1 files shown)


xTVaser:
ok to test

PR #3902, opened 2025-04-26
[jak1] Speedrun mode misc fixes

dallmeyer:
Fixed up a few things for speedrun mode in Jak 1:
- Changed autosplitter to start run in line with ruleset (first frame after blackout where jak is touching ground). Also tweaked the verification text in bottom left to show "Run Started: Yes/No" to go with this
- Changed speedrun menu button combo to match Jak 2/3 (only need to hold L1+R1 and press Start/Select, no longer need to hold X)
- Treat speedrun menu via Select the same as Start (no more potential for funny pause buffer where actors freeze)

warp gate example:
vlcsnap-2025-04-26-12h03m37s116
vlcsnap-2025-04-26-12h03m42s642

already grounded example (run starts once partially out of blackout):
vlcsnap-2025-04-26-12h04m02s085

PR #4126, opened 2026-02-21
[jak2/jak3] Wait to reset autosplitter until after blackout

dallmeyer:
Based on [similar changes we made awhile back for jak1](https://github.com/open-goal/jak-project/pull/3902) - this way runners don't need to figure out the -0.3s offset or whatever, and it won't vary based on where they reset from.


xTVaser:
add to whitelist

PR #4133, opened 2026-03-13
[jak2/jak3] Unbreak split boxes debug option

dallmeyer:
The `Split Boxes` option in jak2/3 currently crashes as soon as you select it because of the `*debug-region-hide-empty*` code added in openGOAL. The split boxes aren't true regions, but they reuse the `debug-draw-region` functions, and crash trying to access `(-> this region on-enter)`.

image
yay split boxes


ManDude:
those don't look like the right colors lol but I guess it's fine

PR #4145, opened 2026-03-31
fix typos in method names

trippjoe:
fix typos in the following method names: 
1. `closest-point-on-mesh`
2. `can-play-stance-ambient`


Hat-Kid:
`can-play-stance-amibent` is a function symbol, not a method, which means it's not a name that came from us, so I would advise against renaming it. It makes the offline tests fail.

PR #4147, opened 2026-04-01
jak3: Finish font DB implementation (support Cyrillic)

### LINKED ISSUE #4001: Finalize Jak 3's font DB
Right now Jak 3's font DB is just a copy of Jak 2's.  This means it doesn't properly support cryllic, and there's some new flags/korean strings that are incorrect.

xTVaser:
Also add in all the other language stubs / copy over the custom text to the correct enum ids.

Closes #4001


Aloqas:
For some reason `game_custom_text` files received some kind of title casing on every word. Button glyph variables such as `<PAD_CIRCLE>` were also affected.

PR #4208, opened 2026-04-11
[jak3] Implement trick display text secret

dallmeyer:
ported the implementation from jak 2 into jak3 pretty much exactly. added a new color for L2 tricks, and shrunk the text/width a bit so it wouldn't end up covered by the minimap.

also fixed small bug in jak 2 (and 3) where dismounting jetboard wouldn't reset the trick display, so displayed score wouldn't match the granted score


goal_src/jak2/engine/target/board/target-board.gc
```diff
@@ -777,6 +777,9 @@
   (set! (-> self board flip-count) 0)
   (set! (-> self board unknown-float00) 0.0)
   (set! (-> self board unknown-float01) 0.0)
+  ;; og:preserve-this reset trick display
+  (#when PC_PORT
+    (reset-combo! *board-trick-tracker*))
```
dallmeyer, line 782: this is the fix to reset trick tracker if you dismounted and are now re-mounting jetboard

decompiler/config/jak3/all-types.gc
```diff
@@ -10325,6 +10325,35 @@
   (progress-misc-fix-projectile-focus #x134d)
   (progress-controller-options-pacman-dpad #x134e)
   (progress-controller-options-analog-sensitivity #x134f)
+  (trick-board-none #x1350)
+  (trick-board-board-spin #x1351)
```
xTVaser: hmm, how come you had to move these down in the enum?  It doesn't seem like their spots were taken up in jak 3.

It's not a big deal but, you should have been able to avoid all of this work of moving the strings to new values...i think?

dallmeyer: there were too many new ones so I would've had to shift all the language and controller options as well to make it fit

in jak 2 but not 3: dark-giant
in jak 3 but not 2: kickspin, method-cross, backgrab, airwalk, darkjak-smack, darkjak-tracking, lightjak, lightjak-regen, lightjak-swoop, lightjak-freeze, lightjak-shield

I guess I could've shifted them backwards and left out/broken up the lightjak ones but that felt kind of weird

xTVaser: oh sorry i see you added new ones in addition to the old, kk let's just delete the old entries instead of commenting them out then (in all-types and in text-h).  

Maybe leave behind a comment saying "gap here if you want to use it" but having the entries remain commented out makes it feel like they are reserved / disabled / something.

xTVaser (approved):

PR #4216, opened 2026-04-12
[jak3] Fix some things relating to resetting speedruns

dallmeyer:
addresses:
- starting new game from title screen with speedrun mode on -> shows intro cutscene, doesn't start autosplitter, verification text not shown
  - resetting run during intro cutscene leaves weird desert state across training course (not an issue if we fix^)
- reset in hero mode after completing training course -> verification text not shown
- resetting during tentacle grab leaves the water bubble sound playing forever

### DIFF
--- a/goal_src/jak3/levels/wascity/wasall-obs.gc
+++ b/goal_src/jak3/levels/wascity/wasall-obs.gc
@@ -348,6 +348,11 @@
 (defstate attacking-0 (tentacle)
   :virtual #t
   :event tentacle-attack-handler
+  ;; og:preserve-this make sure tentacle-loop sound isn't orphaned and looping forever
+  :exit (behavior ()
+    (when (and (!= (-> self next-state name) 'attacking-1) (!= (-> self next-state name) 'kill-player))
+      (sound-stop (-> self sound-id-loop)))
+    )
   :trans (behavior ()
     (set! (-> self sound-id-attack)
           (add-process *gui-control* self (gui-channel background) (gui-action queue) "tentacle" -99.0 0)
@@ -452,6 +457,8 @@
     (if *target*
         (send-event *target* 'end-mode 'bot)
         )
+    ;; og:preserve-this make sure tentacle-loop sound isn't orphaned and looping forever
+    (sound-stop (-> self sound-id-loop))
     )
   :code (behavior ()
     (ja-channel-push! 1 0)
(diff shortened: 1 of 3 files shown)


goal_src/jak3/levels/wascity/wasall-obs.gc
```diff
@@ -452,6 +457,8 @@
     (if *target*
         (send-event *target* 'end-mode 'bot)
         )
+    ;; og:preserve-this make sure tentacle-loop sound isn't orphaned and looping forever
+    (sound-stop (-> self sound-id-loop))
```
dallmeyer, line 461: normally the tentacle always goes `attacking-0` -> `attacking-1` -> `kill-player`, so this is maybe overkill

but in theory `attacking-0` has the same event handler as `attacking-1` that sends it to `kill-player`, so if somehow the player died from `'tentacle` while a tentacle was in `attacking-0`, it could go directly to `kill-player` and would leave the sound looping

goal_src/jak3/pc/features/speedruns.gc
```diff
@@ -670,8 +670,10 @@
           (reset! *autosplit-info-jak3*))))
     ;; Update auto-splitter struct
     (update! *autosplit-info-jak3*)
-    ;; see if we should stop drawing the run info (when you finish arena training)
-    (when (and (!= (-> this category) (speedrun-category custom)) (task-complete? *game-info* (game-task arena-training-1)))
+    ;; see if we should stop drawing the run info (when you finish arena training, but also make sure at least 10s have passed)
+    (when (and (!= (-> this category) (speedrun-category custom))
+               (task-complete? *game-info* (game-task arena-training-1))
+               (time-elapsed? (-> this run-started-at) (seconds 10)))
```
dallmeyer, line 676: this fixes the heromode issue which is unique to heromode because we spawn a function to do the intitialize! 
https://github.com/open-goal/jak-project/blob/fedd27ebfd8532ede1842837ef15127463a4a8c6/goal_src/jak3/pc/features/speedruns.gc#L166-L181

instead of just doing it immediately like we do in regular new game
https://github.com/open-goal/jak-project/blob/fedd27ebfd8532ede1842837ef15127463a4a8c6/goal_src/jak3/pc/features/speedruns.gc#L160-L165

xTVaser (approved):
changes look good -- checked the one thing that effects non-speedrunner scoped code (the tentacle sound fix) -- sound bubble still plays as expected

PR #4217, opened 2026-04-12
Make surround sound not the default

dallmeyer:


goal_src/jak3/engine/game/settings.gc
ManDude: Change this to something like 0 so that it doesn't match the default. This is needed to force the game to perform the code to properly update the setting.

ManDude (changes requested):

ManDude (approved):

PR #4236, opened 2026-04-19
game: support splash screens

### LINKED ISSUE #1496: [game] hack the sony splash back in
Display the SCE splash on boot. This isn't done through GOAL in the original game, so we haven't implemented it.

Hat-Kid:
Closes #1496 

This brings back the SCE splash screens. Also adds a runtime flag `-nosplash` to skip it. Right now, it's on by default, but open to changes on that (maybe always disable in debug to speed up start times, etc.).


ManDude:
awesome, thanks!

Is there any reason you didnt hook the new shader up the same way as the other shaders?

xTVaser:
> maybe always disable in debug to speed up start times

yeah i think we want it off in debug mode or it's going to drive us crazy lol

Hat-Kid:
> awesome, thanks!
> 
> Is there any reason you didnt hook the new shader up the same way as the other shaders?

I did try this at first, but I ran into a bit of a snag because of the way the splash is currently hooked up. Most of the setup is in `opengl.cpp` and, from what I can tell, the other shaders are set up via the `OpenGLRenderer`'s `SharedRenderState` which I have no easy access to from there without making it `public`. This isn't a bucket renderer, anyway, so it's a bit of a special case.

ManDude (approved):

Issue #4235, opened 2026-04-19
Defend Port from attack softlock


LAS3RHUN:
### Acknowledgements

- [x] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak3+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [x] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

During mission 'Defend Port from Attack' given by Torn, where you have to defeat waves of enemies in these mini-arenas, after the 3rd wave (where you have 3 civilians), the moment Jak travels to the last area, enemies don't spawn and you can't control Jak. But his idle animation is still working, so do enviromental animations (smoke, gun's laser etc.) and the music in the background too. I reloaded the save file, restarted the mission but to no avail. I even tried quitting and restarting but that didn't work either.

Didn't let me upload the video so here's a link for youtube:
https://youtu.be/rij1xyAPiII

### How To Reproduce

Play the mission I guess. It's like 6 hours into the story so i won't do it all over again, but as I said, it happens each time I get to that point on this  save file.

### Does this problem occur on original hardware or PCSX2?

Yes, it's unique to OpenGOAL

### Expected Behavior

When Jak automatically arrives to the next area, control should be given back to the player and enemies should start appearing. That doesn't happen.
Here's a link of how the mission's supposed to go:
https://www.youtube.com/watch?v=uKNK1QIL6Ck

### Environment Information

CPU: AMD Ryzen 7 5700x
GPU: NVIDIA GeForce RTX 3060 Ti
OS: Windows 11 Pro 25H2
OpenGOAL Version: Launcher v2.9.1 Tooling v0.3.1

### Game Version

PAL

### Have you set the game to something other than `60fps`?

No

LAS3RHUN:
Apparently going back to Tooling version v0.3.0. solved the issue for me. If anyone stumbles into this I recommend doing so.

ManDude:
Can't replicate. Also I do not recommend downgrading.

LAS3RHUN:
Why not? Can it break the save file or what? I'm not really great at things like this.

Zedb0T:
0.3 has a lot of first release issues that were fixed in a update that you are likely to run into, and settings/saves aren't promised to be backward compatible and may reset to default

dude19288-commits:
[...]

PR #4237, opened 2026-04-20
[jak3] Prevent NaN propagation in spydroid attack state

dallmeyer:
Another patch to prevent NaN propagation. Still doesn't address the root cause of the NaNs, but should fix some mission behavior.

In some cases, crimson-guards will [return NaN vectors for their `bullseye-joint`](https://github.com/open-goal/jak-project/blob/1dff57182004f058848b26f0f8bfbb111a96e480/goal_src/jak3/engine/ai/enemy.gc#L178-L183). I suspect the root cause of these NaNs may be related to the guards missing their torsos

image

In the [spydroid's `attack` state](https://github.com/open-goal/jak-project/blob/1dff57182004f058848b26f0f8bfbb111a96e480/goal_src/jak3/levels/common/enemy/spydroid.gc#L743-L772), this NaN value will end up propagating to the spydroid's position.

This in turn causes issues like the infinite looping addressed in #4225. Or another issue in "Defend HQ From Attack" - every spydroid spawned by the task manager can end up with NaN positions. The spydroids will be nowhere to be seen, but the task-manager still considers them alive and won't spawn any other KG enemies.

### DIFF
--- a/goal_src/jak3/levels/common/enemy/spydroid.gc
+++ b/goal_src/jak3/levels/common/enemy/spydroid.gc
@@ -745,6 +745,9 @@
         (when a0-2
           (let ((s5-0 (new 'stack-no-clear 'vector)))
             (vector-copy! s5-0 (get-trans a0-2 3))
+            ;; og:preserve-this prevent NaN propagation - if bullseye-joint (3) returns NaNs, fall back to root trans (0)
+            (when (is-nan? (-> s5-0 x))
+              (vector-copy! s5-0 (get-trans a0-2 0)))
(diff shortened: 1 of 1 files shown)


xTVaser:
Does this effectively fix the aforementioned issues and we can close those out / make a new issue specifically tracking the root cause of the NaNs?  Or are we unsure if this totally solves things. ie. seems like this issue might be related https://github.com/open-goal/jak-project/issues/4235

dallmeyer:
Hmm I hadn't seen the defend port issue, that does sound like it could be a similar situation as defend hq since it involves both spydroids and guards, but I've never looked into that one

PR #4243, opened 2026-04-23
[jak3] Fix Daxter's death quotes scenes not playing

Nick0778:
This PR fixes a bug in `target-death.gc` that was causing Daxter's death quotes scenes to not play in OpenGoal. 

Basically, these scenes weren’t working because `sv-96` was assigned an invalid level, and as a result the `s4-1` check always returned `#f`. So, the `(when s4-1 ...)` block that contains the logic for these scenes was never executed.

png1

## Before the fix:

https://github.com/user-attachments/assets/58e2f412-814b-40ff-ab64-d22054a7232e

## After the fix:

https://github.com/user-attachments/assets/bf6ac0da-1a62-4cb4-b363-20af898dac7f

### DIFF
--- a/decompiler/config/jak3/ntsc_v1/stack_structures.jsonc
+++ b/decompiler/config/jak3/ntsc_v1/stack_structures.jsonc
@@ -195,8 +195,9 @@
   "(code target-hit)": [[16, "vector"]],
   "apply-pos-from-entity": [[208, "collide-query"]],
   "target-death-main": [
-    [16, "event-message-block"],
-    [96, ["array", "level", 1]]
+    [16, "event-message-block"]
+    // this wrong stack cast was the reason Daxter's death quotes scenes didn't work
+    //[96, ["array", "level", 1]]
   ],
   "find-nearest-focusable": [[48, "vector"]],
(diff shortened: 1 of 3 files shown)


ManDude (changes requested):
The actual fix seems to be replacing `(new 'stack-no-clear 'array 'level 1)` with `(& sv-96)`. Appears to be a decompiler bug, or some wrong decompiler settings for this function.

Nick0778:
> The actual fix seems to be replacing `(new 'stack-no-clear 'array 'level 1)` with `(& sv-96)`. Appears to be a decompiler bug, or some wrong decompiler settings for this function.

I found what was causing this decompiler bug. If I comment out this line here in `stack_structures.jsonc` and run the decompiler it will be fixed.

png2

the result is:

png3

ManDude (approved):

xTVaser:
ok to test

ManDude:
REF file needs to be updated

Nick0778:
> REF file needs to be updated
Done.

ManDude:
thank you for tracking this down!

PR #4266, opened 2026-05-04
decomp: resolve first batch of trivial files

xTVaser:


decompiler/config/jakx/jakx_config.jsonc
Hat-Kid: We probably want to see if we can get all of this info dumped soon before we start getting to files that use these now that all objects pass. Not a huge priority for now, though.

goal_src/jakx/engine/collide/pat-h.gc
```diff
@@ -7,3 +7,262 @@
...
+     )
+    (((pat-material stopproj))
+     "stopproj"
+     )
+    (((pat-material mhshroom))
+     "invisible"
+     )
+    (((pat-material mhswamp))
+     "forcefield"
+     )
```
Hat-Kid: Enum mismatch here.

goal_src/jakx/engine/collide/pat-h.gc
```diff
@@ -7,3 +7,262 @@
...
+    (((pat-event melt))
+     "e04"
+     )
+    (((pat-event drag))
+     "e13"
+     )
+    (((pat-event burnup))
+     "e03"
+     )
```
Hat-Kid: Ditto here.

goal_src/jakx/engine/common-obs/prim-h.gc
```diff
@@ -5,5 +5,324 @@
...
+  (:methods
+    (new (symbol type int texture-id string) _type_)
+    (setup-dma-and-tex (_type_ draw-control) none)
+    )
+  )
+
+
+;; ERROR: failed type prop at 44: Unknown symbol: lookup-texture-id-by-name
+;; WARN: Return type mismatch none vs prim-strip.
+(defmethod new prim-strip ((a0-0 symbol) (a1-0 type) (a2-0 int) (a3-0 texture-id) (t0-0 string))
```
Hat-Kid: This method is failing. Looks like it needs `lookup-texture-id-by-name`.

goal_src/jakx/engine/dma/dma-buffer.gc
```diff
@@ -7,3 +7,150 @@
...
+  (shr (+ (&- (-> dma-buf base) (the-as uint (-> dma-buf data))) 15) 4)
+  )
+
+(defun dma-buffer-free ((dma-buf dma-buffer))
+  (shr (+ (&- (-> dma-buf end) (the-as uint (-> dma-buf base))) 15) 4)
+  )
+
+;; ERROR: Failed store: (s.w! (+ t2-0 8) t3-5) at op 23
+;; ERROR: Failed store: (s.w! (+ t2-0 12) t3-8) at op 32
+(defun dma-buffer-add-vu-function ((dma-buf dma-buffer) (vu-func vu-function) (flush-path-3 int))
```
Hat-Kid, line 424: This function is failing, but I believe this was also the case in the other games and we should just be able to copy Jak 3's manually patched version since we don't want to actually transfer VU programs on PC anyway.

goal_src/jakx/engine/entity/entity-h.gc
```diff
@@ -7,3 +7,132 @@
...
+    )
+  )
+
+;; type entity-camera is defined here, but it is unknown to the decompiler
+
+;; type entity-nav-mesh is defined here, but it is unknown to the decompiler
+
+;; type entity-race-mesh is defined here, but it is unknown to the decompiler
+
+;; type entity-actor is defined here, but it is unknown to the decompiler
```
Hat-Kid, line 127: Not sure I've seen this error before, but I assume this is just from these types being defined in both `entity-h` and `entity` and not being uncommented in `all-types`.

goal_src/jakx/engine/gfx/blit-displays-h.gc
```diff
@@ -7,3 +7,59 @@
...
+(defmethod do-blit-displays ((this blit-displays-work))
+  (-> this horizontal-flip-flag)
+  (none)
+  )
+
+;; WARN: Return type mismatch dma-buffer vs none.
+(defmethod blit-menu-mode ((this blit-displays-work) (arg0 dma-buffer))
+  (set! (-> this horizontal-flip-flag) (the-as symbol arg0))
+  (none)
+  )
```
Hat-Kid: Looks like mismatched methods? The methods probably got scrambled around between games, so they're not the same IDs anymore. We should double check whether the old method signature matches for the given ID before copying them.

goal_src/jakx/engine/math/math.gc
Hat-Kid, line 1: This file has tons of manual patches in the other games and won't compile here as is (e.g. `gpr->fpr` in `logf`). We should just copy all the identical functions verbatim from Jak 3.

goal_src/jakx/engine/math/vector.gc
Hat-Kid, line 1: The automatic detection of the inline vector functions also applies inside of the actual functions themselves here which just turns them into recursive function calls. We could either special case this in the decompiler or just manually patch/copy from Jak 3.

goal_src/jakx/engine/math/vector.gc
```diff
@@ -7,3 +7,1677 @@
...
+         )
+    (init-vf0-vector)
+    (.lvf vf1 (&-> arg0 quad))
+    (.sub.vf.xyz vf1 vf0 vf1)
+    (.svf (&-> arg0 quad) vf1)
+    arg0
+    )
+  )
+
+;; ERROR: function was not converted to expressions. Cannot decompile.
```
Hat-Kid: Some function failed here. I think maybe `vector=`?

goal_src/jakx/engine/math/trigonometry.gc
Hat-Kid, line 1: Lots of manual patches/failing functions here as well, can probably just copy Jak 3 again.

goal_src/jakx/engine/gfx/vu1-user-h.gc
```diff
@@ -5,5 +5,40 @@
 ;; name in dgo: vu1-user-h
 ;; dgos: ENGINE, GAME
 
+;; +++bucket-id-16
+(defenum bucket-id-16
+  :type int16
+  :copy-entries bucket-id
+  )
```
Hat-Kid, line 792: There doesn't seem to be a `bucket-id` enum defined (might have missed it, but it should normally be defined here). We should also ideally see if we can give the new buckets proper names before we encounter lots of code that uses them and also double check if the old names still match.

goal_src/jakx/dgos/menu2.gd
```diff
@@ -80,5 +80,5 @@
   "garage-turntable-ag.go"
   "garage-door-ag.go"
   "interface-cam-ag.go"
-  "menu2-MENU2.o"
+;;   "menu2.o" already in common
```
Hat-Kid: I strongly suspect that this is not a duplicate of the `menu2` code file, but rather the `bsp-header` for a `menu2` level, given that it is the last file in the DGO, so it should be named `menu2.go`. You should be able to confirm this by opening up `menu2-MENU2.asm` if you turn on the `disassemble_data` config option.

There are plenty of instances of code files sharing the same name as art groups in DGOs, but I guess we just never had the same case with levels?

goal_src/jakx/engine/engine/connect.gc
```diff
@@ -187,15 +204,17 @@ New for Jak 2: You can use a child class of [[connection]]."
...
+            (set! (-> (the-as (pointer pointer) v1-10)) (&+ (the-as pointer v1-10) (-> arg2 size)))
             (set! v1-10 (&+ (the-as pointer v1-10) (-> arg2 size)))
             (+! a0-6 1)
             )
           )
         (set! (-> (the-as connectable v1-10) prev0)
               (the-as connectable (&- (the-as pointer v1-10) (the-as uint (-> arg2 size))))
               )
-        (set! (-> (the-as (pointer int32) v1-10)) (-> v0-0 dead-list-end))
+        ;; og:preserve-this cast changed to (pointer collectable)
```
Hat-Kid, line 216: I'm guessing this typo was copypasted from me 😛

goal_src/jakx/engine/game/game-info-h.gc
```diff
@@ -7,3 +7,7 @@
 
 ;; DECOMP BEGINS
 
+(deftype actor-id (uint32)
+  ()
+  :flag-assert #x900000004
```
Hat-Kid, line 12: Can get rid of this assert, I don't think it's supposed to show up in the final output.

goal_src/jakx/engine/level/level-h.gc
```diff
@@ -5,6 +5,240 @@
 ;; name in dgo: level-h
 ;; dgos: ENGINE, GAME
 
+;; max amount of levels in level heap
+(defconstant LEVEL_MAX 10)
+;; total amount of levels, including ones outside level heap (default-level)
+(defconstant LEVEL_TOTAL 11)
```
Hat-Kid: These need to be changed. Looks like there's 18 max levels + the default one now. There are also equivalent C++ constants that probably need to be adjusted too.

goal_src/jakx/dgos/menu2.gd
xTVaser: You are correct.  And yeah not sure where the bug lies in the naming, could be my python script, could be something else.

Hat-Kid (approved):

PR #4271, opened 2026-05-10
Include merc normals for glb extraction

ZornTaov:
Namely all I did was include the normals from MercVertex in with the glb exporter.  I also converted parts of MercVertex to use union structs for faster casting.  The union structs should just work on most compilers from what I researched, if not I can revert that and go a different route for that.


ZornTaov:
oh, noticed I left in my `.name = "NORMAL"` comments.  these don't affect anything, the gltf2.0 standard calls them user-defined and optional.  I added them and others that I can add in a later PR just to keep track of what was what in the glb file.

Hat-Kid:
ok to test

PR #4280, opened 2026-05-19
fix mac dialog during gameplay causes frame rate to drop sharply

trippjoe:
.

### DIFF
--- a/game/graphics/opengl_renderer/OpenGLRenderer.cpp
+++ b/game/graphics/opengl_renderer/OpenGLRenderer.cpp
@@ -39,6 +39,14 @@
 #endif
 #include "common/util/string_util.h"
 
+#if defined(__APPLE__) && (defined(__aarch64__) || defined(__arm64__))
+constexpr bool kUseDirectHudPris2 = false;
+constexpr bool kUseDirectSubtitle = false;
+#else
+constexpr bool kUseDirectHudPris2 = true;
+constexpr bool kUseDirectSubtitle = true;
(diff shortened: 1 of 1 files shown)


trippjoe:
upon further testing i seem to have made an error.

xTVaser:
Please remove the Closes reference, because even if this helps -- the game performs terribly via Rosetta even outside of the dialog/subtitle scenarios, as shown by the profiling that was done (just idling in the game, no dialog).

I would be surprised if it's actually the graphics code that is causing the perf. issue...since that runs on the GPU (no rosetta involved)

trippjoe:
Yes sir

game/graphics/opengl_renderer/OpenGLRenderer.cpp
```diff
@@ -39,6 +39,14 @@
 #endif
 #include "common/util/string_util.h"
 
+#if defined(__APPLE__) && (defined(__aarch64__) || defined(__arm64__))
+constexpr bool kUseDirectHudPris2 = false;
+constexpr bool kUseDirectSubtitle = false;
+#else
+constexpr bool kUseDirectHudPris2 = true;
+constexpr bool kUseDirectSubtitle = true;
+#endif
```
xTVaser, line 48: not sure if you wrote this yourself but, the entire project uses snake_case for the most part.

it would also read better if the variables were defined closer to where they are used, right now these are defined >500 lines away from where they are used.

PR #4281, opened 2026-05-19
jak3: Dev commentary subtitles

Aloqas:
Subtitles for ND dev commentary tracks


xTVaser (approved):
You're a machine.  

Are these really all the subtitles for the commentary?  Can't say I've ever played through with the commentary enabled but I guess I expected it to be a little longer than ~1K lines.

Kinda curious how our code will handle this, since you probably want normal (english or whatever language) subtitles for all the spots where commentary isn't covered.  I'll have to check it out. (I've been educated, in the original game those cutscenes were also dead silent lol, we can probably do better but no rush in that case)

Aloqas:
Luckily the Jak and Daxter Wiki had done all the transcribing and I just copy pasted the lines with some corrections here and there, which sped up the process a lot.

I kinda question the decision to include commentary as a selectable language option, when they're only supposed to be played through the scene player. I guess you can play the game and just have devs yap over the cutscenes, but idk who'd wanna do that when the scene player already exists for that, and not every cutscene even has commentary

ManDude:
I don't think the language option was something we added, actually.

Aloqas:
I mean in the progress menu, it's selectable in the language options, which it originally wasn't

xTVaser:
> I mean in the progress menu, it's selectable in the language options, which it originally wasn't

Ah yeah that's probably something we can improve.  For now it's ended up being there for probably no other reason than making the menu code simpler -- but as you say, some people might want to play the game with commentary.  Doesn't hurt to have the option even if it's left in it's current state though.

PR #4282, opened 2026-05-20
libco: Update to latest version from ares

Ziemas:
This also removes the LIBCO_MPROTECT setting, I don't know why I enabled that but it prevented putting the context swap code in the text section.


xTVaser (changes requested):
Looks good / still compiles and plays fine.  The only thing I notice from the diff which we should probably change is the LICENSE file got dropped -- not sure if whatever source you got the update from also dropped the file but, we should probably retain it atleast in some form.

Ziemas:
True, there is no license file specifically for libco in ares, so I guess it's covered by their top-level license (still ISC). I put a copy of the relevant text in.

PR #4283, opened 2026-05-21
fix: Make port arg usable

himstan:
This fix makes the --port arg usable, so you can override the default port
Useful if you want to bind to multiple instances of the same game at the same time

Example: `/gk -v --port 8130 --game {{.GAME}} -- -boot -fakeiso -debug`


ManDude:
ok to test

PR #4285, opened 2026-05-23
[jak2/3] Fix dark/light eco HUD not flashing after save file is X hours old

dallmeyer:
Once `(-> *display* game-clock frame-counter)` gets too large, in the HUD flashing math the `sin` function breaks down at high values. It's effectively reading the number of ticks (times a multiplier) as degrees, so we can modulo that by 360 to avoid breaking `sin`

This applies to both the eco meter and the dark jak icon in jak 2 - the icon stops flashing twice as early because it's multiplied by x4 vs the eco meter x2. 

example:

you can see dark jak icon stops flashing around 2:36:02
https://youtu.be/3V2GneLSY14?t=9362

and dark eco meter stops flashing between 4:30 and 4:37
https://youtu.be/3V2GneLSY14?t=16235
https://youtu.be/3V2GneLSY14?t=16619

### DIFF
--- a/goal_src/jak2/engine/ui/hud-classes.gc
+++ b/goal_src/jak2/engine/ui/hud-classes.gc
@@ -187,8 +187,8 @@
   (set-as-offset-from! (-> this sprites 7) (-> this sprites 8 pos) 40 40)
   (set-as-offset-from! (-> this sprites 21) (-> this sprites 8 pos) 25 25)
   (let ((v1-12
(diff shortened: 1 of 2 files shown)


wojtulace:
That bug was annoying. I always wondered why it flashes correctly when starting a mission from the mission select menu. Turns out it was caused by a save file. I had a feeling that was the case, but I thought it was related to the number of completed missions or something

ManDude (approved):

PR #4290, opened 2026-05-25
Support identical imgui controller labels

### LINKED ISSUE #4289: Multi-port controller adapter error
Image

Image

Image

It will not allow me to switch to the 2nd port, clicking does nothing. I can only select the 1st port if none are selected.
I accept this is likely not a problem for most adapters.
Seems like it has a problem with both ports having the same name?

Zedb0T:
ImGUI does have ways to make sure the Internal IDs are unique and separate from the display label, however I still thought that appending a count to the end of the name was more clear/understandable.

If you add `ImGui::PushID(i);` Into the loop it appends i to the end of the internal ID without modifying the display label. The other alternative would be to use a syntax like `ImGui::Button((controller_name + "##option1").c_str());` etc....
https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-have-multiple-widgets-with-the-same-label

In both cases I think just adding a number that increments communicates that its a different controller cleaner to the user.

Closes #4289

### DIFF
--- a/game/graphics/opengl_renderer/debug_gui.cpp
+++ b/game/graphics/opengl_renderer/debug_gui.cpp
@@ -166,12 +166,18 @@ void OpenGlDebugGui::draw(const DmaStats& dma_stats) {
         if (ImGui::TreeNode(label.c_str())) {
           const auto num_controllers =
               Display::GetMainDisplay()->get_input_manager()->get_num_controllers();
+          std::unordered_map<std::string, int> controller_name_counts;
           for (int i = 0; i < num_controllers; i++) {
             const auto controller_name =
                 Display::GetMainDisplay()->get_input_manager()->get_controller_name(i);
+            int count = controller_name_counts[controller_name]++;
+            std::string display_name = controller_name;
+            if (count > 0) {
+              display_name += " (" + std::to_string(count) + ")";
+            }
             auto is_controller_active =
                 Display::GetMainDisplay()->get_input_manager()->get_controller_index(port) == i;
-            if (ImGui::RadioButton(controller_name.c_str(), is_controller_active)) {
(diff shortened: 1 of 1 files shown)


game/graphics/opengl_renderer/debug_gui.cpp
```diff
@@ -166,12 +166,18 @@ void OpenGlDebugGui::draw(const DmaStats& dma_stats) {
...
           const auto num_controllers =
               Display::GetMainDisplay()->get_input_manager()->get_num_controllers();
+          std::unordered_map<std::string, int> controller_name_counts;
           for (int i = 0; i < num_controllers; i++) {
             const auto controller_name =
                 Display::GetMainDisplay()->get_input_manager()->get_controller_name(i);
+            int count = controller_name_counts[controller_name]++;
+            std::string display_name = controller_name;
+            if (count > 0) {
+              display_name += " (" + std::to_string(count) + ")";
```
xTVaser: `display_name = fmt::format("{} ({})", display_name, count);`

But yeah agreed about your choice -- the usual way is the `##...` suffix, but this is something we should make obvious to the user that "you have two controllers named the same"

xTVaser:
Was thinking about this -- should the same change be made to the progress menu as well?

Zedb0T:
Yeah I've been tied up, but that was the one area I wanted to mention before we progress (lol) Originally I wanted to use the same logic for appending something to the end of the name that the progress menu does, it iterates the controllers with a dotimes and add a 1/2/3 I didn't see a way that both options would produce different results, but maybe it's worth deciding on a consistent cross language approach to be "certain"

xTVaser:
if the worry is about the menu matching the same order as the imgui one, i wouldn't worry about that -- the imgui one is added as a quick debugging tool basically.

PR #4291, opened 2026-05-26
jak2/jak3: Prevent subtitle processing when subtitles/gamehints are disabled

trippjoe:
This change reuses the existing subtitle visibility condition before doing subtitle processing, instead of only checking it immediately before drawing.

### DIFF
--- a/goal_src/jak2/pc/subtitle2-h.gc
+++ b/goal_src/jak2/pc/subtitle2-h.gc
@@ -259,6 +259,7 @@
     (add-to-queue (_type_ gui-connection) gui-connection)
     (get-active-subtitles (_type_) int)
     (subtitle-format (_type_ subtitle2-line) string)
+    (subtitles-enabled? (_type_) symbol)
     (draw-subtitles (_type_) int)
     (debug-print-queue (_type_) int)
     (debug-print-speakers (_type_) int)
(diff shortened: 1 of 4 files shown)


ManDude:
What's the reason for this change? The reason I run this on every frame regardless of setting is so that toggling it on and off keeps it in a consistent state, instead of now needing to handle this new case of the state being reset at any time. It's also less code.

trippjoe:
The reason for this change is to improve performance/make jak2 playable on silicon mac when subtitles and/or game-hints are disabled. As previously noted, Rosetta dislikes `print-game-text`, short circuiting the subtitle processing prevents `print-game-text` from being reached and prevents performance issues.

ManDude:
`print-game-text` calls `get-string-length` multiple times to do its word wrapping correctly. `draw-string-asm` and `get-string-length-asm` are most likely the real bottlenecks. Not worth complicating existing code when it's getting bypassed by enabling an option that's desirable in the first place. There are also other text-heavy sections.

trippjoe:
> Not worth complicating existing code when it's getting bypassed by enabling an option that's desirable in the first place.

I can't really follow what you're saying here, can you please clarify this?

ManDude:
"Turn all subtitles off for potentially better performance in some areas" doesn't seem like a good tradeoff and all this is doing is saving two `print-game-text` calls per frame.

trippjoe:
The subtitle strings should be formatted and measured every frame even though they aren't going to be rendered.

Do you believe this?

ManDude:
the formatting amounts to calling `format` which is already called dozens of times every frame and every time any text is printed

Issue #3841, opened 2025-01-15
[Mac] Rosetta translation layer causes performance issues, specifically noticable during dialog


jeffreytwoods2:
### Acknowledgements

- [x] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak2+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [x] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

Any time a character speaks during gameplay, the frame rate drops to about ~17fps. It happens with any dialog that has subtitles available, whether or not they are turned on: Praxis's broadcasts in the slums, guards saying "Patrolling this sector", etc. Does not happen during cutscenes.

Hardware trace:

<img width="1051" alt="Image" src="https://github.com/user-attachments/assets/d9abb532-e4fd-4dad-87fe-4843ef79dd04" />

A Crimson Guard said, "Seal off the area" at 0:02. I toggled the "Limit frames" and "Sleep during frame limit" settings; no change.

### How To Reproduce

1. Start the game and go to the slums
2. Hop in a hover car and begin driving through a quiet area
3. Wait for a Crimson guard to say something, or drive next to a Praxis loudspeaker

### Does this problem occur on original hardware or PCSX2?

Yes, it's unique to OpenGOAL

### Expected Behavior

When dialog occurs in the game, the frame rate should remain relatively steady.

### Environment Information

CPU/GPU: Apple M1 (2020 MacBook Pro, 16GB)
OS: MacOS Sequoia 15.2
Launcher: v2.6.3
Tooling: v0.2.21

### Game Version

NTSC 1.0 (black label)

### Have you set the game to something other than `60fps`?

Yes

xTVaser:
Another M1 user reported this in the discord.  To my knowledge this was never a problem on the Intel Macs, and at this time I have no way to test that anymore.

So assuming that is correct, my guess would be that the audio related components are poorly handled by Rosetta's translation layer and there may not be much that we can do.  However that's partially a guess, there may be a simple solution once investigated.

It would be interesting to hear if this also effects Jak 1, if it doesn't that would narrow things down a bit.

jeffreytwoods2:
Just tried on Jak 1; it does the same thing. I'll see if my university has Intel Macs in any computer labs (doubtful, though).

sirj0k3r:
Can confirm the issue on Apple Silicon.

[...]

PR #4293, opened 2026-05-29
jak2: Subtitles/Menu Items slow down fixes for ARM-Based Macs

LKNSI:
Hi!

### Upfront, **this is left as a draft** because I know that I didn't probably follow any of the communities / guidelines (and probably failed some CI (sorry)) 

I wanted to at least signpost what I had to fix for maybe future reference for issues related to:

https://github.com/open-goal/jak-project/pull/4291

https://github.com/open-goal/jak-project/issues/3841

It appears this is a problem that only affects Rosetta (trashing most likely) for things ingame that hit the subtitle (or menu settings) aspects of the codebase - it causes insane slowdowns. After profiling, it came down to the generated MIPS functions that did things around text length measurement. Previous people have noted that disabling subtitles fixes this, which is incorrect. Where this problem begins frankly happens before we even check if subtitles are enabled or not. So ultimately replacing those with C++ functions eliminated this, but I can understand if that's not exactly a goal of this project.

In addition, whenever an NPC (like guards) rendered a voice line, the issue would happen but for a different reason. It appears that on every frame, we were searching for the scene name **very** expensively by transforming the strings to uppercase and then walking that whole table each frame. The fix is around the bounds of hashmaping it. Though i still notice that there's maybe P99 stutters as a result, so not completely fixed either - but wayyy more playable on arm-based macs now.

**In-game, tested up until Jak steals the barons flag. It plays perfectly!**


LKNSI:
@trippjoe since you were working on this problem similarly

goal_src/jak2/engine/ui/text.gc
```diff
@@ -453,187 +453,27 @@ in a single text group and file."
...
+
+(defun pc-print-game-text-draw-line ((line string) (ctxt font-context) (bucket bucket-id))
+  (with-dma-buffer-add-bucket ((buf (-> *display* frames (-> *display* on-screen) global-buf))
+                               bucket)
+    (draw-string line buf ctxt)
+    (set! (-> ctxt color) (-> *font-work* last-color))
+    )
+  (none)
+  )
+
```
LKNSI, line 468: after the original changes, i noticed that the mission list screen got messed up like below - so the above fixes it.

Image

xTVaser:
At a glance -- I would assume the performance gains you are seeing are entirely due to the cache that the AI generated code added, it would also explain why you are still rarely seeing hitches (when the cache misses).

mips2c code in theory should be slower than code emitted by goalc, so it would be a far less invasive / easier to read / and likely a more performant improvement to just add a cache to the `.gc` file to basically memoize the calls to the font functions.  Doing this in C++ likely isn't strictly required, but of course an LLM is going to have a much easier time writing C++ than OpenGOAL.

if this function in particular is a hot-path on rosetta, it would still be interesting to figure out exactly why / do some proper profiling as this still feels like a hack that doesn't solve the root cause (clang is going to compile the original mips2c well enough to not be such a bottleneck)

But yeah incredibly reluctant to see something like this merged:
- the font code is incredibly complex, and as you already found out, it did not do a perfect one shot
- the memoization cache endlessly grows, there's no TTL on it, which is probably not great
- we now have some mips2c functions that are implemented completely differently than others

LKNSI:
> At a glance -- I would assume the performance gains you are seeing are entirely due to the cache that the AI generated code added, it would also explain why you are still rarely seeing hitches (when the cache misses).
> 
> mips2c code in theory should be slower than code emitted by goalc, so it would be a far less invasive / easier to read / and likely a more performant improvement to just add a cache to the `.gc` file to basically memoize the calls to the font functions. Doing this in C++ likely isn't strictly required, but of course an LLM is going to have a much easier time writing C++ than OpenGOAL.
> 
> if this function in particular is a hot-path on rosetta, it would still be interesting to figure out exactly why / do some proper profiling as this still feels like a hack that doesn't solve the root cause (clang is going to compile the original mips2c well enough to not be such a bottleneck)
> 
> But yeah incredibly reluctant to see something like this merged:
> 
> * the font code is incredibly complex, and as you already found out, it did not do a perfect one shot
> * the memoization cache endlessly grows, there's no TTL on it, which is probably not great
> * we now have some mips2c functions that are implemented completely differently than others

Appreciate the response! AI generated indeed (dw, no intention of this being merged!), lisp-like languages are very much out of my wheelhouse - but hopefully this can be directionally helpful for others in the future. It'll probably create more issues in random ways, but since I'm able to enjoy the game in relatively good state (i say this with caution), wanted to share!

There are definitely signs that similar issues exist in ways probably of the same ilk (dark jak is a slowdown issue too.) But without proper profiling honestly i'd just be speculating atp.

xTVaser:
> There are definitely signs that similar issues exist in ways probably of the same ilk (dark jak is a slowdown issue too.) But without proper profiling honestly i'd just be speculating atp.

Right exactly, it's very unlikely that this is the only function that causes Rosetta to freak out -- it's just potentially one of the obvious low hanging fruit (easy to reproduce, easy to point to, it's used _a lot_ everywhere, etc).  And it of course isn't practical for us to memoize every function call, so a proper over-arching solution has to be made.

xTVaser:
Should no longer be required -- ARM64 backend has been merged https://github.com/open-goal/jak-project/pull/4390

PR #4302, opened 2026-06-06
jak1/2/3: fix high fps slowdown

gratefulforest:
`DISPLAY_FPS_RATIO` shrinks values relative to the higher rate of operations when the framerate is higher.

However, currently it always applies the limit needed for the full framerate, even when the game is not running at full speed.

During slowdown, this causes a slower speed for anything adjusted by `DISPLAY_FPS_RATIO`, compared to everything else in the game.

This small fix incorperates slowdown into `DISPLAY_FPS_RATIO`, syncing the speeds of everything during slowdown and eliminating the strange speedups/slowdowns of things when the game lags.

Tested at various framerates in Jak 1/2/3.

This also has the positive side effect of correctly boosting values during slowdown at regular 60 FPS.

### DIFF
--- a/goal_src/jak1/engine/gfx/hw/display.gc
+++ b/goal_src/jak1/engine/gfx/hw/display.gc
@@ -62,7 +62,7 @@
   (-> this time-ratio))
 
 ;; new constant for use in high FPS scenarios
-(defconstant DISPLAY_FPS_RATIO (/ (-> *display* time-factor) 5.0))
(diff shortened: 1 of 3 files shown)


gratefulforest:
Oh, there's one small issue here. Since `DISPLAY_FPS_RATIO` can go above 1.0 when lagging at 60 FPS, usually this is helpful, but when lerping, it can pass it's target if a lerp step ends up above 1.0. I just need to clamp lerps that are scaled by `DISPLAY_FPS_RATIO` to min 1.0.

PR #4311, opened 2026-06-14
d/jx: Another round of files

xTVaser:


goal_src/jakx/engine/common-obs/particle-curves.gc
```diff
@@ -5,5 +5,167 @@
...
+   (lifetime-base    time-frame)
+   (lifetime-offset  time-frame)
+   (flags            particle-curve-flags)
+   )
+  )
+
+
+(def-mips2c birth-func-curve (function int sparticle-cpuinfo sparticle-launchinfo none))
+
+(def-mips2c live-func-curve (function sparticle-system sparticle-cpuinfo vector none))
```
Hat-Kid, line 46: These need to be hooked up still.

goal_src/jakx/engine/entity/entity-h.gc
```diff
@@ -118,13 +121,46 @@ that gets accessed by the accompanying process."
     )
   )
 
-;; type entity-camera is defined here, but it is unknown to the decompiler
+(deftype entity-camera (entity)
+  ((connect  connectable  :inline)
```
Hat-Kid: This is actually a `quaternion` as I recently discovered in #4310.

goal_src/jakx/engine/geometry/path-h.gc
```diff
@@ -20,3 +20,314 @@
...
+      (let ((v0-0 (entity-actor-lookup arg2 'path-actor 0)))
+        (if v0-0
+            (set! arg2 v0-0)
+            )
+        )
+      )
+    (let ((s5-0 (the-as object 0)))
+      (let* ((s4-0 (get-distmap-name arg1))
+             (sv-112 (get-knot-name arg1))
+             (sv-16 0)
```
Hat-Kid: This is a `res-tag`. If you add a `["_stack_", 16, "res-tag"]` type cast, it should improve the output and get rid of the `.pcpyud`s. Ditto for stack slot 32 and 48.

goal_src/jakx/engine/gfx/background/wind-work.gc
```diff
@@ -7,3 +7,13 @@
...
+(define *wind-work* (new 'static 'wind-work
+                      :wind-const (new 'static 'vector :data (new 'static 'array float 4 0.5 100.0 0.0166 -1.0))
+                      )
+        )
+
+(set! (-> *wind-work* to-ptrs 0) (the-as uint (+ 8192 #x70000000)))
+
+(set! (-> *wind-work* to-ptrs 1) (the-as uint (+ 4096 #x70000000)))
+
+(set! (-> *wind-work* to-ptrs 2) (the-as uint (+ #x70000000 0)))
```
Hat-Kid: These should be changed to use the fake scratchpad.

goal_src/jakx/engine/process-drawable/process-focusable.gc
```diff
@@ -5,5 +5,68 @@
 ;; name in dgo: process-focusable
 ;; dgos: ENGINE, GAME
 
+;; +++focus-status
+(defenum focus-status
```
Hat-Kid, line 10: IIRC, these values were previously retrieved from the `Target Stats` debug print which seems to have been removed in this game, so these might not be accurate anymore, but it's probably too early to say for sure.

goal_src/jakx/engine/process-drawable/process-nettable-h.gc
```diff
@@ -7,3 +7,235 @@
...
+  (let* ((s3-0 37)
+         (s5-0 (the-as (function process-nettable net-process-class-info none) (-> arg0 method-table s3-0)))
+         )
+    (cond
+      ((= arg0 process-nettable)
+       (set! (-> arg1 msg-count) 0)
+       (s5-0 (the-as process-nettable #f) arg1)
+       )
+      (else
+        ;; TODO JAKX - FIX COMPILER BUG (write-msg-map-rec (-> arg0 parent) arg1)
```
Hat-Kid: Does `defun-recursive` not work?

goal_src/jakx/kernel/gstring.gc
```diff
@@ -191,25 +191,7 @@
     )
   )
 
-(defun name= ((arg0 object) (arg1 object))
```
Hat-Kid, line 194: This function and `string-word-wrap` were removed for some reason? I guess because they fail CFG and were copied over from ref tests.

goal_src/jakx/levels/common/net-enemy-h.gc
```diff
@@ -7,3 +7,148 @@
...
+
+(set! (-> *net-process-class-info-net-enemy* owner-type) net-enemy)
+
+(net-process-class-register *net-process-class-info-net-enemy*)
+
+;; WARN: Return type mismatch net-process-class-info vs none.
+(defmethod process-nettable-method-40 ((this net-enemy))
+  *net-process-class-info-net-enemy*
+  (none)
+  )
```
Hat-Kid, line 148: Return type might be wrong here. Not sure if there's any callers yet.

Hat-Kid (changes requested):

goal_src/jakx/engine/process-drawable/process-nettable-h.gc
xTVaser: the compiler first complained about the function call not having any arg/return info -- after i added the cast above that fixed that but then there was a new error (which I can't remember)

goal_src/jakx/engine/common-obs/particle-curves.gc
xTVaser: yeah i figured i can just do all the same-function mips2c in one shot, i should do that next round

goal_src/jakx/engine/geometry/path-h.gc
xTVaser: thanks for catching this, when i finish I think I'll run the linter script on all of the new goal_src files to catch anything else like that too.

goal_src/jakx/engine/process-drawable/process-nettable-h.gc
xTVaser: ah my bad, i forgot you have to explicitly declare the function as recursive -- that does satisfy things.

Hat-Kid (approved):

PR #4315, opened 2026-06-16
jak1/2/3: LED sliders

gratefulforest:
Adds both a means to disable LEDs or adjust their strength.
Uses the existing saved `controller-led-brightness` pc-setting as a minimalistic addition.
Named "Controller LED Strength" in Jak 1, and "LED Strength" in 2/3 under their Controller headings, consistent with existing menus.


ManDude:
Won't this cause the LED to have 1% brightness for everyone upon updating?

gratefulforest:
Thanks for checking it out, that's a good catch and a good call, thanks for taking the time. You're right, there is a reset/creation path, but no update path for an already created setting.
When I added this to my mod, I made extra settings, but later realized controller-led-brightness already exists (and is already hooked in to multiply strength, it's just always 1.0).
So it would be strange not to use it and have two different settings for LED brightness, at the same time, adding code to mask what the sliders display would be unlike the other menus and not an ideal implementation.

So I think it'd have to be one of:
1. Add an update path (I know display-settings is versioned, though I'm not sure about pc-settings).
2. Have users slide from 0.0 to 1.0 (not the most ideal user experience).
3. Have everyone's LEDs be dim until selecting (also far from an ideal user experience).

ManDude:
Ideally there is just a menu option that displays values in a percentage instead of 0.0 ~ 1.0. The volume sliders in jak 2/3 have this issue.

gratefulforest:
Thanks for pointing that out, I was wrong to think masking the menu would be unlike other menus, since the vanilla ones do [just that](https://github.com/open-goal/jak-project/blob/2490c94e97f8cd7a15b9a9803f80bbf748a5553b/goal_src/jak3/engine/ui/progress/progress-draw.gc#L793). I've put the range to 0.0 to 1.0, where they can be masked later along with the others, perhaps as a `show-as-percent` to go along `show-decimal` in the [generic slider](https://github.com/open-goal/jak-project/blob/2490c94e97f8cd7a15b9a9803f80bbf748a5553b/goal_src/jak3/pc/progress/progress-generic-h-pc.gc#L121).

PR #4319, opened 2026-06-19
989snd/overlord: implement `PLUGIN_MESSAGE` grain and `STRV` plugin system

### LINKED ISSUE #2582: [jak2] Missing "ambient" sounds
It seems like some ambient sounds are missing, like the guard comments in the city, and various wind noises in ruins. These are separate from the sounds that are tied to a specific object, like the propaganda things (those work).

Hat-Kid:
This implements the functionality for the `PLUGIN_MESSAGE` grain and the 989snd VAG plugin system, allowing sounds to queue up VAG streams.

Closes #2582


Hat-Kid:
@Ziemas if possible, please take a look and make sure I didn't break anything else in the process!

PR #4332, opened 2026-07-02
Jak 2 - Polish base game text translation + updating font db

rafalekkB:
basically initial commit to add support for polish base game text

needs work before you could push to master

im a little rusty so testing along side me if everything is working would be appreciated

jak2 pl menu


rafalekkB:
also a question if we could maybe work on the looks of the special characters because some really hinder readability

xTVaser:
> also a question if we could maybe work on the looks of the special characters because some really hinder readability

you can try tweaking them here https://github.com/open-goal/jak-project/blob/master/common/util/font/dbs/font_db_jak2.cpp#L21 (requires a rebuild of C++, making a change (or just saving the font file) and running `(mi)` to see the changes each time)

but all you can do is horizontally/vertically manipulate the existing glyphs in the game's font to try to recreate the letters, in other words -- there's a limit to how good you can make them..

rafalekkB:
will def take a look at it thanks

rafalekkB:
> you can try tweaking them here https://github.com/open-goal/jak-project/blob/master/common/util/font/dbs/font_db_jak2.cpp#L21 (requires a rebuild of C++, making a change (or just saving the font file) and running `(mi)` to see the changes each time)

is there a specific file that needs rebuilding or the entire project needs a rebuild when changing the font db because i changed it once to test, rebuilt the entire project, run (mi) in repl, booted into the game and the change was there but every time after this no matter what i do the changes no longer appear in the game after taking the same steps.
what am i missing exactly?

xTVaser:
You likely need to save the polish text file to force the `(mi)` to rebuild the font strings with the new encoding information.  Even if there are no changes to the file, saving it will change the last modified timestamp on the file and `(mi)` will think it has been modified.

rafalekkB:
i might retranslate some strings later on when i get more context by replaying the game but so far i think its ready

rafalekkB:
image
the accents used in polish should be looking good now

rafalekkB:
ive done one last pass through the entire game text, fixed everything i noticed and removed memcard related strings,
ive also finished translation on crowdin so polish should be 100% ready to go,
cheers

Issue #3140, opened 2023-11-03
Stars rendering as squares


dallmeyer:
### Acknowledgements

- [X] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak2+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [X] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

Not sure if this changed mid-run or not, but you can see in the clip the stars are not rendering correctly
https://clips.twitch.tv/SplendidImpartialChickenRuleFive--pr4qLg-xv1poy_X

### How To Reproduce

Not sure

### Does this problem occur on original hardware or PCSX2?

Yes, it's unique to OpenGOAL

### Expected Behavior

Stars render correctly

### Environment Information

windows, speedrunner mode enabled

### Game Version

NTSC 1.0 (black label)

### Have you set the game to something other than `60fps`?

No

ManDude:
the minimap border is also gone

notimportant651:
I believe I have the same issue here, but I could tell what the star squares were. When you open the scrapbook menu and press triangle to close it, the last image you left off on becomes the star sprite at night. You can repeat the process to keep changing the sprite for the stars.

Zedb0T:
Interesting, there is also some bug where it uses the wrong texture for praxis 1 shield, I wonder if its related

dallmeyer:
Encountered a variant of this bug again recently, but this time it was the lava particle effects in Dig Site that were square

Issue #3050, opened 2023-10-01
Praxis shield can sometimes use bogus textures


Zedb0T:
https://youtu.be/XSUauRIE4Vo?t=7644

Couldnt replicate so this can probably just be documents and closed as a 1/1000 bug but at least we have a record of it

water111:
Maybe it only happens if you've watched the intro cutscene.

dallmeyer:
Looks like this bug can happen in Jak 3 as well, I was doing a retail mode playthrough (so dolby splash screen did display before playing) and I got bogus dolby textures on the shield of the precursor at the end of the first subrails mission. Doesn't happen if you play the cutscene directly in debug mode

https://github.com/user-attachments/assets/f23f43e7-26cc-4a7f-9742-e0c929a5d4fa

https://github.com/user-attachments/assets/451cf253-2e00-46e0-b613-e665a4cf786a

I'm not sure if this is an artifact of the same underlying bug, but the warp effect during the arena cutscene when it flashes to Seem also bugged out during this retail mode playthrough. Again, doesn't happen in debug mode when playing the scene directly

https://github.com/user-attachments/assets/4ed5ccd1-4e13-4e84-b55f-204491489ef7

https://github.com/user-attachments/assets/2aaa784a-ee75-4748-b301-0d18af401a1c

1rocket1:
all cutscenes that use the blur effect have the same problem as the Arena cutscene

playing in retail mode with cheat mode active, or playing in debug mode, makes the blur effect show up correctly.
having debug text on the screen seems to be the difference, disabling debug-no-zbuf2 in renderer debug breaks the blur effect again

https://github.com/user-attachments/assets/f861e094-38bb-4eb6-a6b4-ec3c876bcc2e

PR #4354, opened 2026-07-22
fix(gfx): night stars rendered the Dolby logo and stretched (jak2/jak3)

semeaj:
## Problem

jak3's night stars (wasteland/Spargus at night) render as small horizontal rectangles instead of round dots. At 8x diagnostic size the rectangles become legible: each star quad is drawing the **Dolby Pro Logic II legal-screen image** instead of the star texture.

gk_jKQRN2jfmS

## Root cause

Two independent defects.

The star draw (`stars-dma`, mips2c method 35) looks up GS VRAM address 0, where the per-frame `sky-textures` page upload places `sky-hotdot`. That upload is processed in bucket order on PC just like PS2 (the `tex-lcom-sky-pre` upload handler runs one bucket before the sky draw), but within a TEX bucket uploads are deliberately flushed before the TextureAnimator so that live animator outputs (clouds, fog) can override the page textures they overlay. The animator's end-of-run pool sync then replays a claim for **every** entry in its VRAM map, which is append-only and never pruned, so a dead entry gets the same override privilege as live clouds. The boot sequence leaves a TextureAnimator composite registered at tbp 0 (its GPU texture still holding the Dolby legal screen), and that stale entry out-competed the sky page upload at address 0 every frame, forever. On PS2 this cannot happen: VRAM contents persist only until someone uploads over them, and a texture nobody writes anymore never wins the address back. (This corrects the first revision's description of the mechanism: the runtime slot-ownership traces identified the same winner, PC-ANIM/0, but the pool is not racing across threads; it is letting un-written entries re-claim.)

Separately, `update-pc-sky-scaling` aspect-corrects the X extents of every sky billboard (sun, green sun, all moon layers, jak3 day-star) except `star-coords`, so the star quads additionally displayed about 33% wider than tall at 16:9.

## Fix

The texture pool now distinguishes the two kinds of slot claim. `move_existing_to_vram` remains the write claim: it evicts the current owner, like any VRAM write on PS2. Its new partner `reassert_in_vram` is a residency claim: it may keep a slot the texture already owns, but never steals one, because stale VRAM contents lose to whoever uploads over them. The animator's pool sync issues a write claim only for entries the game actually wrote this frame (uploads, erases, and draws all stamp the entry) and a re-assert otherwise. Written-once-displayed-later outputs (bigmap) keep serving while unchallenged; entries the game stopped writing can no longer evict a live upload. The invariant is documented in the TexturePool header contract.

A second commit records the frame of each slot's most recent write claim and shows it in the texture pool debug window, so wrong-texture-served reports (#3050 and #3140 look like this same mechanism) can be diagnosed by inspection.

`update-pc-sky-scaling` (jak2 + jak3) scales the `star-coords` X extents by `relative-x-scale`, matching every other sky billboard.

## Test plan

- [x] Full Windows Release build from scratch on this branch; goalc-test suite (compiles the full jak2/jak3 GOAL trees, covering the sky-tng.gc edits): 1507/1509 passed, the two failures being a local test-environment artifact that reproduces without this branch
- [x] jak2 and jak3 boot smokes through the legal screens (where the stale boot composite is created) into title attract with level streaming
- [x] jak3 Spargus at night: stars sample `sky-hotdot` (round dots, not logos), square proportions at stock size
- [x] jak3 bigmap renders correctly (the written-once, displayed-later animator pattern most exposed to the re-assert path)
- [x] jak3 clouds and fog animate correctly (live per-frame anim claims still override the same-bucket sky page upload)
- [x] jak2 ctywide at night: stars clean
- [x] jak2 clouds, map, and progress screens verified on the prior revision; night sky re-verified on this one

---

I work off a self-hosted forge, so this GitHub account is quiet; the slot-ownership traces (per-frame VRAM slot 0 claims from both upload paths, and renderer-side bind servings) are available if anyone wants the raw data.

(AI-assisted)

### DIFF
--- a/game/graphics/opengl_renderer/OpenGLRenderer.cpp
+++ b/game/graphics/opengl_renderer/OpenGLRenderer.cpp
@@ -1451,6 +1451,7 @@ void OpenGLRenderer::dispatch_buckets(DmaFollower dma,
 
   m_render_state.version = m_version;
   m_render_state.frame_idx++;
+  m_render_state.texture_pool->set_frame_stamp(m_render_state.frame_idx);
   switch (m_version) {
     case GameVersion::Jak1:
       dispatch_buckets_jak1(dma, prof, sync_after_buckets);
--- a/game/graphics/opengl_renderer/TextureAnimator.cpp
+++ b/game/graphics/opengl_renderer/TextureAnimator.cpp
@@ -1223,6 +1223,7 @@ void TextureAnimator::handle_texture_anim_data(DmaFollower& dma,
   m_in_use_temp_textures.clear();  // reset temp texture allocator.
   m_force_to_gpu.clear();
   m_skip_tbps.clear();
+  m_current_frame_idx = frame_idx;
 
   // loop over DMA, and do the appropriate texture operations.
   // this will fill out m_textures, which is keyed on TBP.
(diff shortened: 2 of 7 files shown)


trippjoe:
related to #3050 and #3140

semeaj:
Reworked the animator half of this PR; same fix, different home. The per-run tick gate at the animator call site is gone, and the rule now lives in the texture pool as ownership semantics: `move_existing_to_vram` is a write claim and steals the slot, the new `reassert_in_vram` may keep a slot but never steals one, and the animator issues write claims only for entries the game wrote this frame. Draws stamp the entry as well now, not just uploads and erases (the tick version missed that path), and stale entries keep unchallenged slots by construction instead of by special case, which is what written-once patterns like bigmap rely on.

I also corrected the root-cause description in the body: uploads are processed in bucket order on PC too, so this was never a cross-thread race. The actual mechanism is that the animator's end-of-run sync replays claims for its whole append-only map, which gives dead entries the same override privilege as live clouds.

The pool debug window now shows each slot's most recent write-claim frame, so reports in this class (#3050, #3140 look like the same mechanism) should be diagnosable by inspection rather than instrumented builds.

Re-validated on this revision: jak3 night stars, bigmap, and clouds; jak2 night stars; boot smokes for both games.

Issue #3551, opened 2024-06-06
FAILED: common/CMakeFiles/common.dir/custom_data/TFrag3Data.cpp.obj / FAILED: CMakeFiles/cmTC_d2c3d.dir/CurlTests.c.obj 


shoober420:
### Acknowledgements

- [X] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak1+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [X] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

Im compiling on Windows 11 using MSYS2. After running `cmake -B build && cmake --build build -j 16`, I get the following build error.

Build Logs:
[CMakeError.log](https://github.com/user-attachments/files/15692826/CMakeError.log)
[CMakeOutput.log](https://github.com/user-attachments/files/15692862/CMakeOutput.log)

[134/1164] Building CXX object common/CMakeFiles/common.dir/formatter/formatter_tree.cpp.obj
C:/msys64/home/penny/jak-project/common/formatter/formatter_tree.cpp: In member function 'void FormatterTree::construct_formatter_tree_recursive(const std::string&, TSNode, FormatterTreeNode&, std::optional<std::__cxx11::basic_string<char> >)':
C:/msys64/home/penny/jak-project/common/formatter/formatter_tree.cpp:146:10: warning: unused variable 'debug_child' [-Wunused-variable]
  146 |     auto debug_child = ts_node_string(child_node);
      |          ^~~~~~~~~~~
[138/1164] Building C object third-party/sqlite3/CMakeFiles/sqlite3.dir/sqlite3.c.obj
ninja: build stopped: subcommand failed.

### How To Reproduce

Open MSYS2
Clone repo
Run cmake

### Does this problem occur on original hardware or PCSX2?

Yes, it's unique to OpenGOAL

### Expected Behavior

The build to complete.

### Environment Information

i9-13900kf
NVIDIA 4090
Windows 11 / MSYS2
Git Master (https://github.com/open-goal/jak-project/commit/b0b9fc4311c2a0a6e9a07b8b357d3b5e259043f4)

### Game Version

NTSC 1.0 (black label)

### Have you set the game to something other than `60fps`?

No

xTVaser:
The build logs that you included don't seem comprehensive, and the snippet from the compilers output is just a warning, not the actual error.

[...]

PR #4355, opened 2026-07-22
fix: Windows toolchain compatibility (curl 8.21 re-vendor, endless reconfigure loop)

semeaj:
## Problem

Windows builds break with a current local toolchain (Scoop LLVM 22.1.8, CMake 4.4.0, VS 2026), in two independent ways:

1. The build stops at curl's deliberate guard: `#error "no non-blocking method was found/used/set"` in `third-party/curl/lib/nonblock.c`.
2. From the second configure onward, `cmake --build` re-runs CMake in an endless loop (observed 42 consecutive reconfigure cycles in a single build). Likely the same mechanism behind the "endlessly building" VS 2026 note in `docs/setup/dev/vs.md`.

## Root cause

1. `third-party/curl/CMake/CurlTests.c` passes `int *` to `ioctlsocket()`, whose third parameter is `u_long *`. Clang 22 promotes `-Wincompatible-pointer-types` to a hard error in C, so the `HAVE_IOCTLSOCKET_FIONBIO` try_compile silently fails and `curl_config.h` never defines it. Upstream CI does not see this because the windows-2022 runner image ships an older LLVM. GCC 14 promotes the same warning to a hard error, which is very likely the `CurlTests.c.obj` failure reported from MSYS2 in open-goal/jak-project#3551. Upstream curl hit the identical problem with GCC 14 and fixed the probe in curl 8.8.0 (curl/curl#13578).
2. The root CMakeLists copies the build tree's `compile_commands.json` into `<src>/build/` for clangd using `configure_file()`, which registers its input as a configure dependency. CMake rewrites `compile_commands.json` late in every generation, after `CTestTestfile.cmake` and `cmake_install.cmake` (outputs of the same Ninja regen rule), so once the dependency is registered the rule is deterministically dirty and every `ninja` invocation re-runs CMake. A pristine first configure is safe (the file does not exist yet, so the `if(EXISTS ...)` guard skips the copy), which is why the loop looks machine- or IDE-specific.

## Fix

1. Per review, re-vendor `third-party/curl` at the `curl-8_21_0` tag (previously `curl-8_3_0`), which carries the upstream probe fix plus two years of upstream development; `vendor.yaml` updated to match. Adjustments the version jump forced:
   - curl 8.15 removed the native macOS Secure Transport backend (`CURL_USE_SECTRANSP`), so macOS now builds curl against OpenSSL like Linux. The two macOS workflows install Homebrew `openssl@3` and export `OPENSSL_ROOT_DIR` (keg-only), and the macOS setup docs gained the same two lines.
   - `CURL_BROTLI` / `CURL_ZSTD` switched to AUTO-detection in curl 8.10; pinned OFF to keep the previous no-compression behavior and avoid silently linking whatever the CI images happen to have.
   - curl's new top-level `BUILD_EXAMPLES` cache option (default ON) leaked into discord-rpc's identically named option and broke configure at a nonexistent `examples/send-presence` directory; pinned OFF ahead of the third-party subdirectories.

   The diff is dominated by the mechanical tag-tree swap under `third-party/curl` (linguist-vendored, collapsed in review). The hand-written changes are `CMakeLists.txt`, the two macOS workflows, `docs/setup/system/macos.md`, and `vendor.yaml`.
2. Swap `configure_file()` for `file(COPY ...)`: the same clangd copy with no configure dependency registered. (`file(COPY_FILE ... ONLY_IF_DIFFERENT)` would be cleaner still but requires CMake 3.21, above the declared `cmake_minimum_required(VERSION 3.10)`.)

## Test plan

- [x] Fresh `cmake --preset Release-windows-clang` (LLVM 22, no cache seeding) completes and logs `Enabled SSL backends: Schannel`; the FIONBIO probe passes without the previous `#error`
- [x] Full Windows Release build from scratch in the branch worktree (all 1422 targets)
- [x] goalc-test suite: 1509 passed, 0 failed
- [x] Second consecutive configure with `compile_commands.json` present: the regen rule in `build.ninja` has no `compile_commands.json` input; `<src>/build/compile_commands.json` is still refreshed for clangd
- [x] Repeated `ninja` invocations after a full build no longer re-run CMake
- [x] macOS Intel and ARM CI green (first exercise of the OpenSSL backend switch)

---

I work off a self-hosted forge, so this GitHub account is quiet; the configure logs and ninja dirty-node traces from the investigation are available if anyone wants the raw data.

(AI-assisted)


third-party/curl/CMake/CurlTests.c
```diff
@@ -288,7 +288,7 @@ int
 main ()
 {
 
-        int flags = 0;
+        unsigned long flags = 0;
```
xTVaser: it would be better to actually update the dependency instead of just taking the individual change from upstream. (assuming it's actually fixed there, your description says the word "upstream" a lot, but there's no link to the actual libcurl change now that i re-look at it)

semeaj: Fair, I should have linked it. The upstream fix is curl/curl#13578 (commit 7ca0a37, "cmake: fix HAVE_IOCTLSOCKET_FIONBIO test with gcc 14"), shipped in curl 8.8.0. GCC 14 promoted -Wincompatible-pointer-types to a hard error the same way Clang 22 does, and upstream's fix is this exact one-liner.

On updating the dependency: happy to do that. The vendored tree is a clean 8.3.0-DEV snapshot from #3037 with no local patches (the libpsl/libidn2/ssh2 opt-outs all live in the top-level CMakeLists), so it should be a straight drop-in of a current release tarball (8.21.0), plus checking the CMake option names that changed since 8.3 and letting CI validate all three platforms. Would you prefer that in this PR, or as a separate PR so the version jump gets its own review and this toolchain fix isn't blocked on it?

xTVaser: Updating the dependency on this PR would be preferred.  The cmake issue has been broken since VS 2026 was released in beta and as you say it's only a matter of re-vendoring the latest release; and this curl issue is only a problem on the latest clang -- im sure it can stay broken for a few more days 🙂 

Make sure to update the `vendor.yaml` file accordingly, it's how we keep track of the state of our dependencies https://github.com/open-goal/jak-project/blob/master/vendor.yaml#L45

semeaj: Done, re-vendored at curl-8_21_0 and updated vendor.yaml; this CurlTests.c patch is dropped since 8.21 already carries that upstream fix. The jump forced a few build adjustments, documented in the commit message: curl 8.15 removed the Secure Transport backend, so macOS now builds curl against Homebrew openssl@3 like linux (workflows and setup docs updated), and CURL_BROTLI/CURL_ZSTD/BUILD_EXAMPLES are pinned OFF (the last one was leaking into discord-rpc's identically named option and breaking configure). Verified locally on Windows with VS 2026 + clang 22: Schannel and the FIONBIO probe detect correctly, full build and the offline test suite pass.

xTVaser (approved):

PR #4364, opened 2026-07-26
[jak1] Large Rename/cleanup

water111:
Name most methods, variables.

To-do before merging:

- Add back the navigate NaN check
- Look into vector-copy!, vector-zero!, matrix-copy!


water111:
of course this was heavily AI assisted

PR #4369, opened 2026-07-27
custom levels: fix potential use after free in custom region generation on windows

Hat-Kid:
On Windows, the pairs produced by custom regions all become bogus, probably due to some compiler differences.
Changing the pairs that get stored in the regions to not be pointers should hopefully resolve this.
To be extra sure, I also pulled some `add_word(0)`s out into locals to avoid any argument evaluation order shenanigans.


xTVaser (approved):
lgtm 

Image

PR #4373, opened 2026-08-03
Fixed mirror world hud alignment.

### LINKED ISSUE #4362: Mirror World Secret Causes Orb Icon Misalignment
### Acknowledgements

- [x] I have [searched](https://github.com/open-goal/jak-project/issues?q=is%3Aissue+is%3Aopen+label%3Abug+label%3Ajak3+sort%3Aupdated-desc+) for duplicate or closed bug reports
- [x] I understand that I am supposed to provide my own legitimately obtained copy of the game

### Describe the Bug

Enabling the mirror world secret causes the precursor orb icon that displays after picking up an orb to be on the right side of the screen and therefore no longer aligned with the orb count and the visual effect that is supposed to go around the orb icon. The following image shows this.

Image

On PCSX2, the orb icon still appears correctly on the left side of the screen with mirror world enabled.

### How To Reproduce

Earn enough precursor orbs to purchase the mirror world secret, enable it, and then pick up another orb.

Yes, it's unique to OpenGOAL

### Expected Behavior

When I pick up an orb in mirror world mode, I expect the orb icon to appear on the left side of the screen, aligned with the orb count and visual effect.

### Environment Information

[opengoal-support-package.zip](https://github.com/user-attachments/files/30385012/opengoal-support-package.zip)

### Game Version

PAL

No

Jakjm:
This PR fixes the hud icon positions in mirror world, resolving #4362. 
Now the orb and skull-gem icons are correctly positioned on the left whether or not mirror world is active.

Normal:
Screenshot from 2026-08-03 13-01-26
Screenshot from 2026-08-03 13-05-17
Mirror world:
Screenshot from 2026-08-03 12-59-31
Screenshot from 2026-08-03 13-05-35

Closes #4362


goal_src/jak3/engine/ui/hud.gc
```diff
@@ -727,10 +727,12 @@
...
-        (if (get-horizontal-flip-flag *blit-displays-work*)
-            (set! (-> this icons s5-1 icon 0 root trans x) (the float (- 256 (-> this icons s5-1 pos x))))
-            (set! (-> this icons s5-1 icon 0 root trans x) (the float (+ (-> this icons s5-1 pos x) -256)))
-            )
+        ;; (if (get-horizontal-flip-flag *blit-displays-work*)
+        ;;     (set! (-> this icons s5-1 icon 0 root trans x) (the float (- 256 (-> this icons s5-1 pos x))))
+        ;;     (set! (-> this icons s5-1 icon 0 root trans x) (the float (+ (-> this icons s5-1 pos x) -256)))
+        ;;     )
+        (set! (-> this icons s5-1 icon 0 root trans x) (the float (- 256 (-> this icons s5-1 pos x))))
+        (set! (-> this icons s5-1 icon 0 root trans x) (the float (+ (-> this icons s5-1 pos x) -256)))
```
xTVaser: whenever we make a change like this, we put a comment labelling it with the following syntax, that way we have automated checks that warn us that we are potentially removing a fix (which can happen when people re-decompile a file for whatever reason).

example:
https://github.com/open-goal/jak-project/blob/641fb572e876750d8618434d9f7c1a531b43b366/goal_src/jak3/engine/ui/text.gc#L192

the `og:preserve-this` is the important part that needs to be exact, the rest of the comment is just whatever you want to explain why the change is being made.

xTVaser (changes requested):

goal_src/jak3/engine/ui/hud.gc
Jakjm: Hey @xTVaser, thank you for the feedback. 

It's probably not worth keeping the comments of the old code? 
Does it matter where I put the `og:preserve-this`? Can I put it right before 734 or should it go above the block that 734 and 735 are in?

I'm currently working on fixing a lot of menu text alignment for jak3, so I will need to use `og:preserve-this` a lot in that code as well. I'm also renaming variables/functions as I go, which makes the code a lot more readable and helps me work more efficiently at resolving the alignment issues. When re-decompiling files, will those clearer names get preserved as well? I guess my dumb question is: when files are re-decompiled, do contributors pay close attention to the existing version of the file to keep readability changes and not just functional ones?

Jakjm: Please let me know if the new revision is fine.

xTVaser: It's fine, yes you can remove the original code if you want.

xTVaser (approved):

PR #4385, opened 2026-08-19
goalc: arm64 SIMD encoders emitted the wrong instructions

Parker-Aphelion:
Alters the opcode constants and the lane index maths, so the pext, vector shift, shuffle, blend and splat helpers now emit the correct ZIP1/ZIP2, SHL/SSHR/USHR, INS and DUP machine encodings for 128-bit vector operations.

Adds an Imm5 field helper, which the lane-insert and duplicate encodings need.

Follows #4384, from the same Apple Silicon work.

### DIFF
--- a/goalc/emitter/IGenARM64.h
+++ b/goalc/emitter/IGenARM64.h
@@ -641,24 +641,6 @@ InstructionARM64 nop_vf();
 
 InstructionARM64 wait_vf();
 
-/*!
- * MOV <Vd>.S[dst_lane], <Vn>.S[src_lane]. copies one 32 bit lane and leaves the rest of Vd alone.
- * arm64 has no lane mask blend and no shuffle by immediate, so the ops that need one build
- * the result a lane at a time out of these.
- */
-InstructionARM64 ins_vf_lane(Register dst, u8 dst_lane, Register src, u8 src_lane);
-
-/*!
- * MOV <Vd>.H[dst_lane], <Vn>.H[src_lane]. the 16 bit version of ins_vf_lane.
- */
-InstructionARM64 ins_vf_lane_h(Register dst, u8 dst_lane, Register src, u8 src_lane);
-
-/*!
- * Broadcast one 32-bit lane of src to all four lanes of dst.
- * DUP <Vd>.4S, <Vn>.S[lane]
- */
-InstructionARM64 dup_vf_lane(Register dst, Register src, u8 lane);
-
 InstructionARM64 mov_vf_vf(Register dst, Register src);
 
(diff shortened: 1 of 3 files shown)


xTVaser (commented):
Thanks for looking into the existing / few remaining simd instructions, that was the next main step.  Are you planning on also replicating all the existing AVX emitter tests as well just so I don't do duplicate work?
- https://github.com/open-goal/jak-project/blob/master/test/test_emitter_avx.cpp
- https://github.com/open-goal/jak-project/blob/master/test/test_emitter_neon.cpp

goalc/emitter/IGenARM64.h
```diff
@@ -641,6 +641,24 @@ InstructionARM64 nop_vf();
...
+ * MOV <Vd>.H[dst_lane], <Vn>.H[src_lane]. the 16 bit version of ins_vf_lane.
+ */
+InstructionARM64 ins_vf_lane_h(Register dst, u8 dst_lane, Register src, u8 src_lane);
+
+/*!
+ * Broadcast one 32-bit lane of src to all four lanes of dst.
+ * DUP <Vd>.4S, <Vn>.S[lane]
+ */
+InstructionARM64 dup_vf_lane(Register dst, Register src, u8 lane);
+
```
xTVaser: I wouldn't add these to the header as they are ARM64 specific and just helpers for the generic blend/swizzle operations (one of the things I've been trying to cleanup is making the names of all the instructions not instruction-set-specific named, there are still a handful left)

Parker-Aphelion: Yes, planning to. Most of it is written already so please don't duplicate it. I've been splitting it into smaller PRs rather than sending one big one, figured that's easier to review.

I've got execution tests covering about two thirds of what `test_emitter_avx.cpp` hits. What's left is mainly the pcpy/vpshuf/vpsll family and the byte and halfword parallel compares.

One won't replicate: `loadvf_rip_plus_s32`. There's no PC-relative static load on the arm64 path, so it asserts instead of emitting.

I'll put them in `test_emitter_neon.cpp` unless you'd rather they went in `test_emitter_arm64.cpp` with the rest of #4377's.

Parker-Aphelion: Done, moved them into an anonymous namespace in IGenARM64.cpp. Makes sense, they only exist because blend/swizzle/splat need a lane insert and there's no single arm64 instruction for those.

xTVaser: simd instructions related tests should go in `test_emitter_neon.cpp` since that is ARM's SIMD instruction set (just like how AVX is x86's)

it doesn't really matter but it helps to split things up instead of adding another 1K lines to the other file.  Might as well split things up when we can

PR #4387, opened 2026-08-21
game: the sky blend was compiled out entirely on arm64

Parker-Aphelion:
Both sky blend functions have their whole body inside `#ifndef __arm64__`, so on
Apple Silicon they compile to nothing and the sky comes out black.

Added a NEON path and left x86 alone. It does the same maths as the SSE version
and there's a test comparing the two. The test fills the output with junk first,
otherwise an empty function would pass it.


xTVaser:
Assuming this is done in https://github.com/open-goal/jak-project/pull/4390 (has the needed changes for the sky code)

PR #4388, opened 2026-08-21
fix: three ARM64 runtime bugs found running jak1 natively on Apple Silicon

rodrigopsasaki:
Hey! :) 

I have been building my own native ARM64 backend for jak-project since May, independently of the ARM64 effort underway here (I actually didn't know there was one when I started, and using Rosetta was driving me nuts lol). 

It reached the finish line this week. I just completed 100% playthrough of jak1 on Apple Silicon, no Rosetta, around 150fps uncapped, with no glitches. I am now working on porting that backend onto this repo's own ARM64 encoder design (it seems much more disciplined than mine) so the two efforts merge instead of diverging; this PR is the first slice of that: three runtime bugs that live on current master, independent of any codegen backend, each found the hard way during real playing tests.

## jalr never calls back into GOAL on ARM64 macOS

`ExecutionContext::jalr` has platform branches for Linux, x86 macOS, and Windows, but none for ARM64 macOS, so every mips2c-to-GOAL callback silently does nothing and returns stale garbage. In practice: the character falls straight through the floor, because collision probes never get their GOAL callbacks answered. The fix wires in `_call_goal8_asm_arm64`, which already exists in `asm_funcs_arm64.s`, following the same convention `kscheme.cpp` uses. Receipt: `WithGameTests.Mips2C_CallGoal` fails on current master built for ARM64 macOS and passes with this fix.

## GOAL entry trampolines save 3 of the 18 callee-saved registers

`_call_goal_asm_arm64`, `_call_goal8_asm_arm64`, and `_call_goal_on_stack_asm_arm64` BLR into JIT code but only preserve x20-x22. GOAL's coroutine machinery loads whole register files from zeroed thread structs, so calls can return with x19, x23-x28, or d8-d15 clobbered, corrupting whatever the C++ caller kept there. I hit it as a null-pointer fault in `KernelCheckAndDispatch`, whose cached `&s7` in x27 got zeroed by the first successful dispatch. The fix extends all three trampolines to the full AAPCS64 callee-saved set. One pre-existing sp-ordering oddity in `_call_goal_on_stack_asm_arm64`'s restore path is flagged in the commit but deliberately not folded in.

## Sky blend functions compile to empty stubs on Apple Silicon

`SkyBlendCPU`'s two blend functions are wrapped in `#ifndef __arm64__`, which Apple clang defines, so both bodies compile to a bare `ret` and the sky uploads as a zero-filled texture every frame. Removing the guards is the whole fix: the SSE2 fallback path already routes through sse2neon and runs correctly. The same symptom was independently hit by the in-flight ARM64 work here this week, so this one is live, not hypothetical.

## What comes next

I'll continue trying to merge the worlds (let me know if this is helpful, I don't mean to step on any toes if this diverges with your decisions), and in my fork there will be some design docs and learnings from my own process iterating through this.

Thanks for the continued effort on this awesome project. I hope this helps!

### DIFF
--- a/game/graphics/opengl_renderer/SkyBlendCPU.cpp
+++ b/game/graphics/opengl_renderer/SkyBlendCPU.cpp
@@ -22,7 +22,6 @@ SkyBlendCPU::~SkyBlendCPU() {
 }
 
 void blend_sky_initial_fast(u8 intensity, u8* out, const u8* in, u32 size) {
-#ifndef __arm64__
   if (get_cpu_info().has_avx2) {
 #ifdef __AVX2__
     __m256i intensity_vec = _mm256_set1_epi16(intensity);
@@ -49,11 +48,9 @@ void blend_sky_initial_fast(u8 intensity, u8* out, const u8* in, u32 size) {
       _mm_storel_epi64((__m128i*)(out + (i * 8)), result);
     }
   }
-#endif
 }
 
 void blend_sky_fast(u8 intensity, u8* out, const u8* in, u32 size) {
-#ifndef __arm64__
   if (get_cpu_info().has_avx2) {
 #ifdef __AVX2__
     __m256i intensity_vec = _mm256_set1_epi16(intensity);
@@ -88,7 +85,6 @@ void blend_sky_fast(u8 intensity, u8* out, const u8* in, u32 size) {
       _mm_storel_epi64((__m128i*)(out + (i * 8)), out_val);
     }
   }
-#endif
 }
 
 SkyBlendStats SkyBlendCPU::do_sky_blends(DmaFollower& dma,
--- a/game/kernel/asm_funcs_arm64.s
+++ b/game/kernel/asm_funcs_arm64.s
@@ -172,10 +172,24 @@ _mips2c_call_arm64:
 _call_goal_asm_arm64:
   stp	x29, x30, [sp, #-16]!
   mov	x29, sp
-  ;; saved registers we need to modify for GOAL should be preserved
+  ;; Save the full AAPCS64 callee-saved set, not just the registers GOAL's
+  ;; own dispatch touches (x20-x22). GOAL's coroutine machinery loads whole
+  ;; register contexts from zero-initialized thread structs when it switches
+  ;; coroutines, so a GOAL call can come back with x19, x23-x28 or d8-d15
+  ;; clobbered even though GOAL-compiled code itself never touches them --
+  ;; and the C caller on the other side of this BLR is entitled to all of
+  ;; them under AAPCS64, same as any other call boundary.
   ; ARM64 requires 16-byte stack pointer alignment
   stp x20, x21, [sp, #-16]!
   str x22, [sp, #-16]!
+  stp x19, x23, [sp, #-16]!
+  stp x24, x25, [sp, #-16]!
+  stp x26, x27, [sp, #-16]!
+  str x28, [sp, #-16]!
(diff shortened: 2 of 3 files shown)


xTVaser:
Hi,

Quite frankly at this point it seems like every week (now lately multiple times a week) someone has pointed Claude/their AI agent of choice at the problem.  It's impractical to have so many people all working on the same problem / all solving it in their own way.  So it might be best for you to just keep this work in your fork.

As you've mentioned, this repo has been working on finishing the ARM work for quite a while (the asm_funcs file that re-wrote the manual assembly functions is several years old, so it's not a recent development).

I'll be incredibly blunt -- at this point the amount of work needed to finish the ARM64 stuff is quite small, if I get stuck, I can just point claude at the remaining issues myself, but spending time reviewing other people's AI work isn't of interest to me.  For example this makes changes to the assembly functions and what registers have to be preserved before even making any changes to the register allocator -- this doesn't make sense.

Hat-Kid:
Not accepting any commits co-authored by Claude.

rodrigopsasaki:
That's completely valid. I wasn't aware that you had these at any frequency. Apologies for adding to it.

Once again, thank you for the continued effort on this project!

Parrishu:
> Not accepting any commits co-authored by Claude.

Retard.

Hat-Kid:
> > Not accepting any commits co-authored by Claude.
> 
> Retard.

Looks like somebody is still mad over getting banned from Discord 4 years ago for bigotry. I suggest you go and find a different project to obsess over.

Parrishu:
> > > Not accepting any commits co-authored by Claude.
> > 
> > 
> > Retard.
> 
> Looks like somebody is still mad over getting banned from Discord 4 years ago for bigotry. I suggest you go and find a different project to obsess over.

You weren't there when I was banned, are you larping as if you had any part in that?
I seldom check the github to see what's new, and still every single one of your contributions is you being a cocky retard. I suggest you should get a hobby instead of making little Japanese cartoon girls your entire personality as a 25+ year old man, we call that pedophilia in Europe.

Get help.

Hat-Kid:
> You weren't there when I was banned, are you larping as if you had any part in that?

Oh, I was there. Both times. Who knows, maybe I was the one who banned you?

Unfortunately, I wasn't there for the times when you got banned from Luminar's Discord, the Jak subreddit Discord or the Jak Speedruns Discord... sounds like people don't want you in their servers for some reason. Wonder why.

Parrishu:
> > You weren't there when I was banned, are you larping as if you had any part in that?
> 
> Oh, I was there. Both times. Who knows, maybe I was the one who banned you?
> 
> Unfortunately, I wasn't there for the times when you got banned from Luminar's Discord, the Jak subreddit Discord or the Jak Speedruns Discord... sounds like people don't want you in their servers for some reason. Wonder why.

Do you actually talk like this in real life? You have a very corny comic-book esque way of structuring your sentences
ad hominem aside, oh no! I got banned for bullying they/them troons for pretending they are something they're not! my life is truly over now

this is such an own!
> 
image

A 27 year old man btw ^
peak masculinity right here, definitely doesn't look like an underweight socially outcasted dork like Kuitar at the beach irl

Parrishu:
anyway I think I've given you enough attention out of my day, this is probably the highlight of your month (no one talks to you)
good talk, expecting Jak X one day! maybe you can help decompile it instead of "ermm.... actually"ing every actual contributor for the past 4 years

peace

PR #4390, opened 2026-08-22
arm64: Apple Silicon support

### LINKED ISSUE #3841: [Mac] Rosetta translation layer causes performance issues, specifically noticable during dialog

Parker-Aphelion:
Got OpenGOAL building and running natively on Apple Silicon.

This is the rest of the port after the smaller arm64 emitter PRs. I was told pushing one big PR was okay. This covers goalc, the runtime, linker, kernel and the GOAL asm.

The new paths have native tests. I also found four more emitter bugs while running it. They're in scalar sqrt, 128-bit stores, scalar max and indexed stores above 4 GB.

Spent a while cleaning it up so it's easier to read. 4am gotta sleep lol

Closes #3841


game/kernel/common/codegen.h
```diff
@@ -0,0 +1,112 @@
...
+  __builtin___clear_cache((char*)addr, (char*)addr + size);
+#endif
+#else
+  (void)addr;
+  (void)size;
+#endif
+}
+
+// forward declaration avoids the game/runtime.h dependency
+extern u8* g_ee_main_mem_exec;
```
xTVaser: why avoid this dependency?  is there some kind of circular problem blocking the include?

game/kernel/common/codegen.h
```diff
@@ -0,0 +1,112 @@
...
+  u32 instr = 0xd2800000 | ((val & 0xffff) << 5) | reg;  // movz
+  memcpy(dst, &instr, 4);
+  int offset = 4;
+  for (int shift = 1; shift < 4; shift++) {
+    instr = 0xf2800000 | (shift << 21) | (((val >> (shift * 16)) & 0xffff) << 5) | reg;  // movk
+    memcpy(dst + offset, &instr, 4);
+    offset += 4;
+  }
+  return offset;
+}
```
xTVaser: could probably make this a lot more readable by just using the relevant `InstructionARM64` function(s) and memcpy'ing the encodings.  Rather than basically re-implementing them here

goal_src/jak1/kernel/gkernel-h.gc
```diff
@@ -119,7 +119,12 @@
 
 ;; trigger an exception. (GOAL used lw r0, 2(r0))
 (defmacro break ()
-  `(/ 0 0))
+  "trigger an exception."
+  ;; ARM64 division by zero does not trap
+  ;; the trailing 0 keeps the macro return type int
+  `(#cond
+     ((eq? INSTRUCTION_SET 'arm64) (begin (.break) 0))
+     (#t (/ 0 0))))
```
xTVaser: with this change, we should just be using `.break` for all instruction sets, no need for this conditional logic.

goalc/compiler/compilation/Function.cpp
```diff
@@ -266,9 +266,9 @@ Val* Compiler::compile_lambda(const goos::Object& form, const goos::Object& rest
       // got a result, so to_gpr it and return it.
 
       RegVal* final_result;
-      emitter::Register ret_hw_reg = emitter::gRegInfo.get_gpr_ret_reg();
+      emitter::Register ret_hw_reg = emitter::reg_info(m_instr_set).get_gpr_ret_reg();
       if (m_ts.lookup_type(result->type())->get_load_size() == 16) {
-        ret_hw_reg = emitter::gRegInfo.get_xmm_ret_reg();
+        ret_hw_reg = emitter::reg_info(m_instr_set).get_xmm_ret_reg();
```
xTVaser: `get_xmm_ret_reg` should be renamed to not be instruction set specific

goal_src/jak2/kernel/gkernel-h.gc
```diff
@@ -600,8 +600,12 @@
...
 (defmacro break ()
-  "crash the game by dividing by 0."
-  `(/ 0 0)
+  "trigger an exception."
+  ;; ARM64 division by zero does not trap
+  ;; the trailing 0 keeps the macro return type int
+  `(#cond
+     ((eq? INSTRUCTION_SET 'arm64) (begin (.break) 0))
+     (#t (/ 0 0)))
   )
```
xTVaser, line 605: similar

goal_src/jak3/kernel/gkernel-h.gc
```diff
@@ -566,8 +566,12 @@
...
 (defmacro break ()
-  "crash the game by dividing by 0."
-  `(/ 0 0)
+  "trigger an exception."
+  ;; ARM64 division by zero does not trap
+  ;; the trailing 0 keeps the macro return type int
+  `(#cond
+     ((eq? INSTRUCTION_SET 'arm64) (begin (.break) 0))
+     (#t (/ 0 0)))
   )
```
xTVaser, line 571: similar

goal_src/jakx/kernel/gkernel-h.gc
```diff
@@ -581,8 +581,12 @@
...
 (defmacro break ()
-  "crash the game by dividing by 0."
-  `(/ 0 0)
+  "trigger an exception."
+  ;; ARM64 division by zero does not trap
+  ;; the trailing 0 keeps the macro return type int
+  `(#cond
+     ((eq? INSTRUCTION_SET 'arm64) (begin (.break) 0))
+     (#t (/ 0 0)))
   )
```
xTVaser, line 586: similar

goal_src/goal-lib.gc
```diff
@@ -1,5 +1,8 @@
 ;;-*-Lisp-*-
 
+;; old compilers only target x86
```
xTVaser: not sure i understand this comment, i guess it's just to justify the default value here?  But this will be set by `goalc` going forward so, probably not needed to explain at all.

game/system/IOP_Kernel.cpp
```diff
@@ -46,7 +46,7 @@ IOP_Kernel::IOP_Kernel() {
 
 u32 IOP_Kernel::GetSystemTimeLow() {
   auto delta_time = time_point_cast<microseconds>(steady_clock::now()) - m_start_time;
-  return delta_time.count() * 36.864;
+  return iop::detail::system_time_low_from_microseconds(static_cast<u64>(delta_time.count()));
```
xTVaser: this seems unrelated to ARM, but if you just want to extract things into a helper, I don't think it has to be so heavily nested in a namespace in the header, especially since it's not used anywhere else.

you could probably just do the exact same calculation here. `system_time_low_from_microseconds` also doesn't replicate the existing behaviour, it's no longer just a simple `delta_time.count() * (36864/1000)`.  Not sure if that matters but don't know why we are making this change in this PR

test/goalc/CMakeLists.txt
```diff
@@ -1,43 +1,24 @@
...
-        ${CMAKE_CURRENT_LIST_DIR}/test_debugger.cpp
-        ${CMAKE_CURRENT_LIST_DIR}/test_game_no_debug.cpp
-        ${CMAKE_CURRENT_LIST_DIR}/test_jak2_compiler.cpp
-        ${CMAKE_CURRENT_LIST_DIR}/test_variables.cpp
-        ${CMAKE_CURRENT_LIST_DIR}/test_with_game.cpp
-        ${CMAKE_CURRENT_LIST_DIR}/test_type_consistency.cpp
-        ${CMAKE_CURRENT_LIST_DIR}/test_vector_float.cpp
-    )
-endif()
+# GOAL tests use the current process architecture
```
xTVaser: unnecessary comment as this is heavily implied, the existing code was just to avoid running tests that would break on ARM -- this PR solves that

goalc/emitter/Instruction.h
xTVaser, line 281: leave the TODO, because yeah it sounds like these can be cleaned up and are only relevant for x86.

test/goalc/test_goal_kernel.cpp
```diff
@@ -125,14 +125,27 @@ TEST_F(Jak1KernelTest, RunFunctionInProcess) {
 TEST_F(Jak1KernelTest, StateAndXmm) {
   shared_compiler->runner.c->run_test_from_string(
       "(ml \"test/goalc/source_templates/kernel/kernel-test.gc\")");
-  std::string result =
-      send_code_and_get_multiple_responses("(state-test)", 5, &shared_compiler->runner);
+#if defined(__aarch64__)
+  constexpr int response_count = 6;
+#else
+  constexpr int response_count = 5;
+#endif
```
xTVaser: should this be based off the compilers instruction set rather than pre-processor macros?

goalc/emitter/Register.cpp
```diff
@@ -69,10 +72,132 @@ RegisterInfo RegisterInfo::make_register_info() {
...
+  for (size_t i = 0; i < N_SAVED_GPRS; i++) {
+    info.m_saved_all[i] = info.m_saved_gprs[i];
+  }
+  for (size_t i = 0; i < N_SAVED_XMMS; i++) {
+    info.m_saved_all[i + N_SAVED_GPRS] = info.m_saved_xmms[i];
+  }
+
+  info.m_gpr_alloc_order = {X0,  X1,  X2,  X3,  X4,  X5,  X6,  X7,  X8,  X9, X10,
+                            X11, X12, X13, X14, X15, X19, X23, X24, X25, X26};
+  info.m_xmm_alloc_order = {V0, V1, V2, V3, V4, V5, V6, V7, V8, V9, V10, V11, V12, V13, V14, V15};
```
xTVaser: Seems like we are limiting things to only using 16 vector registers despite modern ARM64 platforms having 32 vector registers?

Also anything mentioning `xmm` should be updated to `simd` or something not x86 specific.

test/goalc/source_templates/jak2/kernel-test.gc
```diff
@@ -206,22 +241,76 @@
   0
   )
 
+(defun arm64-throw-vector-check ()
+  (#cond
+   ((eq? INSTRUCTION_SET 'arm64)
+    (let ((src (new 'stack-no-clear 'array 'uint64 4))
+          (dst (new 'stack-no-clear 'array 'uint64 4)))
```
xTVaser: these arm specific functions probably can be simplified and don't need the `cond` -- since they are called by other functions that have the exact same cond (the other cases will never be hit)

test/goalc/test_goal_kernel3.cpp
```diff
@@ -126,14 +126,27 @@ TEST_F(Jak3KernelTest, RunFunctionInProcess) {
 TEST_F(Jak3KernelTest, StateAndXmm) {
   shared_compiler->runner.c->run_test_from_string(
       "(ml \"test/goalc/source_templates/jak3/kernel-test.gc\")");
-  std::string result =
-      send_code_and_get_multiple_responses("(state-test)", 5, &shared_compiler->runner);
+#if defined(__aarch64__)
+  constexpr int response_count = 6;
+#else
+  constexpr int response_count = 5;
+#endif
```
xTVaser: similar here, i suspect this should be using the instr_str off the compiler, not the preprocessor macros

test/goalc/test_goal_kernel2.cpp
```diff
@@ -126,14 +126,27 @@ TEST_F(Jak2KernelTest, RunFunctionInProcess) {
 TEST_F(Jak2KernelTest, StateAndXmm) {
   shared_compiler->runner.c->run_test_from_string(
       "(ml \"test/goalc/source_templates/jak2/kernel-test.gc\")");
-  std::string result =
-      send_code_and_get_multiple_responses("(state-test)", 5, &shared_compiler->runner);
+#if defined(__aarch64__)
+  constexpr int response_count = 6;
+#else
+  constexpr int response_count = 5;
+#endif
```
xTVaser: similar

test/goalc/test_with_game.cpp
```diff
@@ -290,12 +295,15 @@ TEST_F(WithGameTests, DebuggerMemoryMap) {
 }
 
 TEST_F(WithGameTests, DebuggerDisassemble) {
+  // Zydis treats ARM64 words as x86, so instruction lengths are wrong
+#ifndef __aarch64__
```
xTVaser: yes, zydis only supports x86, that's why the `capstone` library was added.  But this PR doesn't attempt to setup the debugger so that's fine (just explaining why)

test/goalc/test_with_game.cpp
```diff
@@ -838,8 +846,14 @@ TEST_F(WithGameTests, SetU64FromFloat) {
...
-                                          {"#xffffffff80000000 1.0000 #xffffffffbf800000\n0\n"});
+  // FCVTZS matches PS2 positive overflow
+  // CVTTSS2SI returns integer indefinite on x86
+  const char* overflowed = emitter::kNativeInstructionSet == emitter::InstructionSet::ARM64
+                               ? "#x7fffffff"
+                               : "#xffffffff80000000";
+  shared_compiler->runner.run_static_test(
+      testCategory, "tricky-floats.gc",
+      {fmt::format("{} 1.0000 #xffffffffbf800000\n0\n", overflowed)});
 }
```
xTVaser, line 857: at a glance this makes me worried....the ARM64 output is _not_ matching the x86 one, and this was seemingly a test to handle tricky floating point cases.  So I'm inclined to think something is incorrect here?

game/kernel/common/kscheme.cpp
```diff
@@ -115,13 +147,19 @@ u64 call_goal(Ptr<Function> f, u64 a, u64 b, u64 c, u64 st, void* offset) {
...
+  return _call_goal_asm_arm64(a, b, c, (void*)(g_ee_main_mem_exec + f.offset), st_ptr, offset,
+                              g_ee_main_mem_exec);
+#elif defined __linux__
   return _call_goal_asm_systemv(a, b, c, fptr, st_ptr, offset);
 #elif defined __APPLE__ && defined __x86_64__
   return _call_goal_asm_systemv(a, b, c, fptr, st_ptr, offset);
 #elif _WIN32
   return _call_goal_asm_win32(a, b, c, fptr, st_ptr, offset);
+#else
+#error "call_goal: no GOAL call trampoline for this architecture or platform"
```
xTVaser: `ASSERT_MSG(false, "...");` would be better than `#error` to match with the rest of this repos error/logging handling.

game/kernel/common/kscheme.cpp
```diff
@@ -131,13 +169,18 @@ u64 call_goal(Ptr<Function> f, u64 a, u64 b, u64 c, u64 st, void* offset) {
...
+  return _call_goal_on_stack_asm_arm64(rsp, 0, 0, (void*)(g_ee_main_mem_exec + f.offset), st_ptr,
+                                       offset, g_ee_main_mem_exec);
+#elif defined __linux__
   return _call_goal_on_stack_asm_systemv(rsp, 0, 0, fptr, st_ptr, offset);
 #elif defined __APPLE__ && defined __x86_64__
   return _call_goal_on_stack_asm_systemv(rsp, 0, 0, fptr, st_ptr, offset);
 #elif _WIN32
   return _call_goal_on_stack_asm_win32(rsp, fptr, st_ptr, offset);
+#else
+#error "call_goal_on_stack: no GOAL call trampoline for this architecture or platform"
```
xTVaser: .

game/kernel/jak1/klink.cpp
```diff
@@ -133,7 +151,7 @@ uint32_t typelink_v3(Ptr<uint8_t> link, Ptr<uint8_t> data) {
  * Link symbols (both offsets and pointers) in "v3 equivalent" link data.
  * Returns a pointer to the link table data after the linking data for this symbol.
  */
-uint32_t symlink_v3(Ptr<uint8_t> link, Ptr<uint8_t> data) {
+uint32_t symlink_v3(Ptr<uint8_t> link, Ptr<uint8_t> data, bool mov32) {
```
xTVaser: explaining the purpose of the `mov32` arg might be nice

game/kernel/jak2/klink.cpp
```diff
@@ -89,6 +89,24 @@ uint32_t cross_seg_dist_link_v3(Ptr<uint8_t> link,
   return 1 + 3 * 4;
 }
 
+/*!
+ * Patches a 32-bit cross-segment GOAL address into a movz/movk pair.
+ */
+uint32_t arm64_other_seg_mov32_link_v3(Ptr<uint8_t> link, ObjectFileHeader* ofh, int current_seg) {
```
xTVaser: if these function's implementation are identical across all 4 games (atleast this one seems to be), then it should probably go in `common/klink`

game/kernel/jak3/klink.cpp
```diff
@@ -502,6 +503,24 @@ uint32_t cross_seg_dist_link_v3(Ptr<uint8_t> link,
   return 1 + 3 * 4;
 }
 
+/*!
+ * Patches a 32-bit cross-segment GOAL address into a movz/movk pair.
+ */
+uint32_t arm64_other_seg_mov32_link_v3(Ptr<uint8_t> link, ObjectFileHeader* ofh, int current_seg) {
```
xTVaser: identical, could be moved to common

game/kernel/jakx/klink.cpp
```diff
@@ -502,6 +503,24 @@ uint32_t cross_seg_dist_link_v3(Ptr<uint8_t> link,
   return 1 + 3 * 4;
 }
 
+/*!
+ * Patches a 32-bit cross-segment GOAL address into a movz/movk pair.
+ */
+uint32_t arm64_other_seg_mov32_link_v3(Ptr<uint8_t> link, ObjectFileHeader* ofh, int current_seg) {
```
xTVaser: identical, common

game/mips2c/mips2c_private.h
```diff
@@ -362,6 +385,8 @@ struct ExecutionContext {
 #elif _WIN32
     gprs[v0].du64[0] = _call_goal8_asm_win32(g_ee_main_mem + addr, args, 0, gprs[s6].du64[0],
                                              gprs[s7].du64[0], g_ee_main_mem);
+#else
+#error "jalr: no GOAL call trampoline for this architecture or platform"
```
xTVaser: `ASSERT_MSG(false, "...");`

game/mips2c/mips2c_table.cpp
```diff
@@ -725,6 +742,8 @@ void LinkedFunctionTable::reg(const std::string& name, u64 (*exec)(void*), u32 s
     addr = (u64)_mips2c_call_systemv;
 #elif _WIN32
     addr = (u64)_mips2c_call_windows;
+#else
+#error "mips2c trampoline: no implementation for this architecture or platform"
```
xTVaser: ASSERT_MSG

goalc/compiler/CodeGenerator.cpp
```diff
@@ -400,18 +395,181 @@ void CodeGenerator::do_goal_function_x86(FunctionEnv* env, int f_idx) {
...
+          m_gen.add_instr(
+              IGen::store_reg_offset_xmm32(m_gen, emitter::ARM64_REG::SP, op.reg,
+                                           allocs.get_slot_for_spill(op.slot) * GPR_SIZE),
+              i_rec);
+        } else if (op.reg.is_128bit_simd(m_gen.instr_set()) &&
+                   (op.reg_class == RegClass::VECTOR_FLOAT || op.reg_class == RegClass::INT_128)) {
+          m_gen.add_instr(
+              IGen::store128_xmm128_reg_offset(m_gen, emitter::ARM64_REG::SP, op.reg,
+                                               allocs.get_slot_for_spill(op.slot) * GPR_SIZE),
+              i_rec);
```
xTVaser: more `xmm` related functions that should be renamed

goalc/compiler/CodeGenerator.cpp
```diff
@@ -400,18 +395,181 @@ void CodeGenerator::do_goal_function_x86(FunctionEnv* env, int f_idx) {
...
+          // FLOAT spills use an 8-byte slot but move only 32 bits
+          m_gen.add_instr(
+              IGen::load_reg_offset_xmm32(m_gen, op.reg, emitter::ARM64_REG::SP,
+                                          allocs.get_slot_for_spill(op.slot) * GPR_SIZE),
+              i_rec);
+        } else if (op.reg.is_128bit_simd(m_gen.instr_set()) &&
+                   (op.reg_class == RegClass::VECTOR_FLOAT || op.reg_class == RegClass::INT_128)) {
+          m_gen.add_instr(
+              IGen::load128_xmm128_reg_offset(m_gen, op.reg, emitter::ARM64_REG::SP,
+                                              allocs.get_slot_for_spill(op.slot) * GPR_SIZE),
```
xTVaser: more xmm functions that should be renamed

goalc/compiler/IR.cpp
```diff
@@ -6,10 +6,11 @@
 
 #include "goalc/compiler/Env.h"
 #include "goalc/emitter/IGen.h"
+#include "goalc/emitter/IGenARM64.h"
 
 #include "fmt/format.h"
 
-// TODO ARM64 - just silencing errors while things are not implemented obviously
+// Each backend leaves some parameters unused.
 #pragma GCC diagnostic ignored "-Wunused-parameter"
```
xTVaser: we should be at a point where we can remove the global file ignore, and switch to marking the unused parameters on a individual basis.

goalc/emitter/CodeTester.cpp
```diff
@@ -301,7 +304,7 @@ void CodeTester::emit_push_all_simd() {
       emit(IGen::store128_gpr64_simd128(m_gen, RSP, XMM0 + i));
     }
   } else if (m_gen.instr_set() == InstructionSet::ARM64) {
-    // TODO - 16 or 32 simd regs available?
+    // The compiler allocator uses V0 through V15.
```
xTVaser: my understanding is aarch64 guarantees 32 vector registers (it was arm32 that only had 16) so since we basically hard-require a 64bit platform, we should be able to use all 32 simd regs.

goalc/emitter/IGenARM64.cpp
```diff
@@ -14,7 +14,7 @@
 // https://armconverter.com/?code=ret
 // https://developer.arm.com/documentation/ddi0487/latest
 
-// TODO ARM64 - just silencing errors while things are not implemented obviously
```
xTVaser, line 17: same here, we should be able to move away from disabling this lint on the entire file now

goalc/emitter/InstructionSet.h
```diff
@@ -2,4 +2,12 @@
 
 namespace emitter {
 enum class InstructionSet { X86, ARM64 };
-};
\ No newline at end of file
+
+//! The instruction set that the current process can execute.
+constexpr InstructionSet kNativeInstructionSet =
```
xTVaser: this is fine but it should only be used in testing (otherwise it prevents the intentional design of being able to cross-compile on non-arm platforms).  I would either add a better docstring explaining that it should only be used for testing OR perhaps this should be moved to one of the `test/` files/helpers instead of being in this `goalc` related file.

xTVaser (commented):
Did a full pass, nothing major just a collection of things that should be cleaned up now.

We should also change the macOS arm artifact to use the native cmake preset instead of rosetta:
https://github.com/open-goal/jak-project/blob/b59a6d1cf30816ec4b38bddcf5d450b453ec36f2/.github/workflows/release-pipeline.yaml#L82-L91

Parker-Aphelion:
Went through the stuff you called out and made the changes. The release job uses the native arm64 preset now and checks out the release tag. Fixed the artifact name and paths that depend on it too.

Changed the shared XMM names to SIMD where they are used by both backends. The real x86 names are left alone. ARM64 now has V0 to V31. V16 is still used as emitter scratch, so the allocator gets the other 31 and the tests save all 32. Also removed the file wide warning ignores.

The tests use the compiler target now instead of the host macros, so cross compiling still works. kNativeInstructionSet is marked for tests only and isn't used outside test/. All four break macros use .break now and the repeated arm64 checks are gone. Didn't add the debugger backend here. The disassembly test checks the compiler target since Zydis is x86 only.

The float one took a bit longer because it did look wrong. I wasn't sure x86 was the right result though, so I checked the EE manual. For finite positive overflow CVT.W.S says 0x7fffffff, so arm64 matches the PS2 there. x86 keeps its existing result and the test explains the difference now. The IOP helper is also gone and the old calculation is back inline. It only keeps the u64 conversion needed before the final u32 wrap on arm64.

For the MOV encoding, the runtime and emitter use the same arm64 helper now without making runtime depend on goalc. Also included the proper runtime header instead of the extra extern. Changed the unsupported paths to ASSERT_MSG, explained what patch_mov32 does and moved the copied linker function into common.

Ran the full tests again and built all-code for all four games on arm64 and x86. Everything passes locally.

Thanks for going through it

.github/workflows/release-pipeline.yaml
```diff
@@ -79,15 +79,16 @@ jobs:
...
     name: "🍎 MacOS"
-    uses: ./.github/workflows/macos-build.yaml
+    uses: ./.github/workflows/macos-build-arm.yaml
     with:
-      cmakePreset: "Release-macos-x86_64-clang-static"
-      cachePrefix: "static-arm"
+      cmakePreset: "Release-macos-arm64-clang-static"
+      cachePrefix: "static"
       uploadArtifacts: true
+      checkoutRef: ${{ needs.cut_release.outputs.new_tag }}
```
xTVaser: I don't think you need to make this change, none of the other workflow files have ever needed to explicitly checkout the tag.  All that needs to be done is to change the `cmakePreset`.  Also the `cachePrefix` needs to stay `static-arm` to ensure it does not clobber the other mac build cache -- there's no harm in the cache key being more explicit.

.github/workflows/release-pipeline.yaml
```diff
@@ -150,13 +151,13 @@ jobs:
...
-          ./.github/scripts/releases/extract_build_unix.sh ./ci-artifacts/macos-arm ./ci-artifacts/opengoal-macos-static-arm ./
+          ./.github/scripts/releases/extract_build_unix.sh ./ci-artifacts/macos-arm ./ci-artifacts/opengoal-macos-arm-static ./
           pushd ci-artifacts/macos-arm
           TAG_VAL=${{ needs.cut_release.outputs.new_tag }}
           tar czf ../final/opengoal-macos-arm-${TAG_VAL}.tar.gz .
           popd
-          chmod +x ./ci-artifacts/opengoal-macos-static-arm/lsp/lsp
-          cp ./ci-artifacts/opengoal-macos-static-arm/lsp/lsp ./ci-artifacts/final/opengoal-lsp-macos-arm-${TAG_VAL}.bin
+          chmod +x ./ci-artifacts/opengoal-macos-arm-static/lsp/lsp
+          cp ./ci-artifacts/opengoal-macos-arm-static/lsp/lsp ./ci-artifacts/final/opengoal-lsp-macos-arm-${TAG_VAL}.bin
```
xTVaser: not sure why this had to change either, these are temporary directories as part of the build and we know that they work -- changing this doesn't fix anything and might potentially cause CI failures in the next release.

Let's only change what is actually needed to make the artifacts be native ARM binaries...

xTVaser (commented):
immediate comments on the CI changes, will review the rest after the builds are passing again

Parker-Aphelion:
I was trying to make the tag checkout and ARM release naming explicit, just a little OCD lol didn’t like how it read. I reverted it.

The ARM test helper still had v8 and v9 in a function x86 was compiling, even though its caller was guarded. The comparison job also uses the compiler from master, so it needed a fallback for .break. The game macros still use .break for both targets like you suggested.

All 1,668 tests passed locally, so hopefully it passes now

xTVaser:
> The comparison job also uses the compiler from master, so it needed a fallback for .break. The game macros still use .break for both targets like you suggested.

This didn't have to change but it's nice to see the comparison succeed with this being the only change.  I'll revert this change in a follow-up PR (in other words, the failure is expected, its to let us know that the compiler is outputting something different, which is a very rare thing to happen at this stage in development).

This looks good to me to merge, I just want to do a test on Windows before doing so.

vilaB:
@Parker-Aphelion @xTVaser it's expected that the files for this release https://github.com/open-goal/jak-project/releases/tag/v0.3.6  with the arm naming still point to x86 builds?

```
opengoal-macos-arm-v0.3.6 % file *
data:      directory
extractor: Mach-O 64-bit executable x86_64
gk:        Mach-O 64-bit executable x86_64
goalc:     Mach-O 64-bit executable x86_64
```
I think some of the fixes from this branch arrived slightly late, the timing was close (this PR was merged in the same minute the release was produced, but it looks it didn't include the changes from here). 

Thank you in advance.

xTVaser:
~~These changes were definitely included in the release, it's likely a cmake / CI bug.  I'll look into it for the next release.~~

Actually no you're right they weren't https://github.com/open-goal/jak-project/compare/v0.3.6...master

Now that I remember it, i did this intentionally because I wanted a release just before the ARM merge just incase something was wrong with it.  Whenever we cut the next release, it will have these changes.  So it's expected that 0.3.6 are still x86 builds.

PR #4398, opened 2026-08-29
[jak3] Add autosplits for side mission golds, orbs per level

dallmeyer:
For hundo, all satellite, all open orbs gamers

Goes along with https://github.com/open-goal/speedrunning/pull/35


goal_src/jak3/engine/game/game-info.gc
```diff
@@ -697,6 +697,132 @@
   this
   )
 
+;; og:preserve-this added enum and helper functions to map orbs to money-per-level index
+(defenum level-money-data-index
```
dallmeyer: this is the reduced list of only levels that have orbs. the ordering is relative to their level-load-info indexing

there is this other level indexing used for `deaths-per-level`, but it doesn't give very much detail (all of haven is just 'city')
https://github.com/open-goal/jak-project/blob/427c9525581d7e66530e7d5997f18cf0827466a7/goal_src/jak3/engine/level/level-info.gc#L10-L28

goal_src/jak3/engine/util/script.gc
```diff
@@ -3623,7 +3623,10 @@
                                         (the-as pickup-type s5-0)
                                         (command-get-float (-> arg0 param 3) 0.0)
                                         #t
-                                        *entity-pool*
+                                        ;; og:preserve-this spawn from script-calling process, for money-per-level tracking
+                                        (if (type-type? (-> arg0 process type) process-drawable)
```
dallmeyer: this is really just for `was-pre-game` and `power-game` giving out orbs, they're the only two users of this `'birth-pickup` script. maybe I should get rid of the `if`?

goal_src/jak3/engine/game/game-info.gc
```diff
@@ -697,6 +697,132 @@
...
+  (waspala 18)
+  (desert 19)
+  (nsta 20)
+  (templex 21)
+  (templeb 22)
+  (templed 23)
+  (max 24)
+  )
+
+(defun level-money-data-remap ((lvl-idx int))
```
xTVaser: youll probably want to mark these other functions with the preserve comment AND/OR if you can, place them above the `;; decomp begins` part of the file

goal_src/jak3/engine/util/script.gc
xTVaser: i mean this does seem to change the behaviour from the original, is it needed or not?

dallmeyer: It is needed, or alternatively we could explicitly call `adjust-money-per-level` from `was-pre-game` and `power-game`, but that seems messier. I think I will just remove the `if` since they're the only users of this script

xTVaser (approved):

PR #4401, opened 2026-08-31
[decompiler] Extracting Sound Effects and Music

Jakjm:
This PR adds the ability to extract sound effects to .wav files for each of the four games, and music to .wav files for Jak1 and Jak2.

Basically what I have done is create a "fake player" which works the same as the player from Player.cpp except we just take the raw audio samples that would be sent to cubeb and then write them to wave files.

I shamelessly stole @Hat-Kid's code for flavas, for parsing the sound effect names and for finding where the sound banks start for Jak 1 lol.


xTVaser (approved):
Looks good, bunch of new code behind a new config flag that is defaulted to `false`, so a pretty safe merge / can always be iterated on if there are issues.

PR #4409, opened 2026-09-07
[jak3] `game-score` decomp cleanup, track minigame rank in autosplitter instead of task status

dallmeyer:
The current autosplit options for side missions that track bronze/gold/silver don't actually work (other than the explicit gold options I added in #4398). This reworks these to track the player's best rank in these missions (0=unranked, 1=bronze, 2=silver, 3=gold), and I have the [corresponding autosplitter script changes](https://github.com/open-goal/speedrunning/pull/35) to split when reaching gold (rank 3). Should be okay to repurpose some of these autosplitter fields since we haven't cut a release using them yet.

Along the way I filled out the `game-score` enum, made use of this enum in place of uints throughout the code, and renamed things for a few related `game-info` methods.


decompiler/config/jak3/all-types.gc
```diff
@@ -11254,25 +11254,25 @@
...
+  (gs-roll-challenge)
+  (gs-spargus-turret-game)
+  (gs-jetboard-challenge)
+  (gs-gun-yellow-2-course)
+  (gs-gun-red-2-course)
+  (gs-ratchet-gun-course)
+  (gs-clank-gun-course)
+  (gs-eco-grid-game)
+  (gs-marauder-challenge)
+  (gs-unused-arena-training)
```
dallmeyer: I followed the naming convention from the corresponding progress text-ids, but we can rename if needed
https://github.com/open-goal/jak-project/blob/6c719ae000f39cd8a5761399bf3bed3cf31a4ca9/goal_src/jak3/engine/ui/text-h.gc#L1280-L1295

Hat-Kid: It looks like the real enum names are here, so I would just use those.

https://github.com/open-goal/jak-project/blob/6c719ae000f39cd8a5761399bf3bed3cf31a4ca9/goal_src/jak3/engine/game/game-info.gc#L1869-L1925

decompiler/config/jak3/ntsc_v1/anonymous_function_types.jsonc
```diff
@@ -27,7 +27,11 @@
     [
       22,
       "(function symbol symbol continue-point game-save resetter-spec none :behavior process)"
-    ]
+    ],
+    [28, "(function game-score (pointer-float))"],
+    [29, "(function game-score int)"],
+    [30, "(function game-score float int)"],
+    [31, "(function game-score int float)"]
```
Hat-Kid: These are not needed, there's no anonymous functions with these IDs.

Hat-Kid (changes requested):

decompiler/config/jak3/ntsc_v1/anonymous_function_types.jsonc
dallmeyer: oop, thx

Hat-Kid (approved):

PR #4410, opened 2026-09-09
cmake: point to correct draco_features.h location

BreakPoints:
Fixes the following error when running the "Install Jak" option in VS:
```
  CMake Error at third-party/draco/cmake_install.cmake:841 (file):
    file INSTALL cannot find
    "C:/Users/{WindowsUsername}/Documents/Coding/jak-project/out/build/Release/draco/draco_features.h":
    No error.
  Call Stack (most recent call first):
    cmake_install.cmake:48 (include)
  
  
  FAILED: [code=1] CMakeFiles/install.util 
  C:\WINDOWS\system32\cmd.exe /C "cd /D C:\Users\Jack\Documents\Coding\jak-project\out\build\Release && "C:\Program Files\Microsoft Visual Studio\18\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" -P cmake_install.cmake"
  ninja: build stopped: subcommand failed.
Install succeeded.
```

The build was still useable but the error was bugging me.

Our version of `draco_features.h` is stored in `third-party/draco/src/draco/draco_features.h` but `draco_install.cmake` was not updated to reflect this change. Use the already defined `${draco_features_file_name}` from `draco_options.cmake` to point to the right place.

Disclosure: I didn't want to spend too much time dealing with cmake so I pointed Qwen at the problem and it found a very simple solution. Using `${draco_features_file_name}` to avoid hardcoding is neat. I fully understand the fix, have tested it and all the descriptions were written by me.
(AI Assisted)

### DIFF
--- a/third-party/draco/cmake/draco_install.cmake
+++ b/third-party/draco/cmake/draco_install.cmake
@@ -60,7 +60,7 @@ macro(draco_setup_install_target)
               DESTINATION "${target_directory}")
     endforeach()
 
-    install(FILES "${draco_build}/draco/draco_features.h"
+    install(FILES "${draco_features_file_name}"
             DESTINATION "${includes_path}/draco/")
 
     install(TARGETS draco_decoder DESTINATION "${bin_path}")


Zedb0T:
How much time did you spend testing this? 

This is an annoying bug, but we have a very well tested fix over on mod base already that seems to resolve the issue without breaking anything.

xTVaser:
So there is definitely something wrong with draco's CMake causing this intermittent problem to happen for some people (perhaps a cmake cache issue, I can never reproduce it on any of the 3 platforms on my end).

In a nutshell, this is how it actually works:
- `draco_features.h` is not stored or committed anywhere, it is generated when CMake runs https://github.com/open-goal/jak-project/blob/master/third-party/draco/cmake/draco_options.cmake#L285
- that file then gets moved to `${draco_features_file_name}` https://github.com/open-goal/jak-project/blob/ce97ce959b8c773097f593bf42f470555f6a6e2b/third-party/draco/cmake/draco_options.cmake#L346-L349
- and then finally, actually moved into the build path via the `install` command that is modified here

However what's extra confusing is that the draco cmake has this section that modifies file paths https://github.com/open-goal/jak-project/blob/ce97ce959b8c773097f593bf42f470555f6a6e2b/third-party/draco/cmake/draco_install.cmake#L40-L54

My assumption is that people have some stale entry in their cache / some weird cmake variables causing some assumption to be broken, the logic above that runs the file path changes never runs again, and then the file never gets moved to where the `INSTALL` command needs it to be.

This change is probably the better way to go because we are just...simply pointing to where the file is generated instead of this nonsense.  The mod-base does a similar thing to dracos original code just in a very strange way....(it loops through all configured files just to move a single file...? https://github.com/OpenGOAL-Mods/OG-Mod-Base/pull/157/changes)

dallmeyer:
> it loops through all configured files just to move a single file

I think this is just doing it for both Release and for Debug folders is all

BreakPoints:
> How much time did you spend testing this?
> 
> This is an annoying bug, but we have a very well tested fix over on mod base already that seems to resolve the issue without breaking anything.

I've run extract, rip-level and repl compile for each game. The games all boot and jumping into a level works fine. I've loaded up multiple .glb exported files from each game into Blender as the AI suggested draco was moved so that tinygltf could access it (#3723 that touches our draco files implies this). So far everything seems as it was before this change. Is draco even being used if the install scrip can be changed with no apparent change to the output? (See Edit)


> My assumption is that people have some stale entry in their cache / some weird cmake variables causing some assumption to be broken, the logic above that runs the file path changes never runs again, and then the file never gets moved to where the `INSTALL` command needs it to be.

What's strange is I can 100% reproduce this on a clean copy of jak-project. Happens when building Release or Debug, static and non static. In VS CMakes output is quickly closed when all it done and the error window is shown with no errors but every time I reopen CMakes output it has this error.


In [upstream draco_options.cmake line 20](https://github.com/google/draco/blob/052a31124c75e64dfbf4f9bc0df066e626aeef41/cmake/draco_options.cmake#L20) is `set(draco_features_file_name "${draco_build}/draco/draco_features.h")` and [draco_install.cmake line 51](https://github.com/google/draco/blob/052a31124c75e64dfbf4f9bc0df066e626aeef41/cmake/draco_install.cmake#L51) is `install(FILES "${draco_build}/draco/draco_features.h"`

In [our draco_options.cmake](https://github.com/open-goal/jak-project/blob/a42c0719967d0869eb9ecf4cfef91a2857991428/third-party/draco/cmake/draco_options.cmake#L20) this was changed to `set(draco_features_file_name "${draco_src_root}/draco_features.h")` but [draco_install.cmake](https://github.com/open-goal/jak-project/blob/a42c0719967d0869eb9ecf4cfef91a2857991428/third-party/draco/cmake/draco_install.cmake#L63) was left with `install(FILES "${draco_build}/draco/draco_features.h"`.

For consistency I could mirror the `${draco_src_root}/draco_features.h` change to `draco_install.cmake` instead of reusing the `${draco_features_file_name}` define but it does look like an intended change just wasn't replicated somewhere else it should have been. Further than that, I do not know.


Edit:
I've compared the .glb output from master vs this PR and almost all files come out exactly the same size (down to the bytes shown in the properties window) with strange exceptions such as `jakc-highres-lod0.glb` loosing ~41KB.
Left is master, right is this PR:
image

PR #4417, opened 2026-09-16
decompiler: limit level extraction workers to hardware concurrency

trippjoe:
fixes an issue from my previous pull request #4278.
this change restores the previous behavior prior and prevents the level extraction process from creating too many threads, ultimately resulting in the decompiler failing.


decompiler/level_extractor/extract_level.cpp
```diff
@@ -417,7 +417,7 @@ void extract_all_levels(const ObjectFileDB& db,
                       game_version_names[config.game_version] / "entities";
   file_util::create_dir_if_needed(entities_dir);
 
-  int num_workers = dgo_names.size();
+  int num_workers = std::min<int>(dgo_names.size(), std::thread::hardware_concurrency());
```
xTVaser: can you make this change inside the overloaded `run` function within the `SimpleThreadGroup` class instead of here so that the mistake is less easy to make going forward (or both).  It's something I did on my branch to try to improve performance that I ultimately scrapped and never merged.

But also, this is very strange that this alone would cause a failure.  Over-allocating threads does not typically cause a crash, it's just a waste / inefficient (CPU scheduler has a harder time distributing the work). Especially considering that basically everyone that has decompiled the game has ran this with WAY more threads than their CPU has available (there are over a hundred DGOs in jak2)

PR #4420, opened 2026-09-20
extractor: dont default to x86 instructions on arm64

trippjoe:


decompiler/extractor/main.cpp
```diff
@@ -138,7 +138,7 @@ ExtractorErrorCode compile(const fs::path& iso_data_path, const std::string& dat
   // Determine which config to use from the database
   const auto version_info = get_version_info_or_default(iso_data_path);
 
-  Compiler compiler(game_name_to_version(version_info.game_name), emitter::InstructionSet::X86);
+  Compiler compiler(game_name_to_version(version_info.game_name), emitter::kNativeInstructionSet);
```
xTVaser: this value is really only intended for testing code (it has a comment explaining that).  We should instead:
- add an argument to the extractor to set the instr set
- which can have a default based on the compiler flags.

Basically copy what is done in `goalc`

xTVaser:
ok to test

